PR #23, opened 2026-03-03
Fix terminal theme in light/system mode

ittaiz:
## Summary

- Terminal was always rendered with a hardcoded dark (Tokyo Night) palette regardless of the app theme
- Added `DARK_TERMINAL_THEME` (Tokyo Night, unchanged) and `LIGHT_TERMINAL_THEME` (GitHub Light)
- A `MutationObserver` on `document.documentElement[data-theme]` detects theme switches and recreates the terminal with the correct palette
- Works for all three theme settings: Dark, Light, and System

## Test plan

- [x] Switch app theme to Light — terminal should use GitHub Light palette
- [x] Switch app theme to System — terminal should follow OS preference
- [x] Switch app theme back to Dark — terminal should return to Tokyo Night
- [ ] Verify git diff colors (red/green) are readable in light mode

### DIFF
--- a/src/mainview/TerminalView.tsx
+++ b/src/mainview/TerminalView.tsx
@@ -408,7 +408,14 @@ function TerminalView({ ptyUrl, taskId }: TerminalViewProps) {
 				termRef.current = null;
 			}
 		};
-	}, [ptyUrl, taskId, resolvedTheme]);
+	}, [ptyUrl, taskId]);
+
+	useEffect(() => {
+		if (termRef.current) {
+			termRef.current.options.theme =
+				resolvedTheme === "light" ? LIGHT_TERMINAL_THEME : DARK_TERMINAL_THEME;
+		}
+	}, [resolvedTheme]);
 
 	function handleDragOver(e: React.DragEvent<HTMLDivElement>) {
 		e.preventDefault();


h0x91b (commented):
Hey there, Claude here — the AI doing a code review on this PR.

Overall the idea is solid — terminal theme should indeed follow the app theme. The `MutationObserver` approach for detecting `data-theme` changes is clean. However, there's one significant issue with the implementation.

## Critical: Full terminal teardown on theme change

Adding `resolvedTheme` to the dependency array of the main `useEffect` (the ~300-line monster that creates the terminal, WebSocket, mouse tracking, resize observer, etc.) means **every theme toggle causes a full destroy + recreate cycle**:

1. WebSocket connection is closed and reopened
2. Terminal WASM instance is disposed and rebuilt
3. All scrollback/terminal state is lost
4. FitAddon, ResizeObserver, mouse tracking — all torn down and re-setup
5. User sees a flash/flicker

**ghostty-web already supports runtime theme changes** via its Proxy-wrapped `options` property ([docs](vendor-docs/ghostty-web/api/terminal.md#runtime-option-changes)):

```typescript
term.options.theme = resolvedTheme === "light" ? LIGHT_TERMINAL_THEME : DARK_TERMINAL_THEME;
```

The fix is to **remove `resolvedTheme` from the main effect's dependency array** and add a separate lightweight effect:

```typescript
useEffect(() => {
    if (termRef.current) {
        termRef.current.options.theme =
            resolvedTheme === "light" ? LIGHT_TERMINAL_THEME : DARK_TERMINAL_THEME;
    }
}, [resolvedTheme]);
```

This way theme switches are instant, no WebSocket reconnect, no state loss, no flicker.

Everything else looks good — theme constants are clean, the `MutationObserver` + `useState` pattern is the right call, and the changelog entry is on point.

ittaiz:
Good catch. Removed `resolvedTheme` from the main effect's deps and added a separate lightweight effect that sets `term.options.theme` directly — ghostty-web's Proxy handles it in real-time with no teardown.

PR #85, opened 2026-03-06
Add "Create PR" button to git operations bar

### LINKED ISSUE #61: Feature Request: Additional button to create a PR in GitHub alongside other git operations
Would be great that after `Push` - we could create a PR without leaving the app - and open it in the browser after created.

An additional button on the top bar.

Image

h0x91b:
New button in TaskInfoPanel between Push and Merge that creates a
GitHub Pull Request via `gh pr create --fill --web`. If a PR already
exists for the branch, falls back to opening it with `gh pr view --web`.

The button is enabled only when the branch has commits and all are
pushed (unpushed === 0). Follows the existing git operation pattern:
RPC handler runs the command in a tmux split pane with monitoring.

Closes #61


h0x91b:
Screenshot 2026-03-06 at 9 31 59

PR #183, opened 2026-03-09
Add Review by Colleague column

ittaiz:
## Summary

- Adds a new built-in `review-by-colleague` status/column that sits between custom columns and `completed` on the Kanban board
- A background PR detection poller (5 min interval) automatically moves tasks from `review-by-user` → `review-by-colleague` when an open non-draft GitHub PR is detected for that branch
- The existing merge polling now also watches `review-by-colleague` tasks, so `branchMerged` fires from either pre-merge status
- Project Settings adds a **Peer Review Column** toggle (default on) to hide the column for projects that don't use peer code review
- When a user has manually reordered columns, the new column is inserted immediately before `completed` (not appended at the tail)

## Implementation details

- `TaskStatus` union extended with `"review-by-colleague"` (active status — worktree/PTY preserved)
- `git.run` exported so the PR poller can call `gh pr list` without spawning directly
- `_resetPRPollerState` exported for test isolation
- `Project.peerReviewEnabled?: boolean` field; `data.updateProject` and `updateProjectSettings` RPC accept it
- All i18n strings added to en/ru/es

## Test plan

- [x] `isActive("review-by-colleague")` returns true
- [x] `checkOpenPRsForPromotion`: promotes on open non-draft PR, skips drafts/empty/never-pushed/wrong-status/null-worktree/peerReviewEnabled=false/already-promoted/gh-failure
- [x] `KanbanBoard`: default order, hidden when disabled, inserted before completed when missing from stored order, stays in stored position, hidden even when in stored order
- [x] `ProjectSettings`: toggle on by default, reflects project field, click flips state, saves correct value
- [x] All 517 tests pass, TypeScript clean


ittaiz:
Off
image
image

On
image

PR #272, opened 2026-03-11
Add collapsible Kanban columns (#211)

### LINKED ISSUE #211: consider - todo, completed and cancelled columns should be collapsable and collapsed by default
I think these are not the main focus
the cancelled is not important 90% of the time
the completed is nice, but it's just nice to look at from time to time
I expect the todo to be a queue that is mostly at size 0, you just add something and lunch it.
for the cases you want to time your tasks, it makes sense to have a column, but again I think it should be collapsed by default.

This way there is more real estate for the actual content.

not a UX expect but IMHO-
expanding can be on hover, and should  auto collapse after exiting the column
if a column have been clicked, it should another click to collapse it

roiros:
## Summary

Implements collapsible Kanban columns (closes #211).

- **Todo, completed, and cancelled** columns collapse by default into narrow 48px strips, freeing horizontal space for active work columns
- **Hover to peek** inside a collapsed column (200ms delay), mouse-leave to re-collapse (400ms delay)
- **Pin icon** in hover-expanded header to keep column open permanently; **filled pin (unpin) icon** to collapse it back
- **"+" button** on collapsed Todo strip for quick task creation without expanding
- **Flex layout**: expanded columns auto-fill available screen width (`flex-1 min-w-[14rem]`)
- **Drag-and-drop** preserved on collapsed columns (drag-expand on enter)
- Collapse state **persists per project** in localStorage
- New `useColumnCollapse` hook with ref-based timer callbacks (no stale closures)
- i18n for English, Russian, Spanish
- 16 new tests (7 hook + 6 column + 3 board)

## Test plan

- [x] `bun run lint` passes
- [x] `bun run test` passes (581 tests)
- [ ] Visual: collapsed columns render as narrow strips with color dot, badge, vertical label
- [ ] Visual: hover over collapsed strip → column expands after 200ms
- [ ] Visual: pin icon appears in header of hover-expanded column
- [ ] Visual: clicking pin keeps column open; filled pin (unpin) icon shows for collapse
- [ ] Visual: "+" button on collapsed Todo creates a task without triggering expansion
- [ ] Visual: columns auto-fill width; collapsing/expanding causes smooth resize
- [ ] Drag a task onto a collapsed column → column expands, task drops correctly

Screenshot 2026-03-11 at 19 32 22

### DIFF
--- a/src/mainview/components/KanbanColumn.tsx
+++ b/src/mainview/components/KanbanColumn.tsx
@@ -446,7 +446,7 @@ function KanbanColumn({
 							aria-label={t("kanban.collapseColumn")}
 							title={t("kanban.collapseColumn")}
 						>
-							{"\u{F0142}"}
+							{"\u{F0403}"}
 						</button>
 					)}
 				</div>
--- a/src/mainview/i18n/translations/en/kanban.ts
+++ b/src/mainview/i18n/translations/en/kanban.ts
@@ -7,7 +7,7 @@ const kanban = {
 	"kanban.failedCreate": "Failed to create task: {error}",
 	"kanban.showMore": "Show more ({count})",
 	"kanban.showLess": "Show less",
-	"kanban.collapseColumn": "Collapse column",
+	"kanban.collapseColumn": "Unpin column",
 
 	// CreateTaskModal
 	"createTask.title": "New Task",
--- a/src/mainview/i18n/translations/es/kanban.ts
+++ b/src/mainview/i18n/translations/es/kanban.ts
@@ -7,7 +7,7 @@ const kanban = {
 	"kanban.failedCreate": "Error al crear tarea: {error}",
 	"kanban.showMore": "Mostrar más ({count})",
 	"kanban.showLess": "Mostrar menos",
-	"kanban.collapseColumn": "Contraer columna",
+	"kanban.collapseColumn": "Desanclar columna",
 
 	// CreateTaskModal
 	"createTask.title": "Nueva tarea",
--- a/src/mainview/i18n/translations/ru/kanban.ts
+++ b/src/mainview/i18n/translations/ru/kanban.ts
@@ -7,7 +7,7 @@ const kanban = {
 	"kanban.failedCreate": "Не удалось создать задачу: {error}",
 	"kanban.showMore": "Показать ещё ({count})",
 	"kanban.showLess": "Свернуть",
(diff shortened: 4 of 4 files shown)


h0x91b (changes requested):
Hey, Claude here — the AI assistant doing the re-review.

Thanks for the fixes in `be37eb8` — stale closures, `useMemo` stability, icon sizes, and the unused i18n key cleanup are all good. However, the main issues from the first review are still unresolved:

---

### Still open from the first review

**1. `transition-all duration-200 ease-in-out` — performance regression (KanbanColumn.tsx:258, :339)**
Both the collapsed strip and the expanded column still have `transition-all` which animates the `width` change. On boards with many tasks this causes a full flexbox reflow on every animation frame. Please remove it — instant snap is the right call.

**2. Hover-expand is still present**
The entire hover-expand machinery (`hoverHandlers`, `HOVER_EXPAND_DELAY`, `HOVER_COLLAPSE_DELAY`, `temporarilyExpanded` state, pin button) is still fully intact. This causes accidental expansions during normal cursor movement across the board. The suggestion was to remove it entirely and use click-only expand/collapse.

**3. `onDragEnd` is defined but never wired (useColumnCollapse.ts:160)**
`collapseDragHandlers` exposes `onDragEnd` but it's never called anywhere in the JSX — neither in the collapsed nor in the expanded column render. Drag that ends abruptly (dropped outside a column) may leave expanded state lingering.

**4. Indentation still inconsistent (KanbanBoard.tsx:484)**
`tipState` prop is at 7 tabs while surrounding props are at 8. Minor, but the fix commit claimed to address this.

---

### New issue: expanded column is too narrow

When a collapsed column is expanded (by hover or click), it renders much narrower than a regular column. See screenshot in PR comments — task cards don't fit, text is oversized relative to the column width. The expanded state should render at the exact same width as a normal (never-collapsed) column. Right now `flex-1 min-w-[14rem]` isn't distributing space the same way when multiple columns are collapsed.

---

### Rebase needed

The branch is behind `main` — please rebase and resolve conflicts before the next review round.

Screenshot 2026-03-13 at 13 53 47

h0x91b:
Hey @roiros, great work on this! The collapsible columns feature is looking solid. I made a small tweak on top — swapped the collapse icon to a pin for better UX clarity and updated the i18n labels to match. Everything else is clean, thanks for iterating on the review feedback. Approving this for merge 👍

PR #454, opened 2026-04-13
feat: warn when creating task on non-base branch

### LINKED ISSUE #451: Warn when creating a task while repo is on a non-base branch
## Summary
Creating a new task should default to the project base branch. If the repo is currently checked out to some other branch, the app should stop and ask the user explicitly whether they want to start from the current checked-out branch or from the configured base branch.

## Problem
The expected safe default is simple: new tasks start from `baseBranch`, and that base branch should be clean and stable.

But in real life, users sometimes leave the main repo checked out on some other branch. In that situation, it becomes too easy to accidentally create a new task/worktree from the wrong context and inherit changes/commits that were never meant to be part of the new task. If this goes unnoticed, the new work can become a mess very quickly.

## Expected behavior
- Default behavior stays: create new tasks from the configured `baseBranch`
- If the repo is currently checked out to a different branch than `baseBranch`, show a confirmation popup
- The popup should make the choice explicit:
  - start from the current checked-out branch
  - start from the configured base branch
- The UI should explain that starting from the current branch may carry over work that belongs to another feature/task

## Why this matters
This is a guardrail issue, not just a convenience improvement. Users can accidentally pollute a fresh task with unrelated branch context and not notice until much later.

## Reported by
Evgeny Alterman

## Acceptance criteria
- Creating a task still defaults to `baseBranch`
- If current checked-out branch differs from `baseBranch`, the user gets an explicit choice before task/worktree creation
- The choice text is clear about the risk of inheriting branch context from another feature
- No popup is shown when the repo is already on `baseBranch`

sworgkh:
## Summary

Fixes h0x91b/dev-3.0#451.

The New Task modal previously pre-filled the branch selector silently with the project's current checked-out branch when it differed from the configured base branch. This made it too easy to accidentally inherit unrelated commits and changes from another feature branch into a new task's worktree.

Now the modal shows an explicit, blocking confirmation banner when the main clone is on a non-base branch. The user must pick between:

- **Start from `<base>`** (primary, recommended) — omits `existingBranch`, worktree is created from the project's base branch
- **Start from `<current>`** — passes `existingBranch` so the worktree starts from the currently checked-out branch

`Save` and `Save & Start` stay disabled until the choice is made. No banner appears when the main clone is already on the base branch.

Covered by updated unit tests in `CreateTaskModal.test.tsx` and translations in en/ru/es.

image


h0x91b (commented):
Thanks for the PR. The visual treatment is fine, but I need the interaction logic to match the intended behavior more closely:

- Please keep the existing branch picker visible and keep auto-filling it with the current non-base branch. The warning should not replace the picker or block the form before Save; it should only appear when the user clicks Save / Save & Start.
- The confirmation should trigger only when the selected branch is the current non-base branch. If the user clears that branch or chooses a different one, task creation should proceed without that warning.
- The warning should also detect when the main repo is dirty and surface that explicitly.
- The dirty-state copy must clarify that uncommitted Git changes stay in the main repo and do not move into the new task. Right now the wording implies that uncommitted changes may carry over, which is misleading.
- The base-branch copy should explicitly mean the project base branch, and the wording/button text should make that unambiguous (for example, project base branch / origin/<base> when available), rather than a generic "main" that can be read as either local or remote.
- The risk copy should focus on committed branch history/context carrying over from the current branch, not on uncommitted changes flowing into the new task.

If you update the PR to keep the picker-based flow and align the warning text/logic with the points above, that will match the behavior I want for this feature.

h0x91b (changes requested):
Thanks for the update — this is very close and the overall direction now matches what I wanted.

There is still one blocking logic issue before I can approve it:

In `CreateTaskModal`, the save-time confirmation can still be skipped if the user clicks `Save` before `getProjectCurrentBranch()` has finished loading. In that case `handleSubmit()` awaits the branch lookup, but then continues using the stale `selectedBranch` value from the old render, so the task can be created from the base branch immediately without ever showing the explicit branch-choice confirmation.

Could you please fix that race so the non-base-branch confirmation is guaranteed even on the first fast `Save` click, and add a regression test for that path?

Once that is fixed, this should be in good shape.

sworgkh:
Fixed in ed782f0.

`handleSubmit` now captures whether the lookup had already completed before Save. If it hadn't (i.e. the user clicked Save before `getProjectCurrentBranch()` resolved), we treat the freshly-loaded non-base branch as the effective selection — stale-null `selectedBranch` from the pre-auto-fill render can no longer bypass the confirmation. When the lookup completed before Save, the visible selection remains authoritative so clearing or switching the chip still suppresses the warning as before.

Added a regression test (`still asks on a fast Save click while branch lookup is in-flight`) that holds `getProjectCurrentBranch` pending until after the Save click, then asserts the confirmation appears and `createTask` was not called.

h0x91b (approved):

PR #468, opened 2026-04-15
Add Tasks Quick Switch

sworgkh:
## Summary

Adds Tasks Quick Switch, a keyboard-driven task switcher that behaves like app switching: hold the configured modifier, tap the trigger key to move through active tasks, and release the modifier to open the selected task.

## What was implemented

- Added a Tasks Quick Switch modal that lists active tasks across projects and keeps the current task preselected.
- Added keyboard navigation for cycling through tasks while the switcher is open.
- Added navigation on modifier release so the selected task opens only after the user finishes switching.
- Added Global Settings for Tasks Quick Switch filters so users can decide which active task types appear in the switcher.
- Added support for both built-in statuses and custom columns in those filters.
- Added configurable shortcut presets plus custom shortcut recording.
- Normalized recorded letter shortcuts by key code so custom bindings stay stable across layouts and modifier combinations.
- Moved custom shortcut capture into a dedicated modal flow.

image
image
Screenshot at Apr 15 17-26-44

## Follow-up fixes and UX polish

- Rendered the quick-switch modal through a portal so it stays centered in the actual viewport instead of drifting inside long scrolled layouts.
- Made quick switch use live in-app settings immediately instead of briefly falling back to stale backend settings on first open.
- Restored scroll position and focus back to the Tasks Quick Switch settings section after saving or canceling custom shortcut recording.
- Improved selected-row contrast in the quick-switch modal so the active choice is easier to see in both light and dark themes.

## Validation

- `bun run lint`
- `bunx --bun vitest run src/mainview/components/__tests__/TasksQuickSwitchModal.test.tsx`
- `bunx --bun vitest run src/mainview/components/__tests__/GlobalSettings.test.tsx -t \"tasks quick switch\"`

`bun run test` still hits the pre-existing Vitest ESM/CJS startup issue in this environment.


ittaiz:
Is this something like `f` key in vimium?
https://sudoscience.blog/2025/07/10/a-few-fast-easy-commands-to-start-using-vimium-c/

sworgkh:
> Is this something like `f` key in vimium? https://sudoscience.blog/2025/07/10/a-few-fast-easy-commands-to-start-using-vimium-c/

More of a Mac Cmd+Tab or Win Alt+Tab

ittaiz:
How many active tasks do you have? I have >20 so I'm a bit concerned about it. 
Maybe I'm wrong. 
Also- how did you "unslop" here? 2500 LOC can easily smell as slop so wondering 🙏 

And thanks for the effort!!!!!!!!!!!!

sworgkh:
> How many active tasks do you have? I have >20 so I'm a bit concerned about it. Maybe I'm wrong. Also- how did you "unslop" here? 2500 LOC can easily smell as slop so wondering 🙏
> 
> And thanks for the effort!!!!!!!!!!!!

Didn't Unslop. will do. though it has a lot of small functionality like settings and all the keybindings handling
I have generally 4-5 repos and there are 1-3 tasks  but I can genuinely focus on 1-3 so I always switch between them and it takes me a lot of mouse clicks and especially precise ones to the down chevron on the top navigation breadcrumbs is chalenging :D

sworgkh:
OK. I think now I Finally Cleaned it up and fixed a lot of small UX issues.
PR is ready.

PR #492, opened 2026-04-21
fix: tmux task-switch glitch and freeze

sworgkh:
## Summary

Fixes the garbled text + flicker on every task switch. Three independent causes, each fixed here:

- **Dropped the capture-pane replay on WS reconnect** (`src/bun/pty-server.ts`). The previous \`capturePane()\` ran at tmux's current internal pane size — stale whenever the last client disconnected at a different size — so captured lines rendered at the wrong width into the new client terminal, producing the overlap that only cleared after manually adding a pane. The client's resize dance forces tmux's SIGWINCH full-pane redraw down the natural PTY data path, which is the authoritative paint.
- **Keyed \`TaskTerminal\` by \`taskId\`** (`src/mainview/components/TaskWorkspacePane.tsx`). Without the key, a task switch left the previous task's cached \`ptyUrl\` in state while \`taskId\` flipped to the new one, causing \`TerminalView\` to remount *twice* per switch — once with (stale url, new taskId) and again once the new url arrived. That's the "clean of screen of the task we leave" flash. With the key, each switch cleanly unmounts the old \`TaskTerminal\` and mounts \`TerminalView\` once.
- **Row-nudged the resize dance** (`src/mainview/TerminalView.tsx`). Nudging columns (\`cols-1\` → \`cols\`) made tmux repaint at the narrower width and then the target width, re-wrapping text between the two paints — the "refresh and realign" flicker. Nudging rows (\`rows+1\` → \`rows\`) keeps text wrapping identical; the only visible difference is one extra blank bottom row that disappears on the second paint.

Decision record at \`decisions/041-defer-initial-capture-pane-on-reconnect.md\` walks through what was tried and why these three stuck.

#skipreview


h0x91b:
works very well! I like it

h0x91b:
I have one machine there that stoped work with the buffer of tmux anymore, not sure why, but temporary I will revert this one and will dig deeper

PR #512, opened 2026-04-26
Fix git fetch failing in SSH-authenticated repos

alonkochba:
## Summary

- `resolveShellEnv()` now captures `SSH_AUTH_SOCK` from the user's login shell alongside PATH, LANG, XDG_CONFIG_HOME, and GH_CONFIG_DIR
- Applied to `process.env` at startup so all spawned processes (git, etc.) inherit the SSH agent socket
- Fixes `git fetch origin --quiet` silently failing for repos with SSH remotes (`git@github.com:...`), while HTTPS repos were unaffected

**Root cause:** macOS `.app` bundles launch with a stripped environment. The app recovered most needed vars but missed `SSH_AUTH_SOCK`, which git needs to authenticate via SSH keys through the agent.

## Test plan

- [ ] Restart the app
- [ ] Open a project whose remote is `git@github.com:...` (SSH)
- [ ] Create a new task — the Fetch step during worktree setup should succeed
- [ ] In the branch picker, click **Fetch** — remote branches should populate
- [ ] `bun run lint` passes (no new type errors)

### DIFF
--- a/change-logs/2026/04/26/fix-git-fetch-ssh-auth-sock.md
+++ b/change-logs/2026/04/26/fix-git-fetch-ssh-auth-sock.md
@@ -1,0 +1,1 @@
+Fixed git fetch failing silently in repos that use SSH authentication (e.g., git@github.com: remotes). The app now captures SSH_AUTH_SOCK from the user's login shell during startup, so the SSH agent socket is available to git and other spawned processes.
--- a/src/bun/__tests__/shell-env.test.ts
+++ b/src/bun/__tests__/shell-env.test.ts
@@ -61,13 +61,14 @@ describe("shell environment bootstrap", () => {
 		expect(spawnMock).not.toHaveBeenCalled();
 	});
 
-	it("captures gh-related config variables from the login shell", async () => {
+	it("captures gh-related config variables and SSH_AUTH_SOCK from the login shell", async () => {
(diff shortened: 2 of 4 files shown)


h0x91b:
Hi, this is Claude (the AI assistant working on this branch) — I did a review pass on the diff.

LGTM. The change mirrors the existing `PATH` / `LANG` / `XDG_CONFIG_HOME` / `GH_CONFIG_DIR` resolution pattern in `shell-env.ts` and `index.ts` exactly — same shape, same empty-value guard, same `original*` log line. Test was extended in place rather than duplicated, and there's a one-line changelog entry. Low blast radius: an unset `SSH_AUTH_SOCK` correctly leaves `process.env` untouched.

Two small non-blocking notes for a follow-up (not for this PR):
- The JSDoc on `resolveShellEnv` still lists only `PATH` / `LANG` / `GH_CONFIG_DIR` / `XDG_CONFIG_HOME` — could mention `SSH_AUTH_SOCK` next time someone touches that block.
- The captured value is a startup snapshot. On macOS, launchd rotates the agent socket on logout/login, so a long-running dev-3.0 process could end up with a stale path until restart. This is a pre-existing limitation of the snapshot approach for *all* env vars resolved here, not something this PR introduced — worth a separate issue if it ever bites.

Otherwise: approve, ready to merge.

PR #527, opened 2026-05-02
docs: rework Linux install section in README

h0x91b:
## Summary

Linux is no longer framed as "headless only" in the README — the brew formula installs a universal `dev3` CLI that works on every Linux box (server or desktop), and we already build a real Electrobun desktop bundle for Linux (`stable-linux-x64-dev-3.0.tar.zst`).

## Changes

- Section heading: **"Linux server — headless mode (`dev3 remote`)"** → **"Linux"**.
- Promoted Homebrew install instructions out of the collapsed `<details>` block — the previous placement was confusing on a fresh VM where `brew` isn't installed yet (you couldn't see how to install it without expanding the spoiler).
- Added a visible warning that the Homebrew installer refuses to run as `root`, with a one-liner to create a regular user. This bites anyone starting from a fresh root-only cloud VM.
- New **"Desktop GUI on Linux"** subsection pointing to the Electrobun `tar.zst` bundle plus the `apt` deps for `libwebkit2gtk-4.0`, `libgtk-3`, etc.
- Pre-built CLI tarball path is now clearly framed as a fallback for minimal containers, not the only option.

Doc-only change — no code touched.


h0x91b:
Subsumed by #529 (`feat/dev3-gui-subcommand`). The README rework from this PR is included verbatim in #529, plus the new `dev3 gui` subcommand replaces the manual `tar.zst` paragraph entirely. Closing to avoid duplicate work.

PR #582, opened 2026-05-25
fix: update Codex config patcher for 0.133 schema renames

ittaiz:
## Summary

Codex 0.133 renamed two keys in `config.toml` and emits deprecation warnings (and silently ignores) the old names:

- `permissions.*.filesystem.":project_roots"` → `":workspace_roots"`
- `[features].codex_hooks` → `[features].hooks`

This was visible on launch as:

```
⚠ Configured filesystem path `:project_roots` is not recognized by this version of Codex and will be ignored.
⚠ `[features].codex_hooks` is deprecated. Use `[features].hooks` instead.
```

The `:project_roots` warning was more than cosmetic — it meant the `[permissions.dev3.filesystem.":project_roots"] "." = "write"` rule we install was being silently ignored on Codex 0.133+, so the dev3 permission profile lost its in-worktree write grant.

Changes:

- `src/bun/codex-config.ts`
  - Emit the new key names when creating sections (`:workspace_roots`, `hooks`).
  - `cleanupLegacySections` migrates existing files in place:
    - Rewrite the two `:project_roots` headers we manage (`permissions.dev3.*` and `permissions.workspace.*`) to `:workspace_roots`. User-owned profiles are left alone.
    - Inside `[features]`: rename `codex_hooks` → `hooks` when only the deprecated alias exists. If both already exist (Codex 0.133+ silently mirrors the alias into `hooks`, so users can end up with duplicates), drop the `codex_hooks` line so Codex stops warning.
- `src/cli/commands/install-skills.ts`: updated the trailing summary line to say `hooks` instead of `codex_hooks`.
- Tests: updated existing assertions to the new key names, added coverage for the rename + the dedupe case.

## Test plan

- [x] `bun run lint`
- [x] `bun run test` (46 backend files / 1175 tests + 62 mainview files / 1162 tests — all green)
- [x] Manual: relaunched the app from this branch against a real `~/.codex/config.toml` containing both legacy keys → patcher rewrote `:project_roots` → `:workspace_roots` and removed the duplicate `codex_hooks`. New Codex panes start without either warning.
- [ ] Reviewer: confirm no third-party `[permissions.<custom>.filesystem.":project_roots"]` sections are touched (only `dev3` + `workspace` are migrated).

### DIFF
--- a/change-logs/2026/05/25/fix-codex-0.133-config-keys.md
+++ b/change-logs/2026/05/25/fix-codex-0.133-config-keys.md
@@ -1,0 +1,1 @@
+Updated the auto-managed `~/.codex/config.toml` patcher to match Codex 0.133 schema: `[permissions.*.filesystem.":project_roots"]` → `":workspace_roots"` and `[features].codex_hooks` → `[features].hooks`. Existing user configs are migrated on next app startup so the deprecation warnings stop appearing.
--- a/src/bun/__tests__/codex-config.test.ts
+++ b/src/bun/__tests__/codex-config.test.ts
@@ -12,7 +12,7 @@ describe("ensureCodexConfig", () => {
 			expect(result).toContain('trust_level = "trusted"');
 			expect(result).toContain('default_permissions = "workspace"');
 			expect(result).toContain("[permissions.workspace.filesystem]");
-			expect(result).toContain('[permissions.workspace.filesystem.":project_roots"]');
+			expect(result).toContain('[permissions.workspace.filesystem.":workspace_roots"]');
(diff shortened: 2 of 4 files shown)


h0x91b:
Thanks a lot for the PR. This is a very good catch, especially the edge case where Codex has already added `hooks` but the old `codex_hooks` key is still present alongside it. We missed that in our first pass.

We compared the implementations and carried that case into #580. The migration there now handles duplicate hook feature aliases in both directions, so it keeps compatibility with both newer and older Codex versions. I also rewrote the last commit in #580 with you as the author, since that part effectively came from this PR.

PR #625, opened 2026-06-07
Cmd+1..9 project switch always lands on the Kanban board

BnayaZil:
## Summary

Switching projects with Cmd/Ctrl+1..9 from inside a task view previously preserved the task layout in split open-mode (#619, gated in #623), landing the user in an empty split with no task selected. There is no task to show in the other project, so the switch now **always navigates to the target project's Kanban board**, regardless of the Task Open Mode setting.

## Changes

- `src/mainview/App.tsx` — removed the `dev3-task-open-mode` gating from the Cmd+1..9 handler; always navigate to `{ screen: "project" }`.
- `src/mainview/__tests__/App.test.tsx` — updated the "switch project (Cmd+1..9)" tests to expect the board landing (reproduce-first: tests went red, then the fix turned them green).
- Changelog entry added.

## Testing

- `bun run lint` — clean
- `bun run test` — 122 files / 2524 tests green (mainview + bun)
- Verified manually via `bun run dev`: Cmd+2 from a split task view, from the full-page task screen, and in fullscreen open-mode all land on the target board; board→board switching and Cmd+9 no-op unchanged.

### DIFF
--- a/change-logs/2026/06/07/fix-cmd-number-switch-lands-on-board.md
+++ b/change-logs/2026/06/07/fix-cmd-number-switch-lands-on-board.md
@@ -1,0 +1,1 @@
+Cmd/Ctrl+1..9 project switching now always lands on the target project's Kanban board, including when the current view is inside a task. This removes the split-mode task-view preservation introduced in #619/#623 — switching projects from a task previously left the user in an empty task layout.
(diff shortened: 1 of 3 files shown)


BnayaZil:
It was a mistake, apparently this behaviour was intended, IMO it's wrong but I guess I might become familiar with it eventually

PR #633, opened 2026-06-09
Fix false "Branch Merged" prompt on PR-review tasks

h0x91b:
## Summary

PR-review tasks are created from an existing branch (`existingBranch`). `deriveTaskBaseBranch` sets their `baseBranch` to that same branch, so the backend merge-detection poller (`checkMergedBranches`) compared the branch against `origin/<itself>` — a comparison that is trivially "merged" — and fired a false **Branch Merged** completion prompt even though nothing reached the real base branch.

The frontend path (`getBranchStatus`) was already protected by its `status.ahead > 0` guard, so the bug only surfaced through the backend poller.

Fix: skip the merge check in the poller when the resolved base branch equals the current branch (no distinct base to merge into).

Includes a regression test in `rpc-handlers.test.ts` verifying the poller does not push `branchMerged` (and never even calls `isContentMergedInto`) when base branch == current branch.

### DIFF
--- a/change-logs/2026/06/09/fix-review-task-false-merge.md
+++ b/change-logs/2026/06/09/fix-review-task-false-merge.md
@@ -1,0 +1,1 @@
+Fixed a false "Branch Merged" prompt on PR-review tasks. Such tasks use the reviewed branch as their base branch, so the merge-detection poller was comparing the branch against itself (trivially "merged"); it now skips tasks whose base branch equals the current branch.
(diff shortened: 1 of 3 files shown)


h0x91b:
Hi — this is Claude (the AI assistant working on this branch).

Closing this in favor of #634, which is now the canonical fix. I verified against the actual stored task data that the reported screenshot was caused by the poller comparing the review branch against itself (Strategy 1 fires before the gh-PR fallback), so #634's Strategy 3 head-commit gate alone wouldn't have covered it.

I've folded the better version of this poller fix (Variant C) into #634: instead of skipping review tasks outright (what this PR did), the poller now falls back to the project's real base branch, so a genuinely merged reviewed PR still gets the prompt. #634 now fixes both layers (poller + gh-PR fallback). Nothing is lost by closing here.

PR #634, opened 2026-06-09
Fix false-positive Branch Merged prompt on PR-review tasks

h0x91b:
## Summary

The merge-detection poller showed a false-positive **"Branch Merged — mark this task as completed?"** prompt on a PR-review task whose branch was *not* merged.

Root cause is in `isContentMergedInto` (`src/bun/git.ts`), **Strategy 3 — the GitHub PR fallback**. It returned `true` for *any* merged PR matching the head branch *name*, with no check that the current HEAD content was actually merged. This bites PR-review tasks specifically, since their worktree sits on an existing/reused branch tied to a PR: a previously-merged PR for a reused branch name (or an old merged PR coexisting with new unmerged work / an open PR) produced the false positive.

Strategies 1 (merge-tree) and 2 (patch-id) are content-based and were empirically confirmed to correctly return `false` for a genuinely unmerged branch.

## Fix

Strategy 3 now fetches `headRefOid` and only returns `true` when it equals the current local HEAD (`git rev-parse HEAD`). In every GitHub merge method (merge / squash / rebase) the head ref tip is left untouched, so a genuine merge always matches, while stale/reused-name PRs do not.

## Tests

Added two repro tests in `src/bun/__tests__/git-merge-detection.test.ts` (real git repo). Confirmed they fail on the old code (false `true`) and pass after the fix. The legitimate "main diverged before AND after squash" detection still returns `true`. Full suite green.

Decision record: `decisions/065-merge-detection-gh-pr-head-match.md`.

---
🤖 This PR was prepared by Claude (the AI assistant working on this branch).

### DIFF
--- a/change-logs/2026/06/09/fix-merge-detection-gh-pr-false-positive.md
+++ b/change-logs/2026/06/09/fix-merge-detection-gh-pr-false-positive.md
@@ -1,0 +1,1 @@
+Fixed a false-positive "Branch Merged" prompt on PR-review tasks. The GitHub PR fallback in merge detection trusted any merged PR matching the head branch name; it now requires the merged PR's head commit to equal the current local HEAD, so a stale/reused-name PR with new unmerged work no longer triggers the prompt.
--- a/decisions/065-merge-detection-gh-pr-head-match.md
+++ b/decisions/065-merge-detection-gh-pr-head-match.md
@@ -1,0 +1,37 @@
+# 065 — Gate the gh-PR merge-detection fallback on head commit match
+
+## Context
+
+The merge-detection poller showed a false-positive "Branch Merged — mark this task as
+completed?" prompt on a PR-review task whose branch was *not* merged. Review tasks are the
+common trigger because their worktree sits on an existing (often reused) branch tied to a PR.
+
+## Investigation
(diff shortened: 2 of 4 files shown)


h0x91b:
Hi, this is Claude (the AI assistant working on this branch) — closing this in favor of #633.

After comparing, #633 identifies the actual root cause of the screenshot bug: PR-review tasks get `baseBranch == branchName` (via `deriveTaskBaseBranch`), so the poller compares the branch against `origin/<itself>` and Strategy 1 (merge-tree) returns trivially `true`. Execution never reaches the gh-PR fallback (Strategy 3) that this PR was patching, so this change does not fix the reported issue.

My `headRefOid` gate addresses a different, rarer edge case (reused branch name with a *distinct* base) and isn't needed for this bug. Going with #633.

h0x91b:
Hi — Claude here (the AI assistant). Closed in favor of #633, which now carries the full superset of both fixes: the poller-level Variant C (review tasks compared against the project's real base branch, not the branch itself) plus this branch's gh-PR Strategy 3 head-commit gate. #633 handles strictly more, so nothing is lost here.

PR #685, opened 2026-06-18
Add Cmd/Ctrl+T project quick-switch palette

BnayaZil:
## What

A VSCode-style **Cmd/Ctrl+T** command palette to jump to a project **by name** instead of by number. Type part of a project name, results are fuzzy-matched and ranked by relevance, and **Enter** jumps to the top match (↑↓ to pick another, **Esc** to close).

Previously the only keyboard way to switch projects was `Cmd+1..9` (by index) — you had to remember each project's position.

## How

- New pure `src/mainview/utils/fuzzyMatch.ts` — subsequence matcher with fzf-style scoring (prefix / word-boundary / consecutive-run bonuses, gap penalties) + matched-index highlights. Fully unit-tested.
- New `src/mainview/components/ProjectQuickSwitchModal.tsx` — the palette overlay, modeled on the existing `TaskSwitcherOverlay` (portal, in-app React, works in both desktop and remote/browser mode — no native dialogs).
- `App.tsx`: `Cmd/Ctrl+T` handler added to the `useGlobalShortcut` chain; extracted a shared `navigateToProject()` helper so the palette and the `Cmd+1..9` index shortcuts share **identical** view-preservation behavior.
- i18n for en/ru/es, one "Did you know?" feature tip, unit + component tests.

## ❓ Open question for @h0x91b

The current scope is **projects only**. The original request floated extending the same palette to **navigate to a specific task** (not just a project) in the future — i.e. fuzzy-match across tasks too, possibly in one unified palette.

Do you want to:
1. Keep this projects-only (as shipped here), and treat task-jump as a separate follow-up?
2. Evolve this into a unified palette (projects + tasks) — e.g. a prefix/section split, or scoped modes?

Happy to follow up in a separate PR once we align on the model. The `fuzzyMatch` util is generic and already reusable for task titles.

## Test instructions

1. Press **Cmd+T** (macOS) / **Ctrl+T** (Linux) anywhere in the app → palette opens, input auto-focused.
2. Type part of a project name (e.g. `users`) → list filters, matched chars highlighted, best match selected.
3. Press **Enter** → jumps to that project's board (preserving task/board view like Cmd+1..9 does).
4. Use **↑↓** to move the selection, **Enter** to open; **Esc** or click-outside closes.
5. Type gibberish → "No matching projects" empty state.


BnayaZil:
## 🎥 Demo — Cmd/Ctrl+T project quick-switch

Captured live in browser mode (`dev3 remote` + headless Chrome driving the real UI with real projects).

### Walkthrough (animated)

Cmd+T quick-switch demo

▶️ Full-quality video (MP4): [cmdt-demo.mp4](https://raw.githubusercontent.com/BnayaZil/dev-3.0/refs/heads/demo/cmdt-quick-switch/demo/cmdt-demo.mp4)

### Key frames

**Cmd+T opens the palette** — all projects, each with its `⌘1..9` hint:

palette open

**Type `users`** → fuzzy-filters to `users` + `users-fed-monorepo`, matched prefix highlighted, top match selected (Enter jumps there):

typed users

**↓ moves the selection** to the second match:

arrow down

<sub>Assets live on a separate `demo/cmdt-quick-switch` branch of the fork, so they stay out of this PR's diff.</sub>

h0x91b:
Nice feature — we keep coming back to the idea of a quick-switch palette, so glad to finally see it land. The code is clean: semantic tokens with no hardcoded colors, i18n across all three locales, solid a11y (`role=dialog/listbox/option`), no native dialogs, and no new toolbar buttons (keyboard-only — which neatly sidesteps our main button-creep anti-pattern). The matcher and modal tests are good.

Before merging — one required change, a scope split, and a documentation ask.

**🔴 Must-fix: change the hotkey `Cmd/Ctrl+T` → `Cmd/Ctrl+K`.**
`Cmd+T` is the universal "new tab" across browsers and terminals, and we have a live terminal underneath it (ghostty/tmux intercepts it). The muscle-memory conflict will bite constantly. `Cmd+K` is free, and across the industry (Slack/Linear/Notion/Vercel) it already means "go to anything / switcher" — the right key for a navigation palette.

**Vision (re: the open question about tasks).** I want to grow this into a full palette following the VSCode/Linear model — two surfaces:
- **`Cmd+K`** — quick-switch / navigation: projects **and** tasks (fuzzy);
- **`Cmd+Shift+P`** — command palette for actions (create task, git, dev-server, etc.).

When we add tasks, reuse the existing matcher at **`src/mainview/utils/fuzzyMatch.ts`** — don't introduce a separate one. It should become our single matcher for short UI entities (BM25 in `src/shared/conversation-search-core.ts` stays for long transcripts only).

**UX documentation (important).** Since we're introducing a new surface — the palette — it needs to be captured in the UX docs before it drifts "off-manifest":
- add the surface (`command_palette` / quick-switch) to `docs/ux/ux-architecture.yaml` and `docs/ux/PRODUCT_UX_BIBLE.md`;
- add an entry to `docs/ux/UX_DECISIONS.md` — why the palette exists, and the **future vision**: it's not just project switching but an entry point that will, over time, absorb navigation (Cmd+K) and commands (Cmd+Shift+P) — task search, actions, and so on;
- run it through `/ux-principal` so placement/hotkeys/scope land canonically.

**Scope**, to keep the PR from sprawling:
- **in this PR** — the `Cmd+T → Cmd+K` hotkey change + the palette's UX-doc entries;
- **follow-up** — tasks in the palette (via `fuzzyMatch`) and the command palette on `Cmd+Shift+P`.

If that works — switch the hotkey, add the UX docs, and let's merge. Thanks!

h0x91b (changes requested):
Requesting changes — details in my comment above (hotkey `Cmd+T → Cmd+K`, plus UX-doc entries before merge).

BnayaZil:
Thanks for the thorough review — all in-scope items addressed in `5ccc051c`.

**🔴 Hotkey `Cmd/Ctrl+T` → `Cmd/Ctrl+K`** ✅
Rebound in `App.tsx` (handler + comment explaining *why not T*: new-tab/terminal conflict). Updated the feature tip and changelog copy to Cmd+K too. Verified `Cmd+K` is otherwise unbound in the shortcut chain.

**📐 UX docs — palette registered as a new surface** ✅
- `ux-architecture.yaml`: new `surfaces.command_palette` (purpose, allowed/forbidden, hotkey, `future:` block) + a `navigation.command_palette` mechanism entry + an `open_questions` item on keeping navigation vs. action-running separated. Bumped `updated_at`.
- `PRODUCT_UX_BIBLE.md`: added the palette to the surface table and a navigation subsection, with the "not the task switcher / not Cmd+T" distinctions.
- `UX_DECISIONS.md`: new dated entry capturing the decision, the Cmd+K rationale, and the **future vision** — Cmd+K grows to fuzzy-search **tasks** (reusing `utils/fuzzyMatch.ts` as the single short-entity matcher; BM25 stays for transcripts), and **Cmd+Shift+P** as the sibling action palette.

> Note: `/ux-principal` isn't installed in my implementing environment, so I authored the entries by hand against the existing manifest format — worth a pass through the skill on your side if you want placement re-validated canonically.

**🔭 Scope split** ✅ — tasks-in-palette and the `Cmd+Shift+P` action palette are deferred to a follow-up PR, as agreed. The matcher is already generic enough to take task titles unchanged.

Full suite green (mainview 1460 + bun 1438), lint clean. The demo GIF/screenshots above still reflect the behavior — only the trigger key changed (Cmd+T → Cmd+K).

h0x91b (approved):
Tested locally — Cmd+K palette works as expected, terminal focus no longer conflicts. Hotkey rebind and UX docs all addressed. LGTM, thanks!

BnayaZil:
🚀

PR #704, opened 2026-06-20
Add Vimium-style task hint navigation on the board

ittaiz:
## What

Adds **Vimium-style task hint navigation** to the Kanban board. Press **`f`** (when not typing in a field/terminal) and a short letter "hint" badge appears on every visible task card; type the hint to jump straight to that task. `Backspace` edits the typed prefix, `Esc` cancels.

A hint behaves exactly like clicking the card: active task → opens its terminal (split/fullscreen per your setting), completed/cancelled/**To Do** → opens the task detail. One hint per task, as requested (not per-link Vimium).

## How

- **`utils/hintLabels.ts`** — Vimium-style hint-string generator (prefix-free, minimal length).
- **`components/TaskHintOverlay.tsx`** — scans `[data-task-id]` cards, assigns hints, renders badges via a portal, and owns the keyboard while active. An **occlusion test** (`elementFromPoint` at the badge anchor must resolve to the card) means hints never render over a modal/palette backdrop or a card clipped under a column/page header — this replaced an earlier, brittle "list every open modal" guard.
- **`App.tsx`** — a guarded bare-`f` activation in the existing global-shortcut handler; closes on route change; gates the Option/Ctrl+Tab task switcher while hint mode is active.
- Follows project conventions: design tokens (`--hint-*` in both themes), i18n (en/ru/es), keymap registry entry, a discovery tip, and a changelog.

## Verification

- Unit + component tests for the generator, overlay (render/filter/commit/backspace/dedup/detach/**occlusion**), todo→detail, and switcher-disable. Full suite green (1525 mainview + 1477 bun).
- Driven live in a browser via `dev3 remote` against an **isolated** seeded fixture (no real data touched): confirmed hints render one-per-card, committing navigates, `Esc`/search-box guards work, a To Do hint opens detail, and `f` over an open modal renders no hints. 0 console errors.

## Review

Hardened across three independent code-review passes (correctness/edge-cases, simplicity/conventions, and a second fresh-eyes round): fixed a StrictMode double-click, occlusion for modals/clipping, the To Do no-op, the task-switcher stacking, and detached-node clicks.

## Test instructions

1. Open a project **Kanban board**; with nothing focused, press **`f`** → letter badges on every visible card.
2. Type a badge → jumps to that task (active → terminal; To Do/completed/cancelled → detail modal).
3. `Esc` cancels; `Backspace` edits the prefix.
4. Focus "Search tasks…", press `f` → types `f`, no hints.
5. Open a modal, press `f` → no hints over it. Scroll a column under its header, press `f` → no phantom badges in the header.
6. Press `f` then Option+Tab (⌃Tab on Linux) → switcher does not open over hints.


h0x91b (changes requested):
Hi, this is Claude (the AI assistant working on this branch) — leaving one request-for-changes before approving.

**Hint badges lag behind the cards on scroll.** The overlay positions every badge in a fixed, viewport-anchored portal and recomputes the coordinates by bumping React state on the `scroll` event:

```ts
const bump = () => bumpReposition((x) => x + 1);
window.addEventListener("scroll", bump, { capture: true, passive: true });
...
style={{ top: rect.top + 4, left: rect.left + 4 }}
```

The browser repaints the scrolled content immediately, but the React re-render that repositions the badge lands a frame or two later — so the badges visibly "chase" their cards. It reads as janky / unpolished.

**Requested change:** move the per-scroll repositioning off the React render path and make it synchronous. Keep a `ref` to each badge element and write `el.style.top/left` directly inside the `scroll`/`resize` handler (read fresh `getBoundingClientRect()` there), instead of triggering a state-driven re-render. Updating the DOM synchronously in the scroll handler tracks the cards tightly and kills the lag.

Notes for whoever picks this up:
- A pure-CSS route (CSS anchor positioning) would be ideal but isn't reliable in WKWebView yet, so synchronous JS is the pragmatic fix.
- Portaling the badge into the card element would get clipped by the column's `overflow`, which is exactly why the fixed overlay exists — so don't go that way.
- While here, consider re-scanning targets on scroll/resize too: right now only cards visible at activation get hints, so cards below the fold never receive one even after you scroll to them.

Everything else looks solid — algorithm is correct and prefix-free, tests are thorough, conventions (tokens, i18n×3, keymap, tip, changelog) all followed. Just want the scroll polish before merge.

h0x91b (approved):
Claude here again (the AI assistant). I went ahead and implemented the requested fix myself in 92a53fd4.

The scroll lag is gone: badge repositioning no longer goes through React state. Each badge registers its DOM node via a ref, and the scroll/resize handler writes `style.top/left` directly (synchronously, before paint) so the badges track their cards tightly instead of chasing them a frame behind.

- Added decision record 075 explaining why synchronous JS won over CSS anchor positioning (not supported in WKWebView) and over portaling into the card (clipped by column `overflow`).
- Added a test asserting a badge repositions on a scroll event with no re-render.
- Deferred the "re-scan targets on scroll for below-the-fold cards" idea to a separate change — doing it mid-session would reshuffle the prefix-free hints while the user is typing.

Full suite green (mainview + bun), lint clean. Approving.

PR #709, opened 2026-06-22
feat: sort Cmd+K quick-switch by recently jumped projects

BnayaZil:
## What

The **Cmd/Ctrl+K project quick-switch** palette now lists the projects you jumped to **most recently first** (MRU order), then the rest in board order. With an empty query the recency order applies; as soon as you type, normal fuzzy ranking takes over.

## How

- New `src/mainview/utils/recentProjects.ts` — a small `localStorage`-backed LRU (`dev3-recent-projects-v1`, capped at 16 ids). `recordProjectJump()` moves a project to the front; `orderByRecency()` is a pure reorder helper.
- Recency is recorded **centrally** in a single `commitNavigation()` chokepoint in `App.tsx`: any navigation that lands on a project records a jump. So **every** entry point is covered automatically and can't silently regress — the palette, `Cmd+1..9`, `Cmd+Shift+1..9`, the `g`-prefix go-to, **Dashboard card clicks**, terminal toggles, and any future one. Deferred (unsaved-changes) navigations record on confirm.
- A pure `projectIdForRoute()` helper lives in `state.ts` (unit-tested).
- The **⌘1-9 badge** is keyed to each project's stable **board index** (`shortcutIndexById`), not its display row — so it keeps matching the global index shortcuts after the rows are reordered.

## Demo — all 5 entry points

Recorded by driving the **real web mode** (`dev3 remote`) with Playwright against 16 real projects (no mocks). Each path opens a different project, then `Cmd+K` confirms it jumped to the top:

Cmd+K recency — all paths

<sub>([MP4 version](https://raw.githubusercontent.com/BnayaZil/dev-3.0/pr-demos/cmdk-allpaths.mp4))</sub>

Automated verification from the run:

```
PASS  ① Dashboard card click  → accomplish      on top
PASS  ② Cmd+K palette select   → thunderbolt     on top
PASS  ③ Cmd+7  (index)         → ai-exam-editor⌘7 on top
PASS  ④ Cmd+Shift+2 (mirror)   → wixos⌘2          on top
PASS  ⑤ g then 9 (go-to)       → dev-3.0⌘9        on top

FINAL MRU stack: dev-3.0⌘9 · wixos⌘2 · ai-exam-editor⌘7 · thunderbolt · accomplish · identity⌘1
```

Note the ⌘N badges stay pinned to the board index even though the rows are recency-sorted.

## Tests

- `recentProjects.test.ts` — LRU storage (front-insert, dedup, cap, corrupt-data) + `orderByRecency`.
- `state.test.ts` — `projectIdForRoute` across every screen.
- `ProjectQuickSwitchModal.test.tsx` — badge follows board index, not display row.
- Full suite green (3035 tests), `bun run lint` clean.


h0x91b (changes requested):
Hi, this is Claude (the AI assistant working on this branch) — did a review pass. Overall the change is clean and safe: pure `orderByRecency`, defensive localStorage handling, versioned `-v1` key, solid edge-case tests (corrupt JSON, non-string filter, cap-16, dedupe), and the `⌘N` badge correctly stays keyed to the board index so it keeps matching the `Cmd+1..9` shortcuts. Lint is clean and the tests pass. Just one thing I'd like addressed before merge.

### 🟡 Please fix: recency coverage is incomplete (and the doc-comment over-claims)

`recordProjectJump` is only called from `navigateToProject` and `goToProjectView`, so two common ways of opening a project never get recorded:

1. **Clicking a project card on the Dashboard** — `ActivityOverview.tsx:50` (`openProject`) calls `navigate({ screen: "project", projectId })` directly, bypassing `recordProjectJump`. This is probably the most common way users open a project, so the MRU list ends up ignoring it — open `Cmd+K` and the project you've been clicking all day isn't at the top, which feels arbitrary.
2. **`Cmd+Shift+1..9`** — `App.tsx:569` calls `navigate(...)` directly instead of going through `navigateToProject`, so it isn't recorded, even though its mirror `Cmd+1..9` (`App.tsx:587`) *is*. Inconsistent.

Because of this, the doc-comment in `recentProjects.ts` ("A jump = navigating to a project via the palette, Cmd+1..9, or the `g`-prefix go-to sequence") is inaccurate — it silently omits both paths above.

**Suggested fix (either is fine):**
- Route the Dashboard click and `Cmd+Shift+1..9` through `recordProjectJump` (ideally funnel them through a shared point so future navigation entry points are covered automatically), **or**
- If the limited scope is intentional, narrow the doc-comment to honestly state exactly which sources are tracked.

Everything else looks good to ship once this is sorted.

h0x91b (changes requested):
small issue above

BnayaZil:
Addressed in `114bcd7f` (rebased on the latest `main`).

Went with the **centralize** option rather than narrowing the doc-comment:

- New `commitNavigation()` chokepoint in `App.tsx` records a jump whenever the destination route lands on a project. Every entry point now funnels through it — Dashboard card click, `Cmd+1..9`, `Cmd+Shift+1..9`, the palette, the `g`-prefix go-to, the `Cmd+`` ` terminal toggle, and any future one — so coverage can't silently regress again.
- The deferred (unsaved-changes) navigations also route through `commitNavigation()` on confirm, so they record on actual commit (and not if cancelled).
- Removed the now-redundant explicit `recordProjectJump` calls from `navigateToProject` / `goToProjectView`.
- Extracted a pure `projectIdForRoute()` into `state.ts` (unit-tested) and deduped the local `getProjectIdForRoute`.
- Rewrote the `recentProjects.ts` doc-comment to describe the real behavior (any project navigation, recorded centrally).

Lint clean, full suite green (3035 tests).

h0x91b (approved):
Claude here again (the AI assistant) — re-reviewed after the fix. This nails it, and then some. 🎯

The new `commitNavigation()` chokepoint is exactly the right shape: every navigation that lands on a project now funnels through one place and records the jump, so the Dashboard card click and `Cmd+Shift+1..9` are covered — along with any future entry point, for free. Nice touches I appreciated:

- Extracting the pure `projectIdForRoute()` into `state.ts` with its own unit tests, instead of duplicating the route-switch.
- Routing the deferred (unsaved-changes) discard/save navigations through `commitNavigation` too, so they don't silently bypass recency.
- Dropping the now-redundant per-call `recordProjectJump` from `navigateToProject` / `goToProjectView` (no double-counting).
- Fixing the doc-comment to match reality, and moving the changelog to 06/23 for the merge date.

I checked out the PR head locally: `bun run lint` is clean and the affected tests pass (90/90 across `state`, `recentProjects`, and `ProjectQuickSwitchModal`). LGTM — approving. 🚀

PR #766, opened 2026-06-30
Run coding agents against a governed LLM backend (Amazon Bedrock)

arditti:
## Why

Not every team can send code to the Anthropic API directly. Enterprises that route through their own cloud — for **data residency, compliance, consolidated billing, or an existing AWS contract** — need the Claude agent to talk to **Amazon Bedrock** instead.

Today dev3 always passes an Anthropic-API `--model` alias (e.g. `claude-opus-4-8[1m]`), which Bedrock rejects with a `400 ... provided model identifier` — so those users can't run the agent at all. This PR adds a first-class **LLM Provider backend**: an agent can point at Anthropic (default) or a governed provider, and dev3 injects the provider's enable flag + the mapped model and drops the native `--model` alias.

dev3 does **not** manage credentials or region — those stay in the operator's own agent config (shell env / `~/.claude/settings.json`). dev3 only chooses the backend and pins the model.

## Where (operational decision)

*Which* backend an agent uses is an ops/deployment choice, not a per-task one. So the provider is configured **per agent** in **Settings → Coding Agents** — each agent targets its own backend, and the toggle only appears for agents that have a registered backend (today: Claude → Anthropic / Amazon Bedrock).

Providers are declared in a **registry** (`src/shared/llm-provider.ts`), keyed by an `LLM_PROVIDER` id and bound to an agent via `agentCommand`. Adding a backend is one registry entry + i18n labels — no call site special-cases a provider. If the placement should differ (global, per-config, install-time prompt), it's a localized change to where the registry is read.

## Details

- **Model mapping**: known aliases → provider-native ids automatically (Bedrock: `<geo>.anthropic.<family>`, geo selectable `global`/`us`/`eu`/`apac`); unknown models are derived so dev3 always pins the model. A per-model override table handles region-pinned inference profiles, ARNs, or new models.

## Tests

- Model mapping + env injection, registry / per-agent option lists, `--model` omission under a provider.
- Settings UI: toggle, geo, override table (Manual/Revert).

`bun run lint` clean. Full suite green locally except 4 pre-existing `git-*` tests that fail only in this sandbox on `spawn git ENOENT` (unrelated to this change).

See `decisions/089-llm-provider-toggle.md` and `agent-support-matrix.md`.


h0x91b:
Hey Eitav — Claude here (the AI assistant working on this repo with Arseny). First off, thanks for this — a clean, registry-driven way to point agents at a governed backend is genuinely useful, and the whole thing is well-tested and documented (decision record + matrix + changelog + tip, nice). Lint is green and the targeted suites all pass on my end.

A couple of small things worth a quick fix, independent of Bedrock:

- The **"Did you know?" tip** copy (`tip.llmProvider.body`) says *"dev3 injects the model and region for you"*, but the design — and the hint text + tests — is that dev3 does **not** inject the AWS region; the customer's own Claude Code config owns credentials/region. Mind tweaking the tip wording so it doesn't imply region is handled for you?
- The **command preview** (`buildCommandPreview`) drops `--model` under Bedrock but doesn't surface the injected `CLAUDE_CODE_USE_BEDROCK` / `ANTHROPIC_MODEL`, so for a Bedrock agent the preview reads like it launches with no model at all. Might be worth folding the provider env into the preview's env line.

The one thing neither Arseny nor I can verify is the actual Bedrock runtime — we don't have a Bedrock account with Anthropic model access. Since you do, could you run a quick sanity check and confirm three things?

1. The mapped model ids carry the `[1m]` marker (e.g. `global.anthropic.claude-opus-4-8[1m]`), because all the default configs use it. Does Claude Code strip `[1m]` from `ANTHROPIC_MODEL` before calling Bedrock, or does the profile id need to be bare? Just want to make sure the default configs don't hit the same `400 ... provided model identifier` this PR fixes.
2. Do the derived ids (`<geo>.anthropic.<family>`, no date / `-vN:0` suffix) resolve out-of-the-box on your account, or does each row need a manual override? If it's the latter, we might soften the "maps automatically" wording.
3. Does the background/fast-model path work without pinning `ANTHROPIC_SMALL_FAST_MODEL`?

If you've already validated these on real Bedrock, a quick "yep, tested" is all we need.

One more thing: `main` has moved on and the branch now has **merge conflicts** (`mergeable: CONFLICTING`), so it can't be merged as-is. Could you rebase on the latest `main`? The conflicting files are:

- `src/bun/agents.ts`
- `src/bun/settings.ts`
- `src/shared/types.ts`
- `src/mainview/components/global-settings/AgentSettingsSection.tsx`
- `src/mainview/tips.ts`
- `src/mainview/i18n/translations/{en,ru,es}/settings.ts`
- `src/mainview/i18n/translations/{en,ru,es}/tips.ts`
- `src/bun/__tests__/agents.test.ts`

Most look like they'll be additive (i18n keys, new default agents/configs, appended tips) rather than logic clashes, so it should be a fairly mechanical rebase. Please run `bun run test` after resolving to make sure nothing drifted. Thanks again!

arditti:
Rebased on latest `main` — all conflicts resolved (the `agents.ts` merge composes the new statusline-settings injection with the provider options; tips re-added post-prune). Both review items addressed: the tip no longer implies region injection, and `buildCommandPreview` now folds the provider env (`CLAUDE_CODE_USE_BEDROCK` + pinned `ANTHROPIC_MODEL`, geo/override-aware) into the preview env line, with tests.

Verified on real Bedrock, answering the three open questions:

1. **`[1m]` works** — `ANTHROPIC_MODEL=global.anthropic.claude-opus-4-8[1m]` initialized as "Opus 4.8 (1M context) · Amazon Bedrock" and completed requests. No 400.
2. **Derived ids resolve out-of-the-box** — the `global.anthropic.<family>` inference profiles worked with no manual overrides ("maps automatically" wording stands).
3. **No `ANTHROPIC_SMALL_FAST_MODEL` pin needed** — the session ran fine without it.

One clarification on regions, since two different things are both called "region" here. The AWS region in the operator's own config (env / `~/.claude/settings.json`) is the **source region** — the control-plane entry point where the request is authenticated, billed, CloudTrail-logged, and where model access must be granted. dev3's geo selector only picks the **inference-profile prefix** (`global`/`us`/`eu`/`apac`), i.e. the **data-plane routing scope**: the set of destination regions Bedrock may route the actual inference to. The two must be compatible — a geographic profile like `eu.anthropic.*` is only invokable from an EU source region, so picking `eu` in dev3 with `us-east-1` credentials will fail at the Bedrock API. `global` (the default) routes from any commercial region, which is why it's the safe default. This split is exactly why dev3 pins only the model id and leaves region/credentials to the operator.

`bun run lint` clean; full suite green except the 4 pre-existing `git-* spawn ENOENT` sandbox failures (reproduced on clean `main`, unrelated).

PR #1108, opened 2026-07-24
Add Amazon Bedrock backend support for the Codex agent

arditti:
## Summary

- The Codex agent gets the same per-agent LLM-backend toggle Claude already has: **OpenAI (native) / Amazon Bedrock**, in Settings → Agents.
- The provider registry (`src/shared/llm-provider.ts`) now supports two model-delivery channels, driven entirely by the `ProviderDefinition`:
  - **env** (`modelEnv` set — Claude on Bedrock): inject `CLAUDE_CODE_USE_BEDROCK=1` + `ANTHROPIC_MODEL`, omit `--model` (unchanged behavior).
  - **flag** (`modelEnv` absent — Codex on Bedrock): rewrite `--model` to the mapped id and append `enableArgs` (`-c model_provider="amazon-bedrock"`).
- Model ids for **all** providers are now derived purely from the config alias (`gpt-5.6-sol` → `openai.gpt-5.6-sol`, `claude-x` → `<geo>.anthropic.claude-x`); the redundant `BEDROCK_FAMILY_SUFFIX` table was removed, so adding a new model never requires a registry edit. Per-model overrides in Settings remain the escape hatch.
- `LLM_PROVIDER.Bedrock` was renamed to `BedrockClaude` in code for clarity next to `BedrockCodex`; the persisted value stays `"bedrock"` so existing user settings are untouched.
- Details in `decisions/163-codex-bedrock-flag-delivered-provider.md`.

## Regions note

Regions are currently **not relevant for OpenAI on Bedrock**: only US regions are supported, there are no cross-region inference profiles (ids are flat `openai.<model>`), and the region lives in the endpoint via the user's own `~/.codex/config.toml` (`[model_providers.amazon-bedrock.aws] region`) — dev3 deliberately exposes no region control for Codex. **Once Bedrock offers OpenAI models in other regions, re-evaluate how dev3 should let users override the region** (e.g. a per-launch `-c model_providers.amazon-bedrock.aws.region=...` enableArg or a registry-driven selector).

## Validation

- Reproduced the failure first: `codex exec --model gpt-5.6-luna` on a Bedrock-backed codex → 404 "model does not exist".
- Confirmed the mapped ids live against Bedrock: `openai.gpt-5.6-luna`, `openai.gpt-5.6-terra`, `openai.gpt-5.5` (and `openai.gpt-5.6-sol` from a working setup) all answer with `-c model_provider="amazon-bedrock"`.
- Generated the launch command through the real launcher code path and executed it end-to-end — codex responded via Bedrock.
- Browser QA of Settings → Agents → Codex: toggle, Codex-specific hint, auto-mapped model table, no geo selector; setting reverted after QA.
- `bun run lint` + `bun run test` green (mainview 186, bun 194, cli 34 test files).

## Test plan

1. Settings → Agents → expand Codex → LLM Provider toggle shows OpenAI / Amazon Bedrock.
2. Select Amazon Bedrock → Codex-specific hint + model mapping table (`gpt-5.6-* → openai.gpt-5.6-*`, `gpt-5.5 → openai.gpt-5.5`), no inference-profile region row.
3. Command preview shows `--model openai.<model> -c 'model_provider="amazon-bedrock"'`; switching back to OpenAI restores the plain alias.
4. Editing a mapping row shows MANUAL + Revert and the preview uses the override.
5. Launching a Codex task with Bedrock selected starts and answers via Bedrock.
6. Claude's provider section (toggle, geo selector, model table) is unchanged.


arditti:
Merged `main` (v1.40.0) to resolve the conflict in `src/shared/llm-provider.ts`: kept this branch's removal of `BEDROCK_FAMILY_SUFFIX` (the Bedrock id is now derived directly in `mapFamily`, so #1112's Opus 5 entry is subsumed — `opus` → `global.anthropic.claude-opus-5` with no registry edit), and updated main's new Opus 5 test to the renamed `LLM_PROVIDER.BedrockClaude` constant.

Heads-up: while updating I found that `main` itself currently fails 2 `tmux-audit` tests (reproduced on a clean `origin/main` checkout) — #1111 added `src/bun/native-terminal-adapter/` without audit overrides for its comment-only tmux references, and `inventory.json` was last regenerated in #1089. Since that would turn this PR's CI red on the cli shard, I included a small ride-along fix (`459379d8`): 4 overrides in `audit.config.json` following the existing MIG-006 'asserts non-use' / MIG-001 parity-runner precedents, plus regenerated `inventory.json`/`inventory.md`. Happy to split it into its own PR if you'd rather keep this one Bedrock-only.

Full suite green locally (mainview 2935, bun 3238, cli 600).

arditti:
Update: main has since reverted #1111 (#1113), which removes the tmux-audit breakage at the source — so I reverted the ride-along audit-fix commit (`602e86eb`) and merged the new main. The PR diff is back to Bedrock-only scope. Full suite green locally against current main.

h0x91b:
Hey, Claude here — the AI assistant that just did a review pass on this branch for Arseny. First off: nice work — the two-channel `ProviderDefinition` generalization (env vs flag delivery) is clean, the `BEDROCK_FAMILY_SUFFIX` removal checks out line-by-line (every entry equaled the derived fallback), and I verified your merge resolution too: the diff vs current main is byte-identical to the pre-merge Bedrock scope, and Opus 5 is correctly subsumed by the derived mapping (`opus` → `global.anthropic.claude-opus-5`, with test coverage). Lint + the provider test files are green on my side as well.

I did find one real (branch-introduced) UI bug and a few smaller things. All of these were re-verified against the current head (`721f9477`), and the bug was confirmed with an actual component render, not just code reading.

### Bug: `ProviderSelector` lacks the `agentCommand` guard

`AgentSettingsSection.tsx` (~line 1417): `const activeDef = getProviderDefinition(provider)` doesn't check `def.agentCommand` against the agent's base command — unlike the two sibling call sites that do (`agentProvider` in `agents.ts:653`, and `buildCommandPreview` in `utils.ts:137`).

Repro: Settings → Coding Agents → expand **Claude** → select Amazon Bedrock → edit Base command to `codex`. Verified by rendering the component:

```
toggle buttons rendered: yes, none active ("bedrock" is not among codex's options)
hint shown: Claude's Bedrock hint (not the Codex one)
geo selector: rendered (Claude-only feature, on a codex-command agent)
edits persist to: providerConfig["bedrock"]  ← orphaned key; the launcher's
  agentProvider() guard rejects the stale id, so launches silently go native
```

Before this branch the mismatch was invisible (`providersForAgent("codex")` was `[]`, so the whole panel hid); now that codex has a backend, the stale panel renders. The reverse direction (Codex+Bedrock → base command `claude`) misbehaves the same way, there the model table would even suggest `openai.claude-opus-5[1m]`. Fix is essentially one line — the same guard as `utils.ts:137` — and/or resetting `llmProvider` when `baseCommand` changes.

### Smaller points

1. **`providerArgs()` is wired into the codex adapter only.** `modelArgs()` is called by all seven adapters; `providerArgs()` by exactly one. The module header promises "no call site needs to special-case it — the registry drives env/args injection", but a future backend with `enableArgs` for e.g. gemini would silently drop them. Either call it in every adapter next to `modelArgs`, or add a test pinning that promise.
2. **The `applyProviderModel` wiring is untested end-to-end.** The function itself is well covered, but both call sites (`agents.ts:644` and `:735` — the duplicated `applyModelOverride(applyProviderModel(...))` composition) are only reached through `resolveCommandForAgent`/`ForProject`, which the handler tests mock entirely. The golden case `codex/provider-bedrock` also feeds a pre-mapped `openai.gpt-5.6-sol`, so the alias rewrite itself isn't pinned anywhere. A small helper composing both applications would collapse the duplication and give one seam to test.
3. **`isThirdPartyProvider` is now dead code** — after this branch its only production reference is its own definition (checked src/, scripts/, docs/); everything else is tests. Repo policy is no dead exports — either delete it (with its test) or use it.
4. **Naming nit:** the codex native option persists `llmProvider: "anthropic"`. You already did the right dance for `BedrockClaude` (rename the constant, keep the wire value) — `Native: "anthropic"` would apply the same trick here.
5. **Doc drift:** `mapModelForProvider`'s docstring still says "Known families use the exact mapped id" (the table is gone) and "Claude Code"; `ProviderSelector`'s docstring still says the provider "drops dev3's --model alias" (now env-channel-only). The `tip.llmProvider` tip ("Run Claude on Amazon Bedrock") is also narrower than the feature now is.

Only the `ProviderSelector` guard feels worth fixing before merge; the rest can ride along or follow up. Thanks again for the thorough decision record (163) — the live-endpoint validation of the flat `openai.<family>` ids and the US-only region caveat made this review much easier.

src/bun/agents.ts
```diff
@@ -639,7 +641,7 @@ export async function resolveCommandForAgent(
 	await applyCodexAccountEnv(baseCmd, extraEnv, options?.accountId);
 	const command = resolveAgentCommand(
 		agentWithPath,
-		applyModelOverride(config, baseCmd, extraEnv),
+		applyModelOverride(applyProviderModel(config, agentWithPath), baseCmd, extraEnv),
```
h0x91b: **medium:** this composition — `applyModelOverride(applyProviderModel(config, agentWithPath), baseCmd, extraEnv)` — is duplicated verbatim at line 735, and neither call site is covered by tests: `resolveCommandForAgent`/`ForProject` are mocked entirely in the handler tests, and the golden case `codex/provider-bedrock` feeds a pre-mapped `openai.gpt-5.6-sol`, so the alias rewrite itself isn't pinned anywhere. If someone drops `applyProviderModel` from one of the two sites, every test stays green.

Suggestion: a small helper (e.g. `resolveLaunchConfig(config, agent, baseCmd, extraEnv)`) composing both applications — collapses the duplication and gives one seam to unit-test.

src/shared/agent-adapters/common.ts
```diff
@@ -14,3 +14,9 @@ export function modelArgs(
 	// `claude-opus-4-8[1m]`). Quote them so zsh doesn't glob-expand.
 	return ["--model", quoteIfUnsafe(config.model)];
 }
+
+/** The active third-party backend's routing args (shell-quoted), e.g. Codex's
+ *  `-c model_provider="amazon-bedrock"`. Empty for the native default. */
+export function providerArgs(options?: AdapterLaunchOptions): string[] {
```
h0x91b, line 20: **medium:** `providerArgs()` is wired into the codex adapter only — `modelArgs()` is called by all seven adapters, this one by exactly one. The module header of `llm-provider.ts` promises "no call site needs to special-case it — the registry drives env/args injection", but a future backend with `enableArgs` for e.g. gemini would silently drop them (no error, args just vanish).

Either call `providerArgs(options)` in every adapter right after `modelArgs`, or add a test pinning that registry promise so the gap fails loudly.

src/shared/llm-provider.ts
```diff
@@ -168,11 +188,38 @@ export function getProviderDefinition(provider: LlmProvider | undefined): Provid
 	return provider ? PROVIDER_REGISTRY[provider] : undefined;
 }
 
-/** True when the provider is a third-party backend (model comes from env, not --model). */
+/** True when the provider is a third-party backend (dev3 pins the model id). */
 export function isThirdPartyProvider(provider: LlmProvider | undefined): boolean {
```
h0x91b: **medium:** `isThirdPartyProvider` is dead code after this PR — its only production reference is this definition (checked src/, scripts/, docs/); everything else is tests. Both former callers moved to `providerOmitsModelFlag`/`getProviderDefinition`. Repo policy is no dead exports: either delete it (with its test block) or use it.

src/shared/types.ts
```diff
@@ -798,7 +798,14 @@ export interface PxpipeProxyStatus {
...
 	Anthropic: "anthropic",
-	Bedrock: "bedrock",
+	/** Amazon Bedrock as the backend for the Claude agent. The stored value
+	 *  stays the historical `"bedrock"` — it is persisted in users' settings
+	 *  (agent.llmProvider + providerConfig keys), so renaming it would silently
+	 *  reset existing Bedrock setups. */
+	BedrockClaude: "bedrock",
+	/** Amazon Bedrock as the backend for the Codex agent. A separate id because
+	 *  the registry pairs each id with one agent command. */
+	BedrockCodex: "bedrock-codex",
```
h0x91b, line 812: optional (non-blocking): with codex on board, its native backend now persists `llmProvider: "anthropic"`, which reads oddly. You already did exactly the right dance one line up for `BedrockClaude` (rename the constant, keep the wire value) — `Native: "anthropic"` would apply the same trick to the default sentinel.

src/mainview/components/global-settings/utils.ts
```diff
@@ -131,15 +131,27 @@ export function buildCommandPreview(
...
+	// Env-delivering backends (Claude on Bedrock) omit --model — the alias would
+	// be rejected with a 400; the pinned model rides ANTHROPIC_MODEL instead.
+	// Flag-delivering backends (Codex on Bedrock) show the rewritten pinned id.
+	const modelViaEnv = providerDef?.modelEnv != null;
+	const previewModel =
+		providerDef && !modelViaEnv && config.model
+			? (providerPinnedModel(llmProvider, providerConfig, config.model) ?? config.model)
+			: config.model;
+
+	if (previewModel && !modelViaEnv) {
```
h0x91b: optional (non-blocking): `if (previewModel && !modelViaEnv)` reads oddly since `previewModel` is populated even when `modelViaEnv` is true (the ternary only rewrites for the flag channel). `if (!modelViaEnv && previewModel)` — or folding the env-channel case into the `previewModel` computation — would make the intent ("env channel → no --model at all") more obvious.

src/shared/agent-adapters/codex.ts
```diff
@@ -42,6 +42,9 @@ export const codexAdapter: AgentAdapter = {
 		// Resume uses a subcommand (assembled at the end). Codex ignores dev3's
 		// generic permission-mode / effort / budget flags.
 		args.push(...modelArgs(config, options));
+		// Backend routing (e.g. Bedrock's `-c model_provider=...`) before the
+		// config's additionalArgs, so an explicit user override still wins.
+		args.push(...providerArgs(options));
```
h0x91b, line 48: optional (non-blocking): two assumptions here I couldn't verify locally, flagging as hypotheses:
1. "an explicit user override still wins" relies on codex's last-wins semantics for repeated `-c` overrides — decision 163 states it too, but nothing pins it; if codex ever changes that, the user's `additionalArgs` override silently loses.
2. All codex presets carry `-c model_reasoning_effort="xhigh"/"ultra"` — whether Bedrock's OpenAI surface accepts those values isn't covered by decision 163's live validation; if it rejects them, e.g. the Sol Ultra presets fail with an opaque error under this backend.

src/mainview/i18n/translations/en/settings.ts
```diff
@@ -98,11 +98,13 @@ const settings = {
...
 	"settings.defaultConfigDesc": "Configuration applied when launching tasks",
 	"settings.llmProvider": "LLM Provider",
-	"settings.llmProviderDesc": "Backend for this agent. A third-party provider injects its model/region and drops dev3's --model flag.",
+	"settings.llmProviderDesc": "Backend for this agent. A third-party provider pins its own model ids in place of dev3's aliases.",
 	"settings.providerNative": "Native",
 	"settings.providerAnthropic": "Anthropic",
+	"settings.providerOpenAI": "OpenAI",
 	"settings.providerBedrock": "Amazon Bedrock",
 	"settings.providerBedrockHint": "Credentials and AWS region come from your own Claude Code setup. Pick the inference-profile region below; model ids map automatically — edit a row to pin a specific profile or ARN.",
+	"settings.providerBedrockCodexHint": "Credentials and AWS region come from your own Codex setup (~/.codex/config.toml, [model_providers.amazon-bedrock]). Model ids map automatically — edit a row to pin a specific id.",
```
h0x91b, line 107: optional (non-blocking): the hint is the only guard against a missing `[model_providers.amazon-bedrock]` section in `~/.codex/config.toml` — a user who flips the toggle without it gets a raw codex error at launch. A cheap preflight (warn in the settings panel if the section is absent) would fail this much friendlier. Fine as a follow-up.

src/shared/llm-provider.ts
```diff
@@ -95,30 +109,35 @@ export interface ProviderDefinition {
...
+		modelEnv: "ANTHROPIC_MODEL",
 		usesGeo: true,
-		// Known families map to `<geo>.anthropic.<family>` inference profiles;
-		// unknown/new families are derived from the alias (always pinned).
-		mapFamily: (family, geo) => `${geo}.${BEDROCK_FAMILY_SUFFIX[family] ?? `anthropic.${family}`}`,
+		// `<geo>.anthropic.<family>` inference profiles, fully derived from the
+		// alias — new models need no registry edit (always pinned).
+		mapFamily: (family, geo) => `${geo}.anthropic.${family}`,
+	},
+	[LLM_PROVIDER.BedrockCodex]: {
```
h0x91b, line 129: optional (non-blocking): the "dev3 ALWAYS pins the model" invariant isn't actually guaranteed on the flag channel — a config with no `model` emits no `--model`, so codex falls back to its own default (control/data plane divergence, the exact thing the header warns about). All built-in codex configs set a model, so it's theoretical today; worth a note here or a guard if custom configs make it reachable.

h0x91b (changes requested):
Claude here again (the AI assistant reviewing for Arseny) — formalizing my earlier comment as a review. Requesting changes for exactly one thing; everything else below and inline is either medium (worth a look) or explicitly optional/non-blocking.

### Blocking: `ProviderSelector` is missing the `agentCommand` guard

(Filed in the review body because `AgentSettingsSection.tsx` itself isn't part of this diff — only its test is.)

`AgentSettingsSection.tsx` ~1417: `const activeDef = getProviderDefinition(provider)` doesn't check `def.agentCommand` against the agent's base command, unlike the launcher (`agentProvider`, `agents.ts:653`) and the preview (`utils.ts:137`), which both grew that guard in this PR.

Repro (confirmed by rendering the component): Settings → Coding Agents → expand **Claude** → select Amazon Bedrock → edit Base command to `codex`:
- the toggle renders with **no active option** (`"bedrock"` isn't among codex's options),
- Claude's Bedrock hint + the geo selector (Claude-only) render under a codex-command agent,
- edits persist into `providerConfig["bedrock"]` — an orphaned key: the launcher's `agentProvider()` rejects the stale id, so launches silently go native while the UI looks configured.

Before this PR the mismatch was invisible (`providersForAgent("codex")` was `[]` → the panel hid); now that codex has a backend, the stale panel renders. The reverse direction (Codex+Bedrock → base command `claude`) misbehaves symmetrically. Fix is essentially the same one-line guard as `utils.ts:137`, and/or resetting `llmProvider` when `baseCommand` changes.

### Optional / non-blocking (no good inline anchor)

- Doc drift: `mapModelForProvider`'s docstring still says "Known families use the exact mapped id" (the table is gone) and "Claude Code"; `ProviderSelector`'s docstring still says the provider "drops dev3's --model alias" (now env-channel-only).
- The `tip.llmProvider` tip ("Run Claude on Amazon Bedrock") is narrower than the feature now is — per the tips-hygiene rule it'd be reworded in the same change that widens the feature.
- A versioned binary path (e.g. `/opt/homebrew/bin/codex-0.145`) silently disables the backend since `agentKey` compares the last path segment verbatim. Pre-existing behavior, just more reachable with two backends.

Everything re-verified against this head (721f9477): the merge with main is clean — the PR diff vs current main is byte-identical to the pre-merge Bedrock scope, and Opus 5 is subsumed by the derived mapping with test coverage. Nice work overall — the two-channel registry design is the right shape.

src/bun/agents.ts
arditti: Fixed in ded899c7/160188b7 — extracted `resolveLaunchConfig(config, agent, baseCmd, extraEnv)` composing both applications; both call sites now go through it, and unit tests pin the seam including the alias rewrite itself (`gpt-5.6-sol` → `openai.gpt-5.6-sol`, not a pre-mapped input).

src/shared/agent-adapters/common.ts
arditti: Went with the first option: every adapter now calls `providerArgs(options)` right after `modelArgs`, and a per-adapter test pins the registry promise (`launchArgs — provider routing args` in agent-adapters.test.ts), so a future backend's enableArgs can't be silently dropped.

src/shared/llm-provider.ts
arditti: Deleted `isThirdPartyProvider` and its test block — confirmed no production references remained.

src/shared/types.ts
arditti: Renamed to `LLM_PROVIDER.Native` (wire value stays `"anthropic"`), with a doc comment explaining the frozen sentinel — same trick as BedrockClaude.

src/mainview/components/global-settings/utils.ts
arditti: Reordered to `if (!modelViaEnv && previewModel)` and added a one-line comment stating the env-channel intent.

src/shared/agent-adapters/codex.ts
arditti: Both hypotheses addressed as far as locally verifiable: (1) the providerArgs-before-additionalArgs ordering is now pinned by an adapter test, and decision 163 states explicitly that the override behavior rides codex's last-wins semantics (observed on 0.145, not contractual); (2) decision 163 now scopes the live validation to model routing and flags `model_reasoning_effort="xhigh"/"ultra"` on Bedrock's OpenAI surface as unverified — needs a live Bedrock check to close.

src/mainview/i18n/translations/en/settings.ts
arditti: Implemented the preflight (ded899c7): new `checkCodexBedrockConfig` RPC reads ~/.codex/config.toml and the settings panel shows a danger-styled warning under the Bedrock hint when the `[model_providers.amazon-bedrock]` section is absent (bare/quoted/subtable headers all detected; pure check unit-tested, both UI states component-tested, en/ru/es strings).

src/shared/llm-provider.ts
arditti: Noted in the registry entry: a config with no `model` emits no `--model`, so codex falls back to its own default — theoretical today since all built-in codex configs set one. Left as a comment rather than a guard to avoid inventing behavior for a config shape that doesn't exist yet.

arditti:
All review items addressed (160188b7 + ded899c7), including the optionals:

**Blocking — ProviderSelector guard:** confirmed the repro exactly as described (component test reproduces it red before the fix). `ProviderSelector` now applies the same `agentCommand` guard as `agentProvider`/the preview: a stale id renders no provider fields, `patchProvider` no-ops (no more orphaned `providerConfig` keys), and the toggle highlights the **native** option — matching what the launcher actually does with the stale id. Also QA'd in a live browser: Claude → Bedrock → base command `codex` now shows only the OpenAI/Bedrock toggle with OpenAI active; no Claude hint, geo, or model table; no console errors. The reverse direction is covered by the same guard.

**Body optionals:** `mapModelForProvider` + `ProviderSelector` docstrings updated (derived ids, env-vs-flag channels, agent-neutral wording); the Bedrock tip now reads "Run agents on Amazon Bedrock" / "points Claude or Codex…" in all three locales.

Per-thread details in the inline replies. Full suite green locally (mainview 2939, bun 3243, cli 606).

h0x91b (approved):
Claude here — validated the two follow-up commits (`160188b7`, `ded899c7`) against every point from my review, and everything checks out. Approving.

What I verified on this head:

- **Blocking bug fixed:** `ProviderSelector` now applies the same `agentCommand` guard as the launcher, and `effectiveProvider` falls back to Native so the toggle shows the launcher's actual behavior instead of a stale panel. Both new stale-id tests pin exactly the repro I filed (no provider fields, native option highlighted).
- **All three mediums addressed:** `providerArgs()` wired into all six adapters with an `it.each` registry-promise test plus an ordering test pinning the last-wins arrangement; `resolveLaunchConfig` collapses the duplicated pipeline into one tested seam; dead `isThirdPartyProvider` deleted with its test.
- **Optionals all taken too** (nice — none were required): `Native` constant with the persistence-comment trick, the codex config.toml preflight (pure `hasModelProviderSection` helper + RPC + UI warning, i18n in all three locales), decision 163 now documents both unverified assumptions, the flag-channel "always pins" caveat, doc drift and the tip reworded.
- **Green everywhere:** CI fully green (lint, 5 test shards, build-check, trivy); locally on `ded899c7` — `bun run lint` clean, `bun run test` 2939 + 3216 + 600 passing.

Resolved all eight review threads. Thanks for the fast and thorough turnaround — taking the optional items as well, with tests for each, was above and beyond.

PR #1167, opened 2026-07-28
Delete orphaned worktree directories instead of blocking task completion

sworgkh:
## Problem

Completing a task failed with:

```
Failed to remove worktree at .../<task>/worktree: fatal: '.../worktree' is not a working tree
```

The worktree directory still existed, but the repo's `.git/worktrees/<name>` metadata was gone (pruned or deleted by hand). `git worktree remove` can never clean that up, so teardown threw on every attempt and the task stayed in review forever with no way out from the UI.

## Fix

`removeWorktree` (`src/bun/git.ts`) now recognizes the orphan case: when `git worktree remove` fails, the path is not in `git worktree list`, and the directory's `.git` link points at a missing gitdir, dev3 deletes the directory directly and runs `git worktree prune`. A live or locked worktree still surfaces the original error — the dangling gitdir check is what keeps `rmSync` away from directories git still owns.

## Tests

`src/bun/__tests__/git-worktree.test.ts` — new regression test that removes `.git/worktrees/<name>` behind a real worktree and asserts teardown succeeds and the branch is deleted. Verified red before the fix, green after; full `bun run test` plus the excluded `git-worktree` file are green.

#skipreview

### DIFF
--- a/change-logs/2026/07/28/fix-orphaned-worktree-blocks-completion.md
+++ b/change-logs/2026/07/28/fix-orphaned-worktree-blocks-completion.md
@@ -1,0 +1,3 @@
+Short: Completing a task no longer stalls
+
+Completing a task no longer gets stuck when its worktree directory lost its git metadata: `git worktree remove` used to fail with "is not a working tree" and the task stayed in review forever. dev3 now detects the orphaned directory and deletes it directly.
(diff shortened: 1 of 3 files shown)


sworgkh:
Closing — this was investigation only, not intended to land.

PR #1169, opened 2026-07-28
Per-project environment variables for agent terminals

arditti:
## What

Adds a per-project environment-variables map that is exported into every task-scoped session: agent terminals, setup/dev/cleanup scripts, column agents, and spawned agent panes (both tmux and native backends).

## How

- New `env: Record<string,string>` field on `Dev3RepoConfig`/`Project`, riding the existing config cascade — editable in Project Settings, shareable via `.dev3/config.json`, and overridable in the gitignored `.dev3/config.local.json` for secrets.
- Unlike other config fields, `env` merges **per key** across cascade layers (worktree local > worktree repo > main local > main repo > projects.json), so a repo file setting one var can't erase UI/local vars. See `decisions/179-per-key-env-config-merge.md`.
- `resolveProjectEnv()` re-reads config at launch time; injected at the lowest precedence so `DEV3_*` lifecycle vars and per-agent-config `envVars` always win.
- Project Settings gains an "Environment Variables" dotenv-style textarea (KEY=value, `#` comments) with inline validation that blocks saving on invalid lines; server-side validation guards the three save handlers. `dev3 config show` prints env key names only (values may be secrets).

## Testing

- Unit: dotenv parse/serialize/sanitize/validate round-trips; per-key cascade merge with real config files; editor save/clear/error-state renderer tests.
- Full suite green (`bun run test`) + `bun run lint`; manual browser QA of the editor on the Project Config and Worktree Config tabs, including error state, save-to-file, and clearing a previously saved value.

### DIFF
--- a/change-logs/2026/07/31/feature-per-project-env-vars.md
+++ b/change-logs/2026/07/31/feature-per-project-env-vars.md
@@ -1,0 +1,3 @@
+Short: Per-project env vars for agents
+
+Project Settings now has an Environment Variables editor, also configurable in .dev3/config.json and .dev3/config.local.json with per-key merging. The variables are exported into the project terminal, every agent terminal, setup/dev/cleanup script, and column agent of the project's tasks. This field is not for secrets; the editor identifies whether its storage is local-only or committed.
--- a/decisions/179-per-key-env-config-merge.md
+++ b/decisions/179-per-key-env-config-merge.md
@@ -1,0 +1,38 @@
+# 179. Per-key merge for the `env` config field
+
+## Context
+
+Per-project environment variables must reach the standalone project terminal,
+every task-scoped agent terminal, setup/dev/cleanup script, column agent, and
+spawned agent pane. They can be set in Project Settings (`projects.json`),
+`.dev3/config.json` (shared), and `.dev3/config.local.json` (machine-local), but
+this field is not for secrets.
+
+## Decision
+
+`env` is the one field that merges PER KEY across all cascade layers (worktree
+local > worktree repo > main local > main repo > projects.json). Implemented as
+an explicit special case after the uniform loop in `applyConfigCascade`;
+launch-time consumers use `resolveProjectEnv()`, which re-reads the files so
+config edits apply on the next launch. Injection order in task sessions:
+project env first, so `DEV3_*` lifecycle vars and per-agent-config `envVars`
+always override it. The UI identifies whether the active storage is local-only
+or committed, and warns that saving worktree repo config commits immediately.
+
+## Risks
+
+Provenance surfaces (`dev3 config show`, UI source badges) attribute `env` to
+its highest-priority source even when the effective map mixes layers. Accepted:
+per-key provenance would complicate three surfaces for marginal value. Values
+remain plaintext in existing config storage, so the product must not present
+this field as secret storage.
+
+## Alternatives considered
+
+Whole-field resolution like every other key: rejected — a repo config defining
+one var would silently erase all UI-configured vars, defeating the "shared file
++ machine-local overrides + UI" use case. A separate dotenv file or `envFile`
+pointer was left out of v1 because it duplicates the existing config cascade
+and gitignore machinery. Keychain/encryption and a variable-name denylist were
+also left out: storage remains unchanged, and cloning a repository already
+trusts its committed configuration.
--- a/src/bun/__tests__/agent-accounts.test.ts
+++ b/src/bun/__tests__/agent-accounts.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { existsSync, lstatSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, lstatSync, mkdirSync, mkdtempSync, readFileSync, rmSync, unlinkSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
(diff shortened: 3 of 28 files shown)


h0x91b (changes requested):
Hey, Claude here — the AI assistant that reviewed this branch together with Arseny. We went through the design questions in detail and landed on a set of concrete changes. Nice work overall: the per-key merge is the right call, it's properly recorded in a decision record, the RPC boundary doesn't trust the client, `sanitizeEnvMap` can't kill the project list, and the test coverage is genuinely good. `bun run lint` and the full `bun run test` are green on this head.

Requesting changes for the items below. Five are required, three are strongly recommended and were agreed to as well.

## Design decisions we settled on (so you don't have to re-litigate them)

- **Storage stays as-is.** No new mechanism, no keychain, no encryption, no dotenv-file loading. The `env` field keeps riding the existing config cascade.
- **Secrets are handled by a UI warning, not by moving the data.** The field is declared "not for secrets", and the UI must say so loudly.
- **A committed `.dev3/config.json` setting env for agent panes is accepted.** If you cloned the repo, you already trust its authors. No denylist for `PATH` / `NODE_OPTIONS` / `*_BASE_URL`.
- **The Project tab showing the merged (cross-layer) env is accepted** — not worth fixing.

## Required

### 1. Warning banner in the env editor — different text per tab

The editor renders in three places with three different destinies for the value:

| Where | Writes to | In git? |
|---|---|---|
| Project tab | `~/.dev3.0/projects.json` | Never |
| Worktree → Repo subtab | `.dev3/config.json` | **Yes** |
| Worktree → Local subtab | `.dev3/config.local.json` | No (gitignored) |

A blanket "will be committed to git" is false on two of the three, and a banner that cries wolf gets ignored exactly where it matters. So:

- **Worktree → Repo**: prominent danger-toned banner — do not store secrets here, this file is committed to git, and **Save commits it immediately** (`ProjectSettings.tsx:1480` passes `autoCommit: true`, so `git add` + `git commit` happen on Save with no confirmation). That last part is worth spelling out; it's stronger than "might end up in git".
- **Project tab** and **Worktree → Local**: a calm line stating the value is stored locally and never reaches the repository.

All strings through `t()` with keys in `en` / `ru` / `es`. No native dialogs (remote/browser mode).

### 2. `src/shared/env-text.ts:6` — CRLF breaks every line

`.` in JS does not match `\r`, so `LINE_RE` rejects any line coming from a CRLF file:

```js
parseEnvText("FOO=bar\r\nBAZ=1") // → { env: { BAZ: "1" }, errors: [{ line: 1, text: "FOO=bar\r" }] }
```

A `.env` pasted from a Windows machine reports *every* line as invalid with Save disabled and no hint why. Strip a trailing `\r` per line before matching, and add a CRLF test. This is a hard requirement — Windows support is close.

### 3. `ProjectSettings.tsx:1897-1913` — subtab switch keeps stale editor state

The Repo/Local ternary renders the same component type at the same position with no `key`, so React reuses the instance and `EnvVarsEditor`'s `text` survives the switch. Meanwhile the parent effect (`:1133`) resets `envTextError` to `false` on `worktreeSubTab` change. Repro: type only an invalid line in Repo, switch to Local — the bad text is still on screen, Save is re-enabled, and the typed vars silently never persist. Fix: `key={worktreeSubTab}` on both `ConfigForm`s, plus a test.

### 4. `src/bun/agent-skills.ts:219-229` — `env` missing from the shipped skill schema

The schema table doesn't list `env`, and the line right under it says "**Only include these fields**" — so every agent that writes `.dev3/config.json` currently believes the key doesn't exist. Since the sanctioned team-sharing path *is* that file, and agents are the ones writing it, add the `env` row and note in it that the field is not for secrets.

### 5. Remove `docs/superpowers/**`, fold the durable parts into `decisions/`

This repo does not use the superpowers spec/plan pipeline, and `docs/` is the GitHub Pages root — 841 lines of process scaffolding in a new undocumented directory is noise. Please:

- `git rm -r docs/superpowers/`
- Move what outlives the branch into `decisions/179-per-key-env-config-merge.md`: the goals, what was deliberately left out of v1, the layer-precedence order, and the new "this field is not for secrets" policy.
- Also update the decision record and the changelog entry so they no longer describe the field as a safe place for secret/personal values.
- Move `change-logs/2026/07/28/feature-per-project-env-vars.md` to the expected merge date.

## Strongly recommended (agreed, please include)

### 6. `src/shared/env-text.ts:30` — a newline in a value bricks the editor

`serializeEnvText` doesn't escape `\n`, so a value containing one (written by hand into the JSON config, or arriving via RPC) round-trips into an invalid second line: `{ KEY: "a\nb" }` → parse reports line 2 invalid and Save is blocked forever, with only a line number as explanation and no way out from the UI. Reject `\n` in `validateEnvMap`, and render such keys read-only with a short explanation.

### 7. `src/bun/repo-config.ts:336-346` — auto-migration copies personal env into the committed file

`env` is in `DEV3_REPO_CONFIG_KEYS`, so `migrateProjectConfig` (which runs automatically on every project load, `rpc-handlers/app-handlers.ts:144`) copies whatever you typed in Project Settings into `.dev3/config.json`. Even setting secrets aside, that silently turns personal values into team values with no prompt and no banner. Also, a cleared editor persists `env: {}`, which flips `hasSettings` and creates a `.dev3/config.json` that would otherwise not exist. Please skip `env` in that loop (`if (key === "env") continue`) — an explicit `dev3 config export` may keep carrying it, since that one is a deliberate user action.

### 8. `src/bun/data.ts:535` — env values written to the log

`log.info("Updating project", { projectId, updates })` dumps the whole updates object, values included, into `~/.dev3.0/logs`. The rest of the branch is careful here (`tmux-pty.ts` logs `envKeys` only, the CLI prints key names only), so this one line stands out. Log `Object.keys(updates.env ?? {})` instead — cheap insurance against the user who didn't read the banner.

## Before re-requesting review

- `bun run lint` clean and the **full** `bun run test` green (not just the touched files — sibling suites cover the same components, and a rebase pulls in code your run never saw).
- New tests for the CRLF case, the subtab `key` fix, and the per-tab banner text.
- Manual browser QA of the editor — the banner is a visible surface, and screenshots in streamer mode (`&streamer=on`).

Thanks — this is close, and none of the above is architectural.

PR #1177, opened 2026-07-29
PR Babysitter phase 1: on-entry column agent, autonomy config, settings UI

arditti:
## What

Phase 1 of native PR babysitting: the built-in review-by-colleague ("PR Review") column becomes the active babysitting phase. When a task's open PR enters the column (auto-promotion or manual move), a babysitter agent launches in the task's tmux session and works the PR in strict order: mergeability/conflicts → CI → review feedback.

## How it works

- **Config**: new additive `babysitter` object (`autonomy`, per-capability `overrides`, `handleComments`) on `Project` + `.dev3/config.json`, next to `builtinColumnAgents["review-by-colleague"]` (agent/config/custom prompt). On-disk format is additive only.
- **Autonomy presets over 6 capabilities** (push / reply / resolve / rebase / rerunChecks / armAutoMerge): **Triage** = read-only (diagnosis into a task note, zero GitHub writes), **Fix** = push + reply + rebase, **Land** = everything incl. arming `gh pr merge --auto --squash --match-head-commit`. Sparse overrides on top; UI shows "Custom" when they diverge.
- **Ships enabled at Triage by default** (safe: zero GitHub writes); `off` is an explicit opt-out; Fix/Land are opt-in.
- **Enforcement is prompt composition** (`composeBabysitPrompt` in `src/shared/types.ts`): each capability emits a MAY/NOT line; hard ceilings are appended to every prompt (never merge/`--admin`/approve, never edit CI/tests/timeouts to go green, never touch drafts, `--force-with-lease` only after a legit rebase, reply before resolve, park via `dev3 task move --status user-questions`). See `decisions/179`.
- **Comment handling**: triages review submissions, inline threads, and conversation comments (Agree/Disagree/Already-fixed/Defer, one reply per comment). At reply-incapable levels comments are still monitored — replies are drafted into a task note, never posted; the UI toggle locks accordingly.
- **Executor**: `review-by-colleague` honored in `columnAgentConfig` (default `builtin-claude`/`claude-bypass-sonnet`, pane title from `customStatusLabels`, no `onExitCommand` — the task stays in the column; the prompt parks when a human is needed).
- **Bug fix found during e2e**: `cause`-carrying moves (PR auto-promotion) took an early branch in `moveTransition` that never emitted `launchColumnAgent` — promotion into the column now launches the babysitter like a manual move does.
- **Hook guard widened** to `--if-status-not review-by-ai,review-by-colleague` (claude hooks, codex hook handler, dev3 skill status line) so agent activity while a PR is in review no longer yanks the task to in-progress.
- **UI**: "PR Babysitter" block in Project Settings → Project Config under the AI Review section (Autonomy select, Handle-comments switch, agent/config selects, Advanced disclosure with capability switches, composed-prompt textarea with reset). i18n en/ru/es.

## Validation

- 40+ new unit/component tests (prompt composition, executor launch matrix, machine promotion effect, config merge, settings UI); full suite green.
- Live e2e on a private sandbox repo: Triage run diagnosed a pre-existing red CI with zero GitHub writes and parked with a note; Fix run answered all three comment surfaces (one reply each), pushed a fix commit to the PR branch only, left threads unresolved, no auto-merge/approve.

Phase 2 (signal wakes, seen-set, budgets, card badges, per-task toggle) is tracked separately.


h0x91b (changes requested):
Hey — Claude here, the AI assistant. I reviewed this branch in a separate worktree (lint green, full `bun run test` green: 7774 pass / 0 fail, plus a browser pass over the new Settings section). Nice work overall: the capability matrix as data, the hard ceilings appended to every generated prompt with a test pinning them, and the deliberate lack of `onExitCommand` are all the right calls.

**Heads-up on process first:** Arseny is not merging this yet. The idea genuinely lands — the need is real — but he wants to live with it for a while before committing to a shape, because in his day-job workflow an agent touching open PRs could cause serious problems. So treat this as "keep it open, keep iterating", not "polish and land". Please don't arm auto-merge.

Below are the findings worth fixing. Pure nitpicks are intentionally left out of this list.

### Blocker

1. **`src/shared/agent-skill-content.ts:222` — the widened hook guard did not reach the hookless skill body.** `SKILL_STATUS_MANUAL` still says `--if-status-not review-by-ai`; only `src/bun/agent-skills.ts:45` got `,review-by-colleague`. That string feeds `GENERIC_SKILL_BODY` → Cursor, Gemini, opencode, generic. Consequence: any hookless agent flips `review-by-colleague` → `in-progress` on its first turn, then its "end of every turn" rule moves the task to `review-by-user`. `facts.prPromoted` is already `true`, so the task never re-promotes — the column silently empties and the babysitter never runs again. Reachable two ways: the new Babysitter Agent dropdown lists Cursor/Gemini, and the *primary* agent of the task does it too on its next turn. Note decision 179 claims the "dev3 skill status line" was widened — it was not. Please add a test asserting all three skill bodies carry the identical guard string.

### Worth fixing

2. **Dead `configId` — `src/mainview/components/ProjectSettings.tsx:16` and `src/bun/lifecycle/executor.ts:423`.** `"claude-bypass-sonnet"` does not exist; the real ids look like `claude-bypass-sonnet5-medium` / `claude-bypass-sonnet5-xhigh`. `findConfig` (`src/bun/agents.ts:476`) silently falls back to `agent.configurations[0]` for an unknown id, i.e. `claude-auto-fable5-medium`. Verified live: `#babysitter-config` holds an id that matches no `<option>`, while the select displays "Auto (Fable 5, Medium)". So the advertised default (bypass + Sonnet) never runs — Auto + Fable 5 does, a different permission mode and a different model. The literal is copied from the pre-existing `review-by-ai` path, so the rot isn't new, but this branch makes it its own documented default and adds a shown-vs-stored mismatch in a new select. Worth a test that the default `configId` actually exists among the agent's configurations.

3. **Opt-out asymmetry — `src/bun/lifecycle/executor.ts:417`.** The `review-by-ai` branch (line 406) honours the existing convention "`builtinColumnAgents` present but key missing = disabled". The `review-by-colleague` branch ignores `builtinColumnAgents` entirely and gates only on `babysitterEnabled()`. A project whose committed `.dev3/config.json` has `builtinColumnAgents: {"review-by-ai": …}` — which until this branch meant "no other builtin column agent" — silently gets a babysitter on upgrade. Either honour the same presence check, or write down in decision 179 why `autonomy` is the only switch.

4. **Flip the default to OFF — this is a decision from Arseny, please implement it as stated.** Right now absent config resolves to `triage` (`src/shared/types.ts:295`), so every existing project starts spawning a column agent right after upgrade. "Read-only" is enforced purely by prompt text, and per finding 2 the process that actually launches is `claude-auto-fable5-medium` (permission mode `auto`) — it can write. Decision 179 accepts the trust model but never addresses consent or token cost.

   **What he wants — model the whole feature on AI review, not just the default.** His words: conceptually this task is a sibling of AI review — you will want prompt editing, agent choice, which account it runs from, which settings preset. So build it in the same shape rather than as its own species:

   - **An explicit enable switch, off by default.** Mirror `autoReviewEnabled` / `peerReviewEnabled`: a plain boolean on `Project` + `.dev3/config.json`, default declared in `DEFAULTS` (`src/bun/repo-config.ts`), a toggle in Project Settings. Absent config must resolve to disabled — nothing may start on upgrade. Whether you keep `autonomy: "off"` as the storage for that, or add a dedicated boolean, is your call; the requirement is one obvious switch that reads like AI review's, and OFF when nobody has said otherwise.
   - **Same layout language as the AI Review section**, so the two read as siblings on the same tab: agent picker, config/preset picker, editable prompt — arranged the same way, with the babysitter-specific Autonomy select and capability overrides living under the switch.
   - **Account selection is on his wish list too** (which account the babysitter runs from, as with agent presets elsewhere). `launchColumnAgent` currently never passes an `accountId` through to `resolveCommandForAgent`, so this is a real gap — treat it as an explicit follow-up rather than something to bolt on hastily here, and say what you think the right scope is.

   Also update decision 179's default-on rationale and the changelog entry — the latter currently advertises "ships enabled at the read-only Triage level", which will no longer be true.

5. **Decision record 179 contradicts itself.** Bullet 1 says "Fix (default) = push/reply/rebase"; bullet 3 says "Default is read-only Triage". Bullet 1 is stale from an earlier design. A future agent will read it and believe writes are on by default. Drop the "(default)".

6. **`src/bun/lifecycle/machine.ts:194` — the pr-promotion launch drops conditions the normal path honours.** `event.launchColumnAgent !== false` is ignored (line 302 respects it), and the surrounding `event.cause` branch already discards `event.guards`, so the `ifStatus: "review-by-user"` guard sent from `prDetected` (line 668) is silently inert for this cause. Also `phase !== "preparing"` still allows a launch while `tearing-down`. It happens to be safe today because `prDetected` checks the column itself and the actor serializes events — but a pane launched into a worktree being torn down is a live edge. Mirror line 302's conditions, or leave a comment explaining why not.

7. **`src/mainview/components/ProjectSettings.tsx:1436` — Off erases config.** Setting Autonomy to `Off` drops the whole `review-by-colleague` entry from `builtinColumnAgents`, discarding the user's hand-written prompt and agent/config choice. Same shape for `handleComments: false`, which is only persisted when the level can reply — flip to Triage and back to Fix and the toggle silently returns to on. Off should be a switch, not an eraser: keep the entry and the flag, and gate on `autonomy` at read time (the executor already does).

### Missing coverage worth closing along the way

- Nothing asserts that `CLAUDE_SKILL_BODY`, `CODEX_SKILL_BODY` and `GENERIC_SKILL_BODY` carry the same `--if-status-not` guard — exactly the gap that let finding 1 through.
- No test that the default `configId` exists (finding 2).
- No test that pr-promotion does *not* launch the column agent while `runtime.phase` is `preparing` / `tearing-down`; only the positive case was added (`machine.test.ts:626`).
- No backend test for the `babysitter` round-trip `updateProjectSettings` → `projects.json` → `resolveProjectConfig`; the UI test only asserts the RPC payload.

h0x91b:
Separate from the fix list — I want to spell out the problem I ran into and hear how you see it.

**Locally we can do anything.** As long as nothing is pushed, the babysitter can rebase, fix, build commits — it touches nobody and costs almost nothing. The boundary isn't the agent doing work, it's the push.

**The push is an expensive boundary, and here is how expensive.** In my case, like in many companies, every PR gets a large armada of review bots thrown at it. Every push re-triggers all of them, from scratch. One such run today costs several dollars — easily $10–20 per single push. And those bots regularly produce nitpicking noise: not problems, just line-by-line opinions.

**And then a loop forms.** Push → the armada wakes up → it leaves a pile of small remarks → the babysitter "fixes" them → push → the armada wakes up again. Every turn of that loop costs tens of dollars, and it is self-sustaining: the bots trigger the agent, the agent triggers the bots. A single PR can eat company resources without bound this way, and this isn't hypothetical — it is exactly what I will hit on the first working day with this feature enabled. So "just push to keep the PR fresh" is not an option.

**But pushing is necessary**, otherwise the PR goes stale: in an active repo the base moves constantly and the branch falls behind fast. And one honest trigger is already visible from that: **if the PR can no longer be merged without a rebase, that is a clear sign it's time to wake the babysitter.**

**I haven't closed this fork myself.** When exactly should it write? On a cron — once a day, once a week? On a button, when I have said "prepare this for merge" or moved the task into some hot pre-merge phase? Something else entirely? Genuinely unclear right now. I lean toward the push requiring **either an explicit human confirmation, or a very narrowly defined condition** — not "the agent decided it was time".

You have probably been thinking about this already, so please share your own view — what have you worked out, and do you have ideas I haven't considered? You are deeper in this than I am, so I want your picture, not agreement with mine.

Two concrete questions:

**A) How do you see the write model?** Which triggers do you consider sufficient for the babysitter to spend a push, and what should require human confirmation?

**B) Surface the phase map in the PR.** Decision 179 references an end-state spec living in the notes of task seq 9 — that is invisible from the PR to everyone, including me while reviewing. Please write out, in the PR or in the decision record itself: what belongs to phase 1, what to phase 2, what to phase 3, and which phase holds the answer to question A. Right now there is nothing to discuss direction against, because the spec is hidden.

arditti:
Thanks @h0x91b  for having a look at it. I suggest we take a step back and discuss what I'm trying to achieve here — maybe there are other tactics to use, or maybe it's not dev-3's ownership at all.

Nowadays I'm delivering an average of 10 PRs a day; 5 require attention, and 4 of those could be closed automatically by the agent that already filed the PR, thanks to the context and tools it has connected. What I'm trying to see is how, using dev3, I handle the 4 out of 5 autonomously — and reliably capture the 1 out of 5 that requires my attention.

That's why my original suggestion included the monitor / apply-but-don't-push / push knobs — I think I need them to gate where the loop escalates to a human operator. Regarding the mechanics of how to actually monitor: I had it planned as phase 2, but I now see it's more important we discuss it first. What I researched about the others: Devin, Claude auto-fix and Charlie converge on the same rails — ignore bot comments by default, budget the wakes, never merge — but none of them treats the push itself as the expensive action, and the webhook-based ones structurally can't see a conflict.

Quick thought on the monitor itself: it can be a loop/cron whose cadence depends on the project's context; where the context is plain, default to 30min/1h and allow modification — or maybe wiring it to the prWatch, which already polls CI, mergeability and review threads at a 5–15min cadence. What I can't find yet is a novel way to *subscribe* to changes instead of polling that doesn't require a bot running on the GitHub side and/or an alias @-mention — so polling stays the honest baseline for now.

One more direction on the auto-push gate: expand the shared `.dev3/config.json` (committed into the repo, so it's the maintainers' voice, and changing it goes through PR review itself) with a global max-level knob — the repo deterministically declares the ceiling of what any dev3 babysitter may do against its PRs: monitor / apply-but-don't-push / push. Local settings can only go below it, never above, and an absent knob means pushes are off — so a repo like your day-job one is protected by default, and auto-fixes only ever run against repos that explicitly opted in. The knob would have to be read from the base branch, not the PR branch, so an agent can't raise its own leash in the same PR.

arditti:
Autonomous activity of agents within the dev3 harness needs deeper discussion. Closing this one — will start a proper issue discussion on this core topic.

PR #1193, opened 2026-07-30
Fix Ctrl+O being stolen from the terminal on macOS

alonkochba:
## Problem

On macOS the "Open in…" picker fires on **Ctrl+O** as well as ⌘O, and because its handler is capture-phase it swallows the keystroke before the focused terminal sees it. Any CLI running in a dev3 pane loses Ctrl+O — for example Claude Code binds it to the transcript / expand-tool-output toggle (moved there from Ctrl+R in its v1.0.113, 2025-09-13), which is simply unreachable from inside dev3 on macOS.

The handler and the registry disagree. `src/mainview/keymap.ts` declares:

```js
{ id: "open-in", keys: { mac: "⌘O", other: "Ctrl+O" }, ... scope: "desktop" }
```

…so on macOS the intended key is ⌘O only. But `src/mainview/App.tsx` gated the branch on `(e.metaKey || e.ctrlKey)`, which accepts either modifier on every platform. Introduced in #1090 (v1.39.1) and carried through the #1106 rework.

## Fix

Add `hasAppModifier(e)` to `src/mainview/utils/platform.ts` — `isMac() ? e.metaKey : e.ctrlKey` — and use it for the `open-in` branch, so the runtime check matches what `keymap.ts` advertises. ⌘O still opens the picker on macOS; Ctrl+O now passes through to the terminal. Non-mac behaviour is unchanged (Ctrl+O still opens the picker), as is the existing `remote` early-return that yields ⌘O to the browser's native Open-File dialog.

`platform.ts` is the established home for OS-dependent keybinding branching (its module docstring already describes exactly this role), so the helper is reusable.

## Scope note

The same `metaKey || ctrlKey` pattern appears on ~24 branches of that handler chain, so other Ctrl-prefixed keys (⌃P, ⌃K, ⌃N, ⌃[, ⌃]) are candidates for the same shadowing on macOS. I deliberately did **not** sweep them: I only verified the Ctrl+O case, and some entries declare extra modifiers off-mac (`zoom-out` is `⌘-` / `Ctrl+Alt+-`), so a blanket substitution needs per-branch judgment. Happy to follow up with a broader pass if you want it — `hasAppModifier` is there to build on.

## Tests

- New `src/mainview/utils/__tests__/platform.test.ts` — `hasAppModifier` accepts only ⌘ on macOS and only Ctrl elsewhere, and is false with no modifier.
- `bun run lint` — clean.
- `src/mainview/__tests__/{keymap,App,TerminalView,help}` — 189 + 17 tests pass.

## Test instructions

1. On macOS, open a task with a terminal pane running Claude Code. Press **Ctrl+O** — expect Claude Code's transcript/expand view to toggle, and **no** "Open in…" picker.
2. Same pane, press **⌘O** — expect the "Open in…" picker modal to open on the task's worktree.
3. Press **⌘O** on the Kanban board with no task open — expect the picker on the current project.
4. On Linux, press **Ctrl+O** — expect the picker (unchanged behaviour).
5. In remote/browser mode on macOS, press **⌘O** — expect the browser's own Open-File dialog, not the picker.

### DIFF
--- a/change-logs/2026/07/30/fix-ctrl-o-mac-terminal.md
+++ b/change-logs/2026/07/30/fix-ctrl-o-mac-terminal.md
@@ -1,0 +1,3 @@
+Short: Ctrl+O reaches the terminal on macOS
+
+On macOS the "Open in…" picker fired on Ctrl+O as well as ⌘O, and its capture-phase handler swallowed the keystroke before the focused terminal saw it — so Ctrl+O-bound features of the CLI running in the pane (Claude Code's transcript/expand toggle, for one) were unreachable. The shortcut now matches the keymap exactly: ⌘O on macOS, Ctrl+O elsewhere.
--- a/src/mainview/App.tsx
+++ b/src/mainview/App.tsx
@@ -8,7 +8,7 @@ import { trackPageView, trackEvent, registerAgents } from "./analytics";
 import type { CodingAgent, GlobalSettings as GlobalSettingsType, Project, RemoteNetInterface, RequirementCheckResult, RosettaWarningInfo, SharedArtifact, SharedImage, Task, TaskDialogSubject, TaskStatus, UpdateChangelog } from "../shared/types";
 import { orderProjectsForDisplay, taskSeqLabel } from "../shared/types";
 import { useGlobalShortcut } from "./hooks/useGlobalShortcut";
-import { isRemote } from "./utils/platform";
+import { hasAppModifier, isRemote } from "./utils/platform";
 import { adjustZoom, applyZoom, ZOOM_STEP, DEFAULT_ZOOM } from "./zoom";
 import { useViewport } from "./hooks/useViewport";
 import { useMobileDenseZoom } from "./hooks/useMobileDenseZoom";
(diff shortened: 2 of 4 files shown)


h0x91b:
Hey Alon — Claude here, the AI assistant that reviewed this on the dev3 side. Merged! 🎉

Thank you for this one. Genuinely: the hardest part of any bug is finding out it exists, and you did the whole job — you traced it to the exact line, spotted that `App.tsx` and `keymap.ts` disagreed with each other, and even dug up *when* Claude Code moved its transcript toggle to Ctrl+O. Reading the PR description was faster than investigating it myself would have been.

And thanks for coming back — going through the history, this is your third PR here (after the git-fetch-over-SSH fix in #512), which makes you one of our repeat contributors. Much appreciated.

Two notes on the review:

- The helper landing in `platform.ts` was the right call — its module docstring already described exactly that role, so it slotted in with nothing to argue about.
- Your scope note about the other `metaKey || ctrlKey` branches was the most valuable part of the PR, and you were right not to sweep them. I counted them properly and opened a follow-up on our board: it is **23** occurrences, of which **7** actually collide with real shell bindings — `Ctrl+H` (backspace), `Ctrl+K` (kill-line), `Ctrl+P`/`Ctrl+N` (history), `Ctrl+G` (abort), `Ctrl+[` (Escape in a terminal) and `Ctrl+Q` (XON). Your `hasAppModifier` is exactly what that follow-up will build on, so the groundwork is already done.

If you feel like taking that sweep yourself, say the word and it is yours — otherwise we will pick it up. And please keep them coming if you hit anything else odd; reports at this level of detail are a gift.

PR #1245, opened 2026-08-04
Add backend-neutral read-only pane capture

h0x91b:
## Summary

- add `TerminalBackend.captureView(sessionId, viewId, request?)` as a backend-neutral, read-only pane observation API with explicit outcomes, identity bracketing, bounded physical rows, UTF-8 byte limits, and plain-text sanitization
- implement one-turn tmux capture and native capture from independently published semantic or compact artifacts
- serialize native publication, readiness, replacement, and cleanup with generation-owned per-session locks, including stale-owner and ABA-safe rollback handling
- keep producer modes explicit (`none | semantic | compact | semantic-and-compact`) and make each artifact settle independently before advertisement

## Production safety and compatibility

Production capture remains disabled: the production native host uses `captureMode: "none"`, advertises no capture surface, and returns `not-enabled` immediately. This PR does not activate capture or migrate `getTerminalPreview`.

The on-disk session record schema remains unchanged. Capture capabilities are additive and optional, frozen N-2 reader fixtures verify older readers continue to accept the records, and existing `native-sessions` paths are not moved, renamed, or deleted. Session locks use a new sibling root; older binaries remain readable but do not participate in the new mutual-exclusion protocol.

## Validation

- `bun run lint`
- 12 focused capture/lock Vitest files: 248 tests
- real tmux backend E2E: 1 test
- real native capture E2E: parser, compact projection, and production-shaped parser-disabled host
- real cross-process session-lock ABA E2E
- `bun run test`: renderer 3,522; backend 4,765 with 3 skipped; CLI 693

## Known unmeasured cases

Native capture and its production-cost envelope remain unmeasured on Windows in this change. Capture continuity and identity behavior across real app, native-host, and tmux-server restart boundaries are also not measured end to end here; the existing lifecycle restart proofs cover session ownership, not this new capture seam.


h0x91b:
CI repair evidence: workflow run 30868106556, Ubuntu package-runtime job 91864201480 reached the DSR probe and observed exactly one shell reply, then failed because the test accepted an earlier persisted snapshot with ingested.replies still at 0. Commit 48308ce64 changes only the POSIX test observation barrier to wait for ingested.replies >= 1; product behavior is unchanged. Local verification: native-live-parser E2E passed 3 consecutive runs, lint passed, and the 12-file capture/lock set passed 248/248 tests.

PR #1255, opened 2026-08-05
Open files from paths mentioned in terminal output

arditti:
## What

File paths that agents mention in terminal output are now rendered as underlined links and open on Cmd/Ctrl+Click, with a new global setting choosing the behavior.

- **Detection** (`src/mainview/terminal-file-links.ts`): a custom ghostty-web link provider finds absolute, `~`, `./`, `../`, relative paths, bare filenames with extensions, and `:line:col` suffixes — including paths soft-wrapped across rows (logical-line stitching via `isWrapped`). Only candidates that exist on disk become links: the new `resolveTerminalPaths` RPC stats each one against the task worktree, then the project directory (10s renderer cache), so prose tokens never linkify.
- **Persistent underlines** (`src/mainview/terminal-link-underlines.ts`): ghostty-web only underlines the hovered range, so a pointer-events-none overlay canvas underlines every resolved link in the viewport (redraws on render/scroll/resize, debounced).
- **Setting**: Settings → Terminal → "File path click action" — *Preview in dev3* (default) / *OS default app* / *Reveal in Finder*. Registered in settings search; en/ru/es strings.
- **Preview modal** (`FilePreviewModal`): markdown rendered via the existing sanitized `MarkdownDocument` pipeline with a Rendered/Raw toggle, images, text (256 KB cap), binary/too-large fallbacks; footer: Copy content, Copy path, Open folder (reveal), Open in default app, Close. Activation requires Cmd/Ctrl so plain clicks keep reaching tmux mouse mode.
- **Remote/browser mode** always previews in-app (host-side open would be invisible remotely).

## Why

Agents constantly print paths ("Draft is ready at kb-playbook-drafts/waf.md") with no way to open them. CMUX has the equivalent feature; its design was researched and the key ideas (existence probe before linkifying, cwd-relative resolution) match this implementation — see `decisions/205-terminal-file-path-links.md`.

## Testing

- 29 new unit tests: candidate regex, wrapped-line offset mapping, provider modifier gating and caching, viewport row math, backend resolution and preview classification.
- Manual QA in browser mode: setting persists, underlines render, Cmd+Click on a relative path in a live tmux terminal opened the markdown preview end-to-end.
- `bun run lint` + full `bun run test` green except two pre-existing environment flakes reproduced identically on `origin/main` (`native-pane-owner-forward` unix-socket path length, `artifactTemplateFileProtocol` local Chrome spawn timeout).


src/mainview/terminal-link-underlines.ts
```diff
@@ -0,0 +1,149 @@
...
+		clearTimeout(redrawTimer);
+		redrawTimer = setTimeout(() => {
+			void redraw();
+		}, REDRAW_DEBOUNCE_MS);
+	}
+
+	// No onResize subscription: the container ResizeObserver + the onRender
+	// that follows any reflow already cover size changes.
+	const subscriptions = [
+		term.onRender(() => scheduleRedraw()),
```
h0x91b: **Blocker: this subscription never fires.** `renderEmitter` exists in the ghostty-web 0.4.0 bundle only as `.event` (here, via `onRender`) and `.dispose` — nothing ever calls `.fire()`, in either the ESM or the UMD build. `startRenderLoop` renders every frame but emits only `cursorMoveEmitter`.

So the overlay's refresh reduces to `onScroll` + the `ResizeObserver`, and `onScroll` only fires when the viewport actually moves (`scrollEmitter.fire` in `scrollToBottom` / `scrollToLine` / the wheel animation). With the terminal pinned to the bottom — `viewportY === 0`, the normal case while watching an agent — new output moves nothing, so `scheduleRedraw()` is never called and no underline is drawn or refreshed. It only appears after a manual scroll or a resize.

Suggestion: drive `scheduleRedraw()` from `TerminalView`'s own write path (it already batches through `writeRafId`) — that's the real "content changed" signal. `term.onCursorMove` does fire if you want a second trigger. Worth also clearing the canvas synchronously on scroll/write and leaving only the expensive recompute debounced, so a stale overlay never sits under fresh text.

The comment just above ("the onRender that follows any reflow already cover size changes") rests on the same assumption.

src/mainview/terminal-file-links.ts
```diff
@@ -0,0 +1,242 @@
...
+	while (startY > 0 && y - startY < MAX_LOGICAL_ROWS && getLine(startY)?.isWrapped) startY--;
+	const rows: LogicalLineRow[] = [];
+	let text = "";
+	for (let rowY = startY; rowY - startY < MAX_LOGICAL_ROWS; rowY++) {
+		const line = getLine(rowY);
+		if (!line) break;
+		if (rowY > startY && !line.isWrapped) break;
+		const next = getLine(rowY + 1);
+		const isLast = !next?.isWrapped;
+		const rowText = line.translateToString(isLast);
```
h0x91b: **Blocker: `translateToString()` is not padded, so index ≠ column.** In ghostty-web 0.4.0:

```js
getChars() { const A = this.cell.codepoint; return A === 0 ? "" : String.fromCodePoint(A); }
// translateToString concatenates only getChars() → blank cells contribute nothing
```

An empty cell has codepoint 0 and yields the empty string, so the row text is never `cols` characters wide. Every index after a blank cell maps one column too far left, which shifts both the hover range and the overlay underline off the text. Same for wide characters — the spacer cell of an emoji/CJK glyph is codepoint 0.

ghostty's own `UrlRegexProvider` avoids this method for exactly this reason and uses its own helper:

```js
lineToText(A) {
  const B = [];
  for (let g = 0; g < A.length; g++) {          // exactly cols iterations
    const C = A.getCell(g).getCodepoint();
    C === 0 || C < 32 ? B.push(" ") : B.push(String.fromCodePoint(C));
  }
  return B.join("");                             // one cell = one char
}
```

Building the logical line that way makes the mapping exact, lets you drop `trimRight` entirely, and makes each row's `offset` simply `rowIndex * cols`. It also removes the CJK caveat listed under Risks in the decision record — with a space for the spacer cell, wide characters line up too.

Where it shows up in practice: lines drawn with cursor positioning (TUI frames, tmux status lines, regions after `\x1b[K`) and any line with an emoji before the path. `echo src/foo.ts` starts at index 0 and looks perfect, which is why it survives casual testing.

src/mainview/__tests__/terminal-file-links.test.ts
```diff
@@ -0,0 +1,175 @@
...
+});
+
+function fakeBuffer(rows: Array<{ text: string; isWrapped?: boolean }>) {
+	return (y: number) => {
+		const row = rows[y];
+		if (!row) return undefined;
+		return {
+			isWrapped: row.isWrapped ?? false,
+			translateToString: (trimRight?: boolean) =>
+				trimRight ? row.text.replace(/\s+$/, "") : row.text,
```
h0x91b: This fake is why the column-mapping bug stays green. It models `translateToString` as "return the row string as-is", i.e. a row padded with spaces — but the real implementation omits codepoint-0 cells entirely, so a real row is *shorter* than `cols` and indices no longer equal columns.

The two exact-coordinate assertions below (`{ start: { x: 6, y: 0 }, end: { x: 10, y: 1 } }`) are therefore checking a contract ghostty doesn't have.

Worth switching the fake to the real semantics (one character per cell, blanks as spaces once the production code does the same) and adding two cases that currently can't fail: a row with an emoji before the path, and a row with a gap of unwritten cells before it.

src/mainview/terminal-file-links.ts
```diff
@@ -0,0 +1,242 @@
...
+			if (!target) continue;
+			const range = mapRangeToBuffer(logical.rows, candidate.start, candidate.end);
+			if (!range) continue;
+			links.push({ target, raw: candidate.raw, range });
+		}
+		return links;
+	}
+
+	return {
+		provideLinks(y, callback) {
```
h0x91b, line 271: `provideLinks` always answers asynchronously (the callback fires inside `.then`), and ghostty's `LinkDetector.scanRow` does `scannedRows.add(y)` *before* awaiting the providers. So while an RPC for a fresh path is in flight, a second `getLinkAt` on that row sees the row as already scanned, finds an empty cache, and returns `undefined`.

Since `handleMouseMove` (throttled at 16 ms) always precedes a click, the practical symptom is that the first hover and the first Cmd+Click on a just-printed path do nothing and the second attempt works. And because `invalidateCache()` runs on **every** `term.write`, that window reopens on every chunk of streaming output.

Suggestion: answer synchronously with only the paths already in the resolve cache, kick off the RPC in the background, and let `onResolutionsChanged` + the next invalidation pick up the rest. Warm-cache lookups then resolve within the same microtask chain, and the overlay can be synchronous too.

src/mainview/terminal-link-underlines.ts
```diff
@@ -0,0 +1,149 @@
...
+		const gen = ++generation;
+		const size = syncSize();
+		if (!size) return;
+		const charWidth = term.renderer.charWidth;
+		const charHeight = term.renderer.charHeight;
+		if (!charWidth || !charHeight) return;
+		const buffer = term.buffer.active;
+		const scrollback = Math.max(0, buffer.length - term.rows);
+		const viewportY = term.viewportY;
+		const rowRanges = await Promise.all(
```
h0x91b: This fans out one `linksForRow` per viewport row, and each row's `resolveCached` batches only its own candidates — `inFlight` deduplicates only byte-identical batches — so a redraw can issue up to `term.rows` concurrent `resolveTerminalPaths` calls. Each one does up to 64 candidates × up to 2 bases of **synchronous** `statSync` in the same Bun process that pumps the PTY WebSocket, plus a `getProject`/`rawLoadTasks` cache stat per call.

`getLogicalLine` also re-stitches the same wrapped block once per row (up to 40 `translateToString` calls each), so a 40-row wrapped paragraph is reassembled 40 times per redraw.

Suggestion: collect candidates for the whole viewport, resolve them in a single RPC, then draw. That also removes the per-row duplicate stitching.

src/mainview/components/FilePreviewModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+				<div className="flex items-center flex-wrap justify-end gap-2 px-4 py-3 border-t border-edge">
+					{textContent !== null && (
+						<button type="button" onClick={handleCopyContent} className={footerButton}>
+							{t("terminal.filePreviewCopyContent")}
+						</button>
+					)}
+					<button type="button" onClick={handleCopyPath} className={footerButton}>
+						{t("terminal.filePreviewCopyPath")}
+					</button>
+					{!fileMissing && (
```
h0x91b: This one is a real remote-mode bug: `reveal` is gated only on `!fileMissing`, while "Open in default app" right below is correctly gated on `isElectrobun`. In a browser/remote session this button calls `openTerminalPath({ mode: "reveal" })`, which opens a Finder window **on the host machine** — invisible to the person who clicked it.

It also contradicts the policy documented in `terminal-path-open.ts` ("Browser mode always previews in-app: `Utils.openPath` would open the file on the HOST machine") and the setting's own description.

`{isElectrobun && !fileMissing && (…)}` for both buttons. A small test asserting neither open button renders when `isElectrobun` is false would lock it down — there are currently no tests for this component.

src/bun/rpc-handlers/terminal-paths.ts
```diff
@@ -0,0 +1,137 @@
...
+			Utils.openPath(statSync(params.path).isDirectory() ? params.path : dirname(params.path));
+		}
+		return;
+	}
+	Utils.openPath(params.path);
+}
+
+async function readFilePreview(params: { path: string }): Promise<FilePreviewResult> {
+	log.info("→ readFilePreview", { path: params.path });
+	if (!isAbsolute(params.path) || params.path.includes("\0")) {
```
h0x91b, line 155: `readFilePreview` (and `openTerminalPath` above) validate only "absolute and no NUL" — the path is never checked against the bases `terminalPathBases` already computes, and it comes straight from the client rather than from the resolve step.

Both handlers are reachable from an authenticated remote browser session, so this returns the contents of any file on the machine (`~/.ssh/id_rsa`, `~/.aws/credentials`), and `mode: "system"` hands any path to `Utils.openPath` — macOS `open` will *execute* a `.command` / `.app` bundle.

There is precedent (`listDirectory` already exposes the filesystem to the renderer behind the same cookie auth), so this may well be an acceptable trade-off here — but right now it reads as not-yet-decided rather than decided. Two cheap options: reuse `terminalPathBases` and require the path to sit under a worktree / project / home, or have `resolveTerminalPaths` hand back an opaque handle and accept only that.

src/bun/rpc-handlers/terminal-paths.ts
```diff
@@ -0,0 +1,137 @@
...
+	} catch {
+		return { kind: "not-found" };
+	}
+	if (st.isDirectory()) return { kind: "directory" };
+	const size = st.size;
+	const ext = params.path.split(".").pop()?.toLowerCase() ?? "";
+	const imageMime = PREVIEW_IMAGE_MIME[ext];
+	if (imageMime) {
+		if (size > FILE_PREVIEW_MAX_IMAGE_BYTES) return { kind: "too-large", size };
+		const buffer = readFileSync(params.path);
```
h0x91b: A 20 MB image is read with synchronous `readFileSync` on the main process and then base64-encoded into a single ~27 MB JSON RPC response. On the desktop that's a visible stall of the process that also drives tmux and the PTY stream; through a Cloudflare tunnel in remote mode it's much worse.

Suggestion: drop `FILE_PREVIEW_MAX_IMAGE_BYTES` to something like 2–4 MB (a preview doesn't need more), and move both reads to the async `fs/promises` variants.

Minor, same function: `buffer.subarray(0, FILE_PREVIEW_MAX_TEXT_BYTES).toString("utf-8")` can cut a multi-byte character in half and show a replacement glyph at the end of a truncated preview.

src/mainview/components/FilePreviewModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+			return <p className="text-fg-3 text-sm p-6">{t("terminal.filePreviewLoading")}</p>;
+		}
+		switch (preview.kind) {
+			case "text":
+				return (
+					<div className="min-h-0 flex-1 overflow-auto p-4">
+						{isMarkdown && !showRaw ? (
+							<MarkdownDocument body={preview.content} />
+						) : (
+							<pre className="font-mono text-xs leading-relaxed text-fg whitespace-pre-wrap break-words">
```
h0x91b: **UI polish — this is the biggest visual win in the modal.** Non-markdown files (i.e. most files an agent mentions) render as a single-colour `<pre>` at 12 px with `whitespace-pre-wrap`, so:

- there's no syntax highlighting at all — a wall of uniform monospace;
- `pre-wrap` wraps long lines instead of scrolling horizontally, which breaks indentation alignment and makes code hard to scan;
- there's no line-number gutter, even though the link that opened the file often carried a `:line:col` suffix.

Cheapest meaningful fix, no new dependency (the repo has no highlighter, and adding one for this is probably not worth it): `whitespace-pre` + `overflow-x-auto`, and a line-number gutter with `tabular-nums text-fg-muted select-none`. If the `:line` from the link is available, scrolling to it and giving it a soft `bg-accent/10` row highlight would make the feature feel complete.

Also worth capping the measure for the rendered-markdown branch: at 896 px and `text-sm` a paragraph runs roughly 105–110 characters against a comfortable 65–75. `max-w-[70ch] mx-auto` on the prose container (leaving `pre` and images full width) fixes it.

src/mainview/components/FilePreviewModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+						)}
+						{preview.truncated && (
+							<p className="mt-4 text-fg-muted text-xs">{t("terminal.filePreviewTruncated")}</p>
+						)}
+					</div>
+				);
+			case "image":
+				return (
+					<div className="min-h-0 flex-1 overflow-auto p-4 flex items-center justify-center">
+						<img src={preview.dataUrl} alt={path} className="max-w-full max-h-full object-contain" />
```
h0x91b: **UI polish.** Measured on a real 128×128 PNG: the image renders 1:1 in an 896×272 box with `outline-style: none`.

- A 1 px low-opacity outline keeps the edge crisp against the surface — pure white at 10 % in dark mode, pure black at 10 % in light mode (a tinted neutral picks up the surface colour and reads as dirt on the edge).
- Transparent PNGs sit directly on the modal surface with nothing behind them, so it's impossible to tell transparency from a matching background.
- The header shows only the path — no dimensions, no file size, which is usually the first thing you want from an image preview.

Also: `ImageLightbox` (`src/mainview/components/ImageLightbox.tsx`) already exists and takes `{ paths, currentIndex, onClose }`. Routing image previews to it, or at least matching its treatment, would avoid a second image viewer in the app.

src/mainview/components/FilePreviewModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+				aria-labelledby="file-preview-title"
+				tabIndex={-1}
+				className="bg-overlay border border-edge rounded-2xl shadow-2xl w-[min(56rem,92vw)] max-h-[85vh] flex flex-col outline-none"
+			>
+				<div className="flex items-center gap-3 px-4 py-3 border-b border-edge">
+					<h2
+						id="file-preview-title"
+						className="min-w-0 flex-1 text-fg text-sm font-mono truncate"
+						title={path}
+						dir="rtl"
```
h0x91b: **UI polish — the `dir="rtl"` trick has a visible side effect.** Measured in the browser: `direction: rtl` with `text-align: start` means *start = right*. When the path overflows you get the intended head-truncation, but when it fits (e.g. `/nope/definitely-not-here.txt` in the not-found state) the whole title sits flush against the right edge next to the ✕. So the header identity jumps sides depending on path length.

It also buries the one thing the reader is looking for: the filename ends up at the far right end of a small mono line.

Suggestion: two lines instead — the filename as the title (`text-fg font-semibold`, normal direction) and the directory below it in `text-xs text-fg-muted`, middle-truncated. That fixes the alignment jump and puts the filename where the eye lands first.

src/mainview/components/FilePreviewModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+						<button type="button" onClick={() => handleOpen("reveal")} className={footerButton}>
+							{t("terminal.filePreviewOpenFolder")}
+						</button>
+					)}
+					{isElectrobun && !fileMissing && (
+						<button type="button" onClick={() => handleOpen("system")} className={footerButton}>
+							{t("terminal.filePreviewOpenSystem")}
+						</button>
+					)}
+					<button
```
h0x91b: **UI polish — action hierarchy in the footer.** The blue `bg-accent-fill` primary is on `Close`, the least consequential action, while the four actions the modal exists for (Copy content / Copy path / Open folder / Open in default app) are all identical flat ghost text. Reading the footer, nothing tells you what this dialog is *for*. And `Close` appears twice (✕ in the header plus this button).

Suggestion: `Close` becomes quiet/secondary, and the primary emphasis either goes to the destination action ("Open in default app") or to nothing at all — a preview dialog is fine without a primary.

Two more, same footer:

- The action set changes per state (I measured 7 / 5 / 4 / 4 / 3 buttons across the six states, and "Copy content" only appears once text has loaded), and the row is right-aligned — so buttons shift horizontally under the cursor after load. A stable order with disabled states, or keeping the variable action last, avoids the reflow.
- None of the seven buttons has press feedback. The repo's convention is `motion-safe:active:scale-[0.96]` with an explicit `transition-[background-color,color,transform] duration-150 ease-out` — 43 other call sites use it (e.g. `TaskCard.tsx:787`, `ListEditor.tsx:50`).

And the ✕ in the header (line 167) is a text glyph with `hover:bg-elevated`; the house pattern is an inline SVG (`w-4 h-4`, `strokeWidth={2}`, `hover:bg-fg/8`) — see `TaskDetailModal.tsx:420`. A glyph can't match the surrounding text's optical weight and renders differently across platform fonts.

h0x91b (changes requested):
Hey — Claude here, the AI assistant working on this repo with Arseny. I went through the branch commit by commit, read the installed `ghostty-web@0.4.0` bundle to check the API assumptions, ran lint + the branch tests, and drove the preview modal in a headless browser through all six of its states.

**First: thank you.** This is a feature Arseny has taken a run at three separate times and never finished — it kept getting reverted. Having it done end to end (regex detection + on-disk verification, wrapped-line stitching, the preview modal, the global setting, tests, and a decision record) is genuinely valuable, and the shape you picked is the right one. Verifying candidates against the filesystem instead of getting clever with the regex, and gating activation on Cmd/Ctrl so plain clicks still reach tmux, are both exactly right — I checked `handleMouseDown` in ghostty and it only handles the scrollbar, so that policy holds.

I'm requesting changes for two things, and both are the same kind of problem rather than a design mistake: an assumption about ghostty-web's API that type-checks fine but doesn't hold at runtime in the installed bundle. They're also the most likely reason the earlier attempts felt broken, so they're worth nailing down.

### Blockers

**1. `term.onRender` never fires in ghostty-web 0.4.0, so the underline overlay doesn't refresh under live output.** In the shipped bundle `renderEmitter` appears exactly twice — as `.event` in the constructor and in `.dispose`. There is no `.fire()` anywhere, in either the ESM or the UMD build:

```
$ grep -o "renderEmitter\.[a-z]*" node_modules/ghostty-web/dist/ghostty-web.js | sort | uniq -c
   1 renderEmitter.dispose
   1 renderEmitter.event
```

That leaves `onScroll` and the `ResizeObserver`, and `onScroll` only fires on actual viewport movement. While the terminal is pinned to the bottom (`viewportY === 0`, the normal case) new output moves nothing, so no redraw is scheduled at all — underlines only show up after a manual scroll or a window resize. `TerminalView` already batches writes through `writeRafId`; driving `scheduleRedraw()` from there is the honest "content changed" signal. `onCursorMove` does fire (see `startRenderLoop`) if you want a second trigger.

**2. `translateToString()` drops empty cells, so string index ≠ screen column.** `getChars()` returns `""` for codepoint 0, and `translateToString` concatenates only `getChars()`, so a row is never padded to `cols` — which means every index after a blank cell maps one column too far left. ghostty's own `UrlRegexProvider` deliberately avoids that method and uses its own `lineToText()`, pushing `" "` for codepoint 0 and control chars, one character per cell, over the full `line.length`. That's the 1:1 mapping this code needs, and it also makes the CJK caveat in the decision record disappear: a wide char occupies two cells, the second yields a space, and the index lines up exactly.

In practice it bites on any line drawn with cursor positioning (TUI frames, tmux status lines, cleared regions) and on any line with an emoji before the path — so ordinary agent output. Simple cases like `echo src/foo.ts` start at index 0 and look perfect, which is what makes it sneaky.

### Worth fixing before merge

- **First hover / Cmd+Click on a freshly printed path silently does nothing.** `LinkDetector.scanRow` marks a row scanned *before* awaiting providers, and `invalidateCache()` runs on every `write`, so an async provider loses that first lookup. Answering synchronously from the resolve cache (and resolving in the background) fixes it and makes the overlay synchronous too.
- **Browser/remote mode: the preview modal's "Open folder" opens Finder on the host.** One-line gate, details inline.
- **One RPC per viewport row**, each up to 64 candidates of blocking `statSync`, on the process that also pumps the PTY WebSocket.
- **`readFilePreview` / `openTerminalPath` accept any absolute path** without checking the bases `resolveTerminalPaths` already computed.
- **20 MB images are read synchronously and shipped as a ~27 MB base64 data URL** over RPC.

### Preview modal polish

Arseny specifically asked for this part — the popup works, it just doesn't look finished yet. I drove all six states in a browser and measured them: the box is **896 px wide in every state** while its height swings **180 px → 765 px**, so a one-sentence "Binary file (413 KB)" message lands in a very large empty room. Concrete notes are inline; the summary:

- Code files get a bare `<pre>` with `whitespace-pre-wrap` — no highlighting, no line-number gutter, and wrapping instead of horizontal scroll, which destroys indentation. This is the single biggest visual win available.
- Rendered markdown runs the full 896 px, roughly 105–110 characters per line against a comfortable 65–75.
- `dir="rtl"` on the title right-aligns short paths and left-aligns long ones, so the header identity jumps between states. A bold filename plus a muted directory line reads better and puts the filename where the eye lands first.
- The blue primary button is on `Close`, while the four real actions are flat ghost text — and the action set changes per state, so the row reflows.
- The `✕` is a text glyph; the house pattern (e.g. `TaskDetailModal.tsx:420`) is an inline SVG with `strokeWidth={2}` and `hover:bg-fg/8`.
- None of the seven buttons has the repo's `motion-safe:active:scale-[0.96]` press feedback (43 other call sites use it).
- The `` has no outline and no backing for transparency; `ImageLightbox` already exists and may be worth reusing.

### Housekeeping

Two mechanical things that will matter at merge time: `decisions/205-…` collides with `205-cleanup-script-orphaned-tmux-client-watchdog.md` already on `main`, and the changelog entry sits under `2026/08/05` while the changelog policy wants the expected merge date. I left the smaller nits out of the inline comments on purpose to keep this readable — happy to add them if you want the full list.

### What I verified

`bun run lint` is clean, the branch's 40 tests pass, `git status` is clean. Both blockers come from reading `node_modules/ghostty-web/dist/ghostty-web.js` directly rather than from the type definitions. The modal states were driven in headless Chromium at 1440×900. I did not run the full `bun run test`, and I did not test on Linux.

Really nice work on the hard parts — the detection and resolution design is solid, and once the two ghostty-web assumptions are corrected this should finally stick.

arditti:
Thanks for the deep review — both bundle-level findings were exactly right, and everything is addressed in the latest push (rebased on main, so the decision-record conflict is gone too).

**Blockers**
1. **onRender is dead** → the overlay no longer subscribes to it. TerminalView's batched write path (`writeRafId` flush) now calls `contentChanged()` on the overlay — clear synchronously, recompute debounced — with `onScroll` + the ResizeObserver covering viewport moves and layout. Verified underlines now appear while pinned to the bottom under streaming output.
2. **translateToString is unpadded** → replaced with a cell-exact `lineToText` (one UTF-16 unit per cell over `line.length`; codepoint 0, control, and astral codepoints become spaces — astral too, so an emoji's surrogate pair can't shift string indexes off columns). `trimRight` is gone, offsets are exact, and the CJK caveat is deleted from the decision record. The test fake now models the real semantics (explicit codepoint-0 gaps, wide-char spacer cells) with new gap-before-path and emoji-before-path cases.

**The rest**
- **First hover/click lost**: `provideLinks` now answers synchronously from the resolve cache (stale-while-revalidate, 10s TTL); unknown candidates go to one batched background RPC and `onResolutionsChanged` repaints the overlay. No async callback anywhere in the provider path anymore.
- **RPC-per-row**: the overlay collects the whole viewport via `linksForRows` (one batched resolve; wrapped logical lines stitched once, sibling rows skipped).
- **Open folder in browser mode**: gated on `isElectrobun` like the other host-side action; added a FilePreviewModal test suite asserting host actions don't render in browser mode.
- **Backend**: async `fs/promises` throughout; image cap 4 MB; UTF-8-safe truncation (backs off continuation bytes); and `openTerminalPath`/`readFilePreview` now refuse paths outside the home dir + registered project roots — decided and documented in the decision record (same exposure class as `listDirectory`, but bounded; symlink escape accepted and noted).
- **Modal polish**: line-number gutter + `whitespace-pre` + horizontal scroll for code, and the `:line` from the link now scrolls into view with a `bg-accent/10` highlight; markdown prose capped at `max-w-[70ch]`; header is filename (semibold) + middle-truncated directory, `dir=rtl` removed; footer is all-quiet (no primary), stable order with the variable action leftmost; house-style SVG ✕ with `hover:bg-fg/8`; `motion-safe:active:scale-[0.96]` press feedback on every button; images get `ring-1 ring-fg/10` on a `bg-base` backing plus a dimensions · size caption.
- **Housekeeping**: decision record renumbered to 208; changelog entry moved to the expected merge date.

Full `bun run test` is green locally except the pre-existing `native-pane-owner-forward` unix-socket-path-length environment flake (fails identically on `origin/main` here). Happy to take another pass if anything reads off.

src/mainview/terminal-link-underlines.ts
arditti: Fixed in 53dea9af — the overlay no longer subscribes to onRender. TerminalView's batched write flush calls contentChanged() (clear now, recompute debounced); onScroll + the container ResizeObserver cover the rest. The misleading comment is gone too.

src/mainview/terminal-file-links.ts
arditti: Fixed in 53dea9af — replaced with a cell-exact lineToText (one UTF-16 unit per cell over line.length; codepoint 0, control, and astral codepoints become spaces, so surrogate pairs can't shift indexes either). trimRight is gone and the decision record's CJK caveat is deleted.

src/mainview/__tests__/terminal-file-links.test.ts
arditti: Fixed in 53dea9af — the fake now models real cell semantics (explicit codepoint-0 cells, wide chars occupy glyph+spacer). Added the two cases you named: a codepoint-0 gap before the path and an emoji before the path, both asserting exact screen columns.

src/mainview/terminal-file-links.ts
arditti: Fixed in 53dea9af — provideLinks answers synchronously from the resolve cache (stale-while-revalidate, 10s TTL); unknown candidates go to a batched background resolve and onResolutionsChanged repaints the overlay. No async callback remains in the provider path.

src/mainview/terminal-link-underlines.ts
arditti: Fixed in 53dea9af — the overlay now calls a synchronous linksForRows(viewportRows) that stitches each wrapped logical line once (sibling rows skipped) and queues ALL unknown candidates into one background RPC (16ms coalescing, 64-per-call chunks matching the backend cap).

src/mainview/components/FilePreviewModal.tsx
arditti: Fixed in 53dea9af — 'Open folder' is gated on isElectrobun alongside 'Open in default app', and a new FilePreviewModal test suite runs with isElectrobun=false asserting neither host-side action renders.

src/bun/rpc-handlers/terminal-paths.ts
arditti: Decided and documented in 53dea9af (decisions/208): both handlers now require the path to sit under the home directory or a registered project root — same exposure class as listDirectory but bounded. Symlink escape is accepted and noted in the record's Risks.

src/bun/rpc-handlers/terminal-paths.ts
arditti: Fixed in 53dea9af — both reads use fs/promises now, the image cap is 4 MB, and text truncation backs off UTF-8 continuation bytes so a cut never shows a replacement glyph.

src/mainview/components/FilePreviewModal.tsx
arditti: Fixed in 53dea9af — code renders with a line-number gutter (tabular-nums text-fg-muted select-none), whitespace-pre with horizontal scroll, and the :line carried by the link scrolls into view with a bg-accent/10 row highlight (threaded from the candidate through activate → the preview event). Markdown prose is capped at max-w-[70ch] mx-auto (pre/images stay full width). Skipped the highlighter dependency as suggested.

src/mainview/components/FilePreviewModal.tsx
arditti: Fixed in 53dea9af — images get ring-1 ring-fg/10 on a bg-base backing plus a naturalWidth×naturalHeight · size caption. Kept the modal rather than routing to ImageLightbox (it's coupled to the shared-images strip), but matched its treatment.

src/mainview/components/FilePreviewModal.tsx
arditti: Fixed in 53dea9af — dir=rtl is gone; the header is now the filename (text-fg font-semibold) with the directory middle-truncated below it in text-xs text-fg-muted.

src/mainview/components/FilePreviewModal.tsx
arditti: Fixed in 53dea9af — no primary at all now: every footer action is quiet, Close included (the ✕ stays as the second close affordance per house pattern). Stable order with the variable Copy content action leftmost so the right-aligned row never shifts under the cursor after load; motion-safe:active:scale-[0.96] press feedback on all buttons; ✕ replaced with the TaskDetailModal inline-SVG pattern (w-4 h-4, strokeWidth 2, hover:bg-fg/8).

h0x91b (commented):
Claude again. Re-reviewed the whole thing on `53dea9af`, ran the checks, and drove it in a browser against a real tmux terminal. **Both blockers are genuinely fixed — I verified them by measurement, not by reading the commit message.** All 12 threads are resolved.

The `lineToText` approach came out better than what I suggested: mapping astral codepoints to a space as well means surrogate pairs can't shift indexes either, which ghostty's own `lineToText` gets wrong. Nice catch.

### How I verified the two blockers

I printed a line with two CJK-wide glyphs in front of a real path into a project terminal (`中中 see src/mainview/TerminalView.tsx done`) and read the underline straight off the overlay canvas pixels:

```
underline segment: x0 = 81px, width = 261px   (charWidth = 9px)
→ column 9, spanning 29 cells
expected: 中中 = 4 cells, space, "see" = 3, space → path starts at column 9,
          "src/mainview/TerminalView.tsx" = 29 chars   ✔
with the old translateToString path this would have been column 7 (2 cells lost)
```

That single measurement covers both blockers: the column is exact (blocker 2), and the underline appeared **from live output alone** — no scroll, no resize, no window interaction (blocker 1). I also confirmed the eager clear works: a `clear` plus a non-resolving path leaves the overlay empty rather than stranding the previous underline.

Also verified in the browser: line 1200 of a 4 000-line file opens highlighted and scrolled into view; the header renders `direction: ltr` with the filename as the title in every state (the old right-alignment jump on short paths is gone); the image state reports `128×128 · 30 KB` with the ring; and with `isElectrobun` false neither host-side open action renders.

One thing I checked because I was worried about it and it turned out fine: the per-line `<div>` gutter on a 4 000-line file is 4 001 rows / ~12 000 DOM nodes, and dispatch-to-painted was **106 ms** including the file read. No action needed — just noting it so nobody has to wonder.

`bun run lint` is clean; `terminal-paths`, `settings`, `terminal-file-links`, `terminal-link-underlines`, `FilePreviewModal` and `TerminalView` suites are all green (130 tests).

### Two small things the fix surfaced

Neither blocks anything — details inline.

1. The path gate landed on `openTerminalPath`/`readFilePreview` but not on `resolveTerminalPaths`, so a path that exists *outside* the allowed roots still becomes a link and then reports "File not found" when clicked. Verified live: `check /etc/hosts now` underlines `/etc/hosts` (column 6, 10 cells), and clicking it shows the not-found copy.
2. The renumbering left one stale reference: the module doc comment points at `decisions/206-terminal-file-path-links.md`, but the record is 208 (206 is the base-colour-token one from `main`).

### Still open from my summary, and deliberately your call

The dialog is still `w-[min(56rem,92vw)]` in every state, so the binary / too-large / not-found / directory messages sit in an 896 × 188 box. Purely cosmetic and it was never an inline thread — mention it only so it isn't lost.

Decision record 208 is a good one: it names both ghostty-web realities as *verified in the bundle, not the typings*, and records `translateToString` under rejected alternatives as "the bug class that sank earlier attempts". That's exactly the note the next person needs.

I'm leaving the review as a comment rather than an approval — the merge call is Arseny's, and he'll take it from here. Thanks for the fast, thorough turnaround on this; it reads like a finished feature now.

src/bun/rpc-handlers/terminal-paths.ts
```diff
@@ -0,0 +1,182 @@
...
+}): Promise<{ resolved: Record<string, ResolvedTerminalPath | null> }> {
+	const resolved: Record<string, ResolvedTerminalPath | null> = {};
+	const candidates = params.paths.slice(0, RESOLVE_TERMINAL_PATHS_MAX);
+	const bases = await terminalPathBases(params);
+	for (const raw of candidates) {
+		resolved[raw] = null;
+		if (!raw || raw.length > TERMINAL_PATH_MAX_LEN || raw.includes("\0")) continue;
+		const expanded = raw === "~" || raw.startsWith("~/") ? join(homedir(), raw.slice(1)) : raw;
+		if (isAbsolute(expanded)) {
+			resolved[raw] = await statPathKind(resolvePath(expanded));
```
h0x91b: Small asymmetry the new gate introduced: `isTerminalPathAllowed` protects `openTerminalPath` and `readFilePreview`, but `resolveTerminalPaths` still stats any absolute path. So a path that exists *outside* home and the project roots becomes a link, and clicking it lands on `readFilePreview`'s `{ kind: "not-found" }` → the modal says "File not found — it may have been moved or deleted."

Verified live in a project terminal: `check /etc/hosts now` draws an underline at column 6 spanning 10 cells (exactly `/etc/hosts`), and the preview then reports it as missing. Agents mention `/etc/...`, `/opt/homebrew/...` and `/usr/local/...` often enough that this will show up.

Two ways out, either is fine:

- gate `resolveTerminalPaths` with the same predicate, so those paths never underline in the first place (consistent: if we won't open it, it isn't a link);
- or give the refusal its own result kind and copy ("outside the allowed folders") so the message stops claiming the file is gone.

The first also trims a little filesystem probing from the resolve path.

src/bun/rpc-handlers/terminal-paths.ts
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Backend for Cmd/Ctrl+Click file-path links in terminal output: resolve
+ * regex-detected candidates against the task worktree / project directory,
+ * open them per the user's setting, and feed the in-app preview modal.
+ *
+ * `openTerminalPath` and `readFilePreview` take a client-supplied absolute
+ * path (the modal only knows the path), so both gate on
+ * {@link isTerminalPathAllowed}: the path must sit under the home directory
+ * or a registered project. Same exposure class as `listDirectory` behind the
+ * same auth, but bounded — see decisions/206-terminal-file-path-links.md.
```
h0x91b: Stale reference from the renumbering: this points at `decisions/206-terminal-file-path-links.md`, but the record shipped as `208-terminal-file-path-links.md` — `206` is `206-base-color-token-shadowed-text-base.md` from `main`, so the comment now sends the reader to an unrelated decision.

It's the only leftover: `grep -rn "206-terminal-file-path-links" src/ decisions/` returns just this line.

h0x91b (approved):
Claude here again — the AI assistant that ran the two review passes on this PR.

Everything from both rounds is addressed, and I verified the two remaining nits myself instead of sending you back for another lap. Hope that's alright: I pushed one commit to this branch, `49ebf7d` — "Gate terminal path resolution to the home directory and project roots".

What it does:

- `resolveTerminalPaths` now runs the same scope check as `openTerminalPath`/`readFilePreview`. The check moved into `statPathKind`, which covers both branches — worth knowing: the regex accepts `..` segments, so `cat ../../.ssh/id_rsa` was a *relative* candidate that `resolve(base, …)` walked straight out of the worktree. Gating only `isAbsolute` would have missed that.
- The root list is hoisted out of the candidate loop, so it stays one `projects.json` lookup per call rather than up to 64.
- Fixed the doc comment's reference to the renumbered decision 208, plus a one-clause changelog tweak and the matching sentence in the decision record.

Verified: `bun run lint` clean, full suite green (9577 tests), and a live check in the running app — printing `see src/mainview/TerminalView.tsx and /etc/hosts now` underlines only the first path now, at exactly the right column, and Cmd+Click still opens the preview.

Thanks again for building this — genuinely useful, and the cell-exact `lineToText` you wrote is better than what I suggested: mapping astral codepoints to a space handles surrogate pairs, which ghostty's own provider doesn't. Approving.

PR #1343, opened 2026-08-13
Render Mermaid diagrams in Markdown

arditti:
## Summary
- replace the custom Markdown HTML pipeline with Streamdown
- render Mermaid fences through the official Streamdown plugin and Mermaid 11.16.1
- keep wide diagrams sharp with native-width horizontal scrolling
- preserve repo-relative images, rich Markdown diffs, sanitization, and invalid-diagram source fallback

## Verification
- 124 affected renderer tests
- `bun run lint`
- `bun run build`
- dark and light browser visual QA


src/mainview/components/pr-review/markdown.tsx
```diff
@@ -1,53 +1,217 @@
...
+	return `${RELATIVE_LINK_URL_PREFIX}${encodeURIComponent(href)}`;
 }
 
-/** Render a markdown *document* (e.g. a .md file) into safe HTML. Same sanitize
- * allowlist as comments, but `breaks: false` — in documents a single newline is
- * a soft wrap, not a <br> (GitHub renders files vs comments the same way). */
-export function renderMarkdownDocument(body: string): string {
-	const html = marked.parse(body, { gfm: true, breaks: false, async: false });
-	return sanitizeHtml(html, SANITIZE_OPTIONS);
+function unprotectRelativeLink(href: string | undefined): string | undefined {
```
h0x91b: **Sanitizer bypass — the one thing I'd call blocking.**

The `__markdown_link__` prefix is unwrapped *after* Streamdown has hardened the URL, and the prefix itself is something a comment author can type. So `[x](https://dev3.invalid/__markdown_link__/<url-encoded>)` puts an arbitrary URL straight into `href`.

I ran this through `CommentMarkdown` in a throwaway vitest file:

```
[a](JaVaScRiPt:...)                 -> [blocked]        <- direct schemes are handled correctly
[x](https://dev3.invalid/__markdown_link__/<encoded>) ->
   data:text/html,<script>alert(1)</script>
   vbscript:msgbox(1)
   file:///etc/passwd
   views://mainview/index.html
```

All four land in `href` verbatim. `javascript:` is the only one that doesn't, and only because React 19 blocks it separately — so the guard here is somebody else's fuse rather than the sanitizer.

Suggestion: only unwrap URLs that your own remark plugin actually marked (e.g. track them on the mdast node / a private symbol rather than re-detecting the prefix from the string), and run a scheme allowlist on whatever comes out of `decodeURIComponent`.

src/mainview/components/pr-review/markdown-images.tsx
```diff
@@ -26,6 +36,20 @@ export function isDiskImageSrc(src: string): boolean {
 	return Boolean(src) && !src.startsWith("//") && !/^[a-z][a-z0-9+.-]*:/i.test(src);
 }
 
+/** Keep relative sources intact while they pass through Streamdown's URL hardener. */
+export function protectDiskImageSrc(src: string): string {
+	return `${DISK_IMAGE_URL_PREFIX}${encodeURIComponent(src)}`;
+}
+
+function unprotectDiskImageSrc(src: string): string {
```
h0x91b: Same shape as the link one, lower severity: `unprotectDiskImageSrc` runs on any `src`, so `x` reaches `resolveDiskImagePath` and drives a disk read from a third-party comment. It is bounded (the bun side gates paths to the home dir plus project roots, and only images come back), but it is the same trust assumption. Worth fixing together with the link case.

package.json
```diff
@@ -67,18 +67,21 @@
...
 		"@headless-tree/react": "^1.6.3",
 		"@lobehub/icons": "5.8.0",
+		"@streamdown/mermaid": "1.0.2",
 		"bidi-js": "^1.0.3",
 		"electrobun": "1.18.1",
 		"fuzzysort": "^3.1.0",
 		"ghostty-web": "^0.4.0",
 		"js-toml": "1.1.0",
 		"marked": "^18.0.6",
+		"mermaid": "11.16.1",
```
h0x91b: Could you pin this to `11.16.0`? Our internal npm mirror does not carry `mermaid@11.16.1` (nor `cytoscape@3.34.1`, which this pulls in), so `bun install` fails outright on machines that install through it.

I verified `11.16.0` locally: `cytoscape` falls back to `3.34.0` on its own, no second pin needed, and `bun run lint` plus the three markdown suites stay green. The lockfile needs regenerating after the change.

decisions/2026/08/13/streamdown-markdown-mermaid.md
```diff
@@ -0,0 +1,21 @@
...
+
+Markdown previews and PR comments used a custom `marked` plus `sanitize-html` pipeline. Adding Mermaid to that HTML-string pipeline would also require custom lifecycle, loading, error, and theme handling.
+
+## Investigation
+
+Streamdown 2.5 provides GFM and raw HTML parsing, hardened sanitization, React component overrides, and an official `@streamdown/mermaid` plugin. Its sanitizer behaves consistently under the app's happy-dom tests, while source-backed React rendering also removes the old HTML rewriting used for repo-relative images.
+
+## Decision
+
+Use Streamdown for all shared Markdown rendering and its official Mermaid plugin with Mermaid 11.16.1. Dev3 continues to own external-link behavior, repo-relative image reads, theme token mapping, and rich-diff chunking; wide sequence diagrams retain readable SVG dimensions and scroll horizontally instead of shrinking, while `marked` remains only for splitting Markdown diffs into source blocks.
```
h0x91b: Version is written out here too — needs to read `11.16.0` alongside the `package.json` pin.

src/mainview/components/pr-review/markdown.tsx
```diff
@@ -1,53 +1,217 @@
...
+	return (
+		<Streamdown
+			key={`${rendererConfig.theme}:${isDocument}:${imageBaseDir ?? ""}:${imageRootDir ?? ""}`}
+			mode="static"
+			dir="auto"
+			className={`space-y-0${isDocument ? " dev3-md-doc" : ""}`}
+			components={components}
+			controls={false}
+			lineNumbers={false}
+			linkSafety={{ enabled: false }}
```
h0x91b, line 151: Not a blocker, just flagging it for whoever reads this next: with `linkSafety` off, every guarantee about link URLs rests on the remark/rehype pipeline plus the component override above. That is a reasonable trade, but it is exactly why the prefix round-trip matters — a one-line comment here saying what replaces link safety would help.

tailwind.config.js
```diff
@@ -1,6 +1,9 @@
 /** @type {import('tailwindcss').Config} */
 export default {
-	content: ["./src/mainview/**/*.{html,js,ts,jsx,tsx}"],
+	content: [
+		"./src/mainview/**/*.{html,js,ts,jsx,tsx}",
+		"./node_modules/streamdown/dist/*.js",
```
h0x91b, line 5: Nit / heads-up: the style build now scans a dependency's dist output, and `TaskDiffViewer.css` hooks onto `[data-streamdown="..."]` attributes. Both are coupled to Streamdown internals that no test covers, so a minor bump could silently break the layout. The exact `streamdown` version pin already helps; worth a sentence in the decision record so the next person knows the coupling is deliberate.

h0x91b (commented):
Hey — Claude here, the AI assistant working with Arseny. I pulled this branch down locally, got it installed and ran the checks, so this is a real read of the code rather than a drive-by.

First: this is a genuinely nice change. Replacing the hand-rolled `marked` + `sanitize-html` pipeline with one renderer that also owns Mermaid removes a whole class of HTML-string post-processing, and the fact that repo-relative images became React components instead of DOM rewriting is a real improvement. The decision record and the test coverage around sanitization are appreciated — most PRs this size ship with neither.

What I ran: `bun run lint` clean, and `markdown.test.tsx` / `markdown-diff.test.tsx` / `markdown-images.test.tsx` all green (after pinning `mermaid`, see below).

There is one finding I would call blocking, plus a few smaller notes — all inline:

1. **Blocking:** the `__markdown_link__` prefix round-trip lets a comment author put an arbitrary URL into an `<a href>`, bypassing Streamdown's link hardening. Repro is in the inline comment.
2. Same pattern on the image side, lower impact.
3. `mermaid` needs pinning to `11.16.0` — `11.16.1` is not on our internal mirror, so `bun install` fails there. This is the one that blocks us from running the branch at all.
4. Mermaid is mocked in every test, so nothing exercises real diagram-label sanitization — worth one test with the real thing, or an explicit note in the decision record that it is uncovered.
5. A couple of nits on the Tailwind/CSS coupling to Streamdown internals.

Could you also rebase on `main`? The branch is a fair way behind at this point, and I'd like the checks to run against current `main` before this goes in.

Happy to re-check once the fixes land — just ping me on the PR.

src/mainview/components/pr-review/markdown.tsx
arditti: Fixed — and thank you, the repro was exact.

`markdown-urls.ts` now owns the whole mechanism. The token prefix carries a nonce drawn once per session from `crypto.getRandomValues`, so a prefix an author types never matches. On top of that `protect` accepts nothing but a scheme-less relative URL and `unprotect` re-checks that after decoding, which makes "a token can only ever yield a path, never a scheme" an invariant of the module rather than something the call sites have to get right.

I did try your suggestion first — a registry of URLs the plugin actually issued, per render. It does not work, and the reason is worth knowing: **Streamdown reuses its unified processor across component instances.** The remark plugin captured by the first mount keeps running for every later one, so a second instance issues into its own registry and reads tokens minted by the first. Every relative link renders correctly the first time in a process and silently loses its href from the second render on — my own suite caught it, which is how I found it. A nonce needs no shared state and cannot go stale that way.

All four payloads from your repro now stay out of `href` (test: "keeps unsafe schemes out of href, written plainly or wrapped as a token"), and I checked it in the running app too, not just under happy-dom.

One thing I found while in here: with the token bypass closed, `views://` written plainly is already dropped, but by `rehype-sanitize`'s protocol list — `rehype-harden` itself runs with `allowedProtocols: ["*"]` and returns any non-http(s) scheme as-is. So the scheme allowlist is doing real work in one place: `unprotect` hands the `a` override an href the sanitizer never inspected. `safeLinkHref` restates the policy exactly there.

src/mainview/components/pr-review/markdown-images.tsx
arditti: Fixed by the same change — `markdown-images.tsx` no longer knows the prefix exists. It receives an already-resolved `src` and only treats it as disk-backed when `isRelativeUrl` holds, so a hand-written token is just a remote URL that fails to load. Test: "ignores a hand-written token instead of reading it off disk" asserts `readFilePreview` is never called for your `..%2F..%2F..%2Fetc%2Fpasswd` payload.

package.json
arditti: Pinned to `11.16.0`. The direct pin alone was not enough, though — `streamdown` and `@streamdown/mermaid` both depend on `mermaid` by range (`^11.12.2`), so the lockfile still resolved `mermaid@11.16.1` and `cytoscape@3.34.1` underneath them and your mirror would still have failed. There is now an `overrides` block pinning both across the tree; the lockfile is regenerated and contains exactly one `mermaid@11.16.0` and one `cytoscape@3.34.0`. That also drops the second copy of Mermaid from the bundle.

decisions/2026/08/13/streamdown-markdown-mermaid.md
arditti: Updated to `11.16.0`, and the record now says why the pin is doubled up in `overrides` rather than leaving it as a bare version.

src/mainview/components/pr-review/markdown.tsx
arditti: Added — the comment now says what link safety was doing (a confirmation modal per external link), that the hardener still runs, and that `unprotect` plus `safeLinkHref` in `markdown-urls.ts` are what decide the final `href`. The mechanism itself is written up in `decisions/2026/08/17/relative-markdown-urls-are-nonce-tokens.md`.

tailwind.config.js
arditti: Noted in the decision record's Risks section: both the Tailwind content glob over `node_modules/streamdown/dist` and the `[data-streamdown="…"]` selectors in `TaskDiffViewer.css` are called out as deliberate coupling to Streamdown's distributed output that no test covers, which is why the version is pinned exactly.

arditti:
Thanks — that was a genuinely useful review, and the blocking one was real. All five points plus the rebase are in, pushed as one follow-up commit on top of the original.

**1. Sanitizer bypass (blocking).** The prefix round-trip is gone. `src/mainview/components/pr-review/markdown-urls.ts` now owns the mechanism end to end: the token prefix carries a nonce drawn once per session, `protect` accepts nothing but a scheme-less relative URL, and `unprotect` re-checks that after decoding — so a token can only ever yield a path, never a scheme. `safeLinkHref` then restates the sanitizer's scheme allowlist at the point the `a` override writes `href`, since that value is one `rehype-sanitize` never inspected.

Worth flagging: I tried your suggestion (a per-render registry of URLs the plugin actually issued) first and it does not work — **Streamdown reuses its unified processor across component instances**, so the plugin captured by the first mount keeps running for later ones and a second registry never sees the tokens. Relative links render fine on the first render in a process and silently lose their href from the second on. Details in the inline reply and in the decision record.

**2. Image side.** Same fix; `markdown-images.tsx` no longer knows the prefix exists. A hand-written token is just a remote URL now, and a test asserts `readFilePreview` is never called for it.

**3. `mermaid` pin.** `11.16.0` — but the direct pin alone would not have unblocked you: `streamdown` and `@streamdown/mermaid` both depend on `mermaid` by range, so the lockfile still carried `11.16.1` and `cytoscape@3.34.1` underneath them. There is now an `overrides` block pinning both across the tree, so the lockfile holds exactly one `mermaid@11.16.0` and one `cytoscape@3.34.0`. Side benefit: one Mermaid in the bundle instead of two.

**4. Mermaid label sanitization.** A real-Mermaid test is not possible here — `mermaid.render` returns an *empty* SVG under happy-dom, which has no SVG layout, so the test would assert nothing. I checked it in a browser instead and wrote the result into the decision record: a label holding ``, `<script>` and a `javascript:` anchor renders with the handler, the script and the href gone, and nothing executes. A bare `` tag does survive as an inert element — the same exposure a plain markdown image already carries in this app.

**5. Tailwind/CSS coupling.** Called out in the record's Risks as deliberate and untested, which is why the version is pinned exactly.

**Rebase.** Rebased onto current `main` (was 52 behind). `bun run lint` clean; renderer suite 257 files / 3908 tests green. The only red locally is `src/bun/__tests__/native-pane-owner-forward.test.ts` (10 tests), which fails identically on a clean `origin/main` checkout on this machine — pre-existing, not from this branch.

Two things that came out of the rebase and are worth a look:

- The PR's global `IntersectionObserver` stub (`test-setup.ts`), which reported every observed element as visible so Streamdown would render Mermaid, also defeated the lazy batching in `Changelog` and `TaskDiffViewer` — it broke `Changelog.test.tsx`'s "renders day groups incrementally" once current `main` was underneath it. It is now an opt-in `test-utils/immediate-intersection.ts` used only by the markdown suites, and `test-setup.ts` is back to what `main` has.
- New tests: `markdown-urls.test.ts` for the URL policy, plus your repro payloads as regression tests at the component level for both links and images.

Two decision records: the mechanism is in `decisions/2026/08/17/relative-markdown-urls-are-nonce-tokens.md`, and the existing `streamdown-markdown-mermaid.md` picked up the version, pinning rationale, coupling and Mermaid-coverage notes. Ready for another look whenever you have a moment.

h0x91b:
Claude again — re-checked everything on `9f12f7f`, and this is a really good turnaround. Thank you for taking the security note seriously rather than patching around it; the nonce approach is a better answer than what I suggested, and moving the whole thing into `markdown-urls.ts` so "a token can only ever yield a path" is a module invariant is exactly the right shape.

What I verified locally, not just read:

- **The bypass is closed.** I re-ran my original payloads (`data:text/html,<script>`, `vbscript:`, `file:///etc/passwd`, `views://`) through `CommentMarkdown`, both written plainly and wrapped in the old prefix. Written plainly they produce no anchor at all; wrapped, the token stays an inert `https://dev3.invalid/...` URL and the payload never reaches `href`. The image variant no longer drives a disk read either.
- **No token leakage.** Rendered a document with a relative link, a relative image, an in-page anchor and an external link: nothing containing `dev3.invalid` survives into the DOM, and relative links and images still resolve normally (`docs/guide.md`, image title `shots/board.png`). So there is nothing for an author to copy back.
- **Install and pins.** `bun install` now goes through cleanly on our mirror. Nice catch on the `overrides` — I had only asked for the `dependencies` pin and missed that Streamdown and `@streamdown/mermaid` both take `mermaid` by range; the lockfile now carries a single `mermaid@11.16.0` and `cytoscape@3.34.0`.
- **Checks.** `bun run lint` clean, the six `pr-review` suites green (84 tests), and the full suite green apart from one unrelated known flake in `App.test.tsx` (streamer-mode Cmd+2) that passes 129/129 in isolation and is untouched by this branch.
- **Docs.** The Mermaid-coverage gap and the Streamdown-internals coupling are both written down honestly in the risks section now, including what you checked in a real browser and the one thing that does survive. That is more useful than a test would have been.

Turning my repro into the regression tests in `markdown-urls.test.ts` and `markdown.test.tsx` was the right move — if this ever regresses, it fails loudly.

Nothing further from me. Over to @h0x91b for the merge call.

h0x91b:
Claude here again (the AI that did the two review passes) — merged as `3889e94`.

Nothing changed in the code since my re-review: the branch content on `852f2df` is byte-identical to `9f12f7f`, the only delta was the rebase onto current `main`. All 11 checks were green on that head, so I only resolved the six threads and squashed it in.

For the record, the thing that was actually holding the merge was those unresolved conversations, not any check — my comments, so only I could close them. Sorry for the extra wait. Thanks again for the nonce rewrite and the regression tests; that turned a real bypass into something the suite now guards.

PR #1344, opened 2026-08-13
PR origin-task deep link: opt-out setting + #1340 review fixes

### LINKED ISSUE #1340: PR origin-task deep link: add an opt-out setting + review follow-ups from #1339
Hi, this is Claude (the AI assistant) — Arseny asked me to review #1339 and file the follow-ups here.

#1339 is being merged as-is so the feature can be exercised end to end. This issue collects one new requirement plus the review findings, all against the merged code.

## 1. Opt-out setting (new requirement, highest priority)

Not everyone wants a public PR to advertise that they use dev3, so the footer must be switchable.

- Add a `GlobalSettings` flag (e.g. `prOriginTaskLink?: boolean`, `src/shared/types.ts:851`), defaulting to **on** — same `?? true` pattern as `importShellEnv`.
- Surface it in `src/mainview/components/global-settings/BehaviorSettingsSection.tsx` via `SettingsToggle`, with i18n keys in `en` / `ru` / `es`.
- When off, `createPullRequest` (`src/bun/rpc-handlers/git-operations.ts:568`) sends the plain prompt with no deep-link block.
- Mention the escape hatch in the footer itself, in parentheses — something like `(this footer can be turned off in dev3 settings)` — so a reader who does not want it knows where to go.

The decision record already listed this toggle under "Alternatives considered" as deferred (`decisions/2026/08/13/pr-origin-task-deep-link.md`); this issue un-defers it.

## 2. The block starts with `---`, which can render as a heading

`buildTaskPrDeepLinkSection` (`src/shared/deep-link.ts:107`) opens with `---`. In Markdown, `---` directly under a line of text is setext syntax: the preceding line becomes an `<h2>` and no horizontal rule is drawn. The prompt only says "append at the very end of the PR description", so the tight-append case — the common one — is exactly the broken one.

Fix: start the section with a leading newline, or state "preceded by a blank line" in the instruction. The test asserting `startsWith("---\n")` (`src/bun/__tests__/deep-link.test.ts`) needs updating with it.

## 3. Multi-line prompt may submit early on non-Claude agents (unverified)

This is the first prompt on the agent-handoff path that contains literal newlines. Text reaches the pane as raw bytes (`sendKeysGuarded` → `send-keys -H`), and the only byte filtered on the way is `U+0000` (`src/shared/pane-input.ts:515`). Claude Code absorbs a fast newline into its input box (see the comment at `src/bun/agent-prompt.ts:169`), but dev3 also supports Cursor Agent, Codex, Gemini CLI and OpenCode, and none of them were checked.

If any of them treat `0x0a` as Enter, the handoff submits halfway and the rest of the Markdown lands as a second, junk prompt.

Cheapest check that would disprove it: send a two-line `dev3 message` into a Codex or Gemini task and watch the pane. If it does submit early, flatten the block to a single line or deliver it as a separate stage.

## 4. `docs/open.html` — param dispatch order makes `new-task` unreachable

The page checks `task`, then `project`, then `new-task`. A URL like `?new-task&project=p1&text=hi` therefore resolves to `dev3://project/p1`, and the `new-task` branch can never read its own `project` param. Harmless today (only `task=` links are generated), but the page is the public entry point for the whole grammar. Check `new-task` first, or key off an explicit `kind` param.

## 5. Asymmetric encoding of the task id

`buildTaskWebLink` runs the id through `encodeURIComponent`; `buildTaskDeepLink` does not; `open.html` decodes and re-encodes it. For an id containing `/`, the two links in the same footer resolve to different targets (`dev3://task/a%2Fb` vs `dev3://task/a/b`, which `parseDeepLink` strips). Latent while ids are UUIDs, but the new test pins the mismatch as expected behavior. Encode on both sides, or document that ids are UUIDs and drop the encoding.

## 6. The URL grammar now lives in three places, one of them untested

`src/shared/deep-link.ts` states in its header that keeping build and parse together "guarantees they never drift". The JS copy inside `docs/open.html` is a third copy with no test and no guard, so a grammar change in TypeScript would pass CI while the public page keeps redirecting to the old shape. Minimum: a test asserting the page contains the literal `dev3://task/` prefix. Better: generate the page's script from the module.

## Nitpicks

- `open.html` — `navigator.clipboard.writeText(...).then(...)` has no `.catch`: on denial the button silently never says "Copied", plus an unhandled rejection.
- `open.html` — the Open button starts as `href="#"` and only comes alive via JS. Harmless on a JS-only page, but `/` would be a better initial value.
- The page handles `project` and `new-task`, yet only `buildTaskWebLink` exists — speculative surface with no caller.
- User-visible behavior changed (every PR now carries a footer) and `ask-dev3` was not updated. Optional, but worth a line once the toggle exists.

## Checked and fine

No `javascript:` injection is possible on the page — every branch prefixes `dev3://` itself and the id goes through `encodeURIComponent`; the raw link is written with `textContent`. Prompt length is also fine: ~550 bytes against the tmux `if-shell` "command too long" wall at roughly 4500. Changelog entry, decision record, `Short:` line and tests are all present and correctly shaped, and no tip was added, which is the right default here.

BnayaZil:
Closes #1340.

Follow-ups to the merged PR-deeplink feature (#1339), against merged code.

## 1. Opt-out setting (new requirement)

`GlobalSettings.prOriginTaskLink`, **default on** — only an explicit `false` is stored (same shape as `importShellEnv`). Surfaced as a **"Task link in pull requests"** toggle in **Global Settings → Behavior**, mirroring `suggestCompletingTasksAfterMerge` (planned through `/ux-principal`: manifest-compliant, durable configuration → Settings; no new surface). When off, `createPullRequest` sends the plain PR prompt with no footer. The footer copy names the escape hatch (`… (this footer can be turned off in dev3 settings)`).

## 2. `---` no longer renders as a heading

`buildTaskPrDeepLinkSection` now starts with a blank line before `---`. Appended straight after the description, `text\n---` was setext syntax (H2, no rule); a blank line makes it a horizontal rule. Test updated from `startsWith("---\n")` to assert the blank-line separation.

## 3. Single-line handoff prompt (multi-line early-submit risk)

`createPrAgentPrompt` no longer embeds the multi-line markdown block. It appends the newline-free `buildTaskPrDeepLinkLine` and describes the `---` divider in words, so the whole prompt is one line — it can't submit early on agents that read `\n` as Enter. Guarded by a test asserting the prompt contains no `\n`.

## 4. `open.html` dispatch order

`new-task` is now matched **before** `project` (it is the only kind that also reads `project`), so `?new-task&project=p1&text=hi` resolves to `dev3://new-task?project=p1&text=hi` instead of `dev3://project/p1`.

## 5. Symmetric id encoding

`buildTaskWebLink` and `open.html` no longer percent-encode the id (ids are UUIDs), matching `buildTaskDeepLink` — both footer links now resolve to the identical target.

## 6. Grammar guard for `open.html`

New test (`open-html-grammar.test.ts`) asserts the page still emits every `dev3://` prefix and dispatches `new-task` before `project`. (The "generate the page from the module" option is noted as future work in the decision record — this is the minimum guard the issue asked for.)

## Nitpicks

- `open.html` clipboard `writeText(...)` now has a `.catch` (no silent failure / unhandled rejection).
- The Open button starts at `href="/"` instead of `#`.
- Re-lands the agent-skill guidance (corrected for #2) so **agent-opened** PRs (`gh pr create`) carry the footer too, and adds a line to `ask-dev3` about the footer + the opt-out.

## Verification

- `open.html` driven headlessly across all grammar cases — task / project / new-task / the `?new-task&project=` ambiguity / empty-state — each builds the correct `dev3://` link. Verbatim ids confirmed.
- Backend + renderer tests covering every touched area pass: `deep-link`, `settings`, `open-html-grammar`, `rpc-handlers` (createPullRequest on/off + single-line), `BehaviorSettingsSection` (default-on + opt-out), `GlobalSettings`, `settings-registry`.
- `bun run lint` is clean on the changed files.

## How to check the work

1. **Settings → Behavior → "Task link in pull requests"** — on by default. Toggle it off, open a PR from a task: the description has no `🔗 Origin task…` footer. Toggle on: the footer returns.
2. **Open a PR (button or ask the agent to `gh pr create`)** — the description ends with a `---` rule then `🔗 Origin task in dev3: open in dev3 · dev3://task/<id> (this footer can be turned off in dev3 settings)`. The `---` is a rule, not a bold heading.
3. **`docs/open.html?new-task&project=p1&text=hi`** — hands off to `dev3://new-task?project=p1&text=hi` (not `project/p1`). No params → "Nothing to open".

## Testing notes

Unchanged offline-sandbox caveat: the dev-server can't build here (missing `posthog-js`/`recharts`/`sanitize-html`, no network to `bun install`), so the toggle wasn't screenshotted in a running app — it's covered by the component tests instead and mirrors an existing toggle exactly. The full suite's only failures (`native-pane-owner-forward`, `updater`, `epipe`, and the module-missing renderer files) reproduce identically on a clean `main` and touch none of the changed files. CI runs the full suite with deps.


h0x91b:
Bnaya — my fault entirely, and I'm sorry. I simply missed that you had already picked up #1340 and were on it. I wanted to cut a release, saw the list still open, and hacked it together myself in #1342, which is already merged. That's on me, not on you — I should have checked the open PRs first, or just asked you.

Reading yours now, it is better than mine in the parts that matter most: the single-line handoff prompt is the right call (my version left the newline risk on non-Claude agents wide open), and you also remembered to update the ask-dev3 skill and the deep-link section of the agent protocol, which I skipped entirely. Your toggle and settings-normalization tests are better than what I wrote too.

So I'm closing this one only because it now conflicts with what's already in main — not because it was worse. I'm opening a follow-up right away that takes your improvements over mine, and it credits you for them explicitly; the feature itself is yours from #1339 onward and the changelog says so.

Thanks for the work, and sorry again for the duplicated effort — that one is squarely mine.

h0x91b:
Follow-up is up: #1348. It carries your single-line prompt, the skill/ask-dev3 updates, your toggle and settings tests, and your in-place update of the original decision record. Credit is on the PR body, the changelog entry and the decision record — the feature is yours end to end, I just pushed the buttons.

Sorry again for stepping on your work here.

PR #1348, opened 2026-08-14
Keep the PR handoff prompt on one line and document the footer in the skill

h0x91b:
Takes the parts of @BnayaZil's #1344 that were better than the merged #1342. The origin-task deep link is his feature (#1339), and every idea in this PR is his — it exists only because #1344 had to be closed on a conflict after #1342 landed first.

From #1344:

- **Single-line handoff prompt.** `buildTaskPrDeepLinkLine` is the newline-free footer content and `createPrAgentPrompt` passes that, describing the `---` divider in words. The prompt is typed into the pane as raw bytes, so a newline could submit it early on an agent that reads `\n` as Enter — this removes the question instead of trying to answer it per agent. `buildTaskPrDeepLinkSection` still renders the full block for the skill and the tests.
- **Skill updates.** The dev3 agent skill (`agent-skill-content.ts`) and the ask-dev3 flow map (`agent-skills.ts`) now state the footer and where to switch it off, so a hand-written PR description matches what the button asks for.
- **Tests.** The toggle's default-on/opt-out behavior in `BehaviorSettingsSection`, the settings normalization case, and a guard that the assembled prompt contains no newline.
- **The original decision record** is updated in place to point at the follow-ups, as he had it.

Kept from #1342 where the two differed: the link stays last in the prompt (so the auto-merge sentence cannot end up inside the copied line), ids stay percent-encoded on both sides with a matching decode in `parseDeepLink`, and `?text=…` without `new-task` keeps working on the open page.

Decision record: `decisions/2026/08/14/pr-deep-link-single-line-handoff.md`.


BnayaZil:
❤️

BnayaZil:
I don't care for the credit just excited to have links back to tasks in dev3 🥳🤩

PR #1350, opened 2026-08-14
Publish the Windows build through the canary update channel

h0x91b:
Hi, this is Claude (the AI assistant that wrote this branch).

Windows has been packaged and launch-proven on every in-scope merge for a while, and none of it reached a human: the output was a 30-day CI artifact behind a GitHub login. This adds the missing publisher.

- `win-x64` joins `CANARY_PLATFORMS`. The token is `win`, not `windows`, because that is what `getPlatformPrefix()` builds the feed URL from and what names electrobun's build folder.
- New reusable `release-build-windows.yml` with exactly one caller: `build-win-x64` in `canary-publish.yml`, gated on `windows-proof` like every other publisher. **`release.yml` gets no Windows caller** — the build is unsigned and will not be signed, so it goes where the stakes are low. The release-page leg is a separate task.
- `create-release-artifacts.sh` gains a `win` branch **and an OS allowlist**. The allowlist is the load-bearing half: `windows` previously fell through to the linux branch and would have published a well-formed `canary-windows-x64-update.json` that no client ever fetches.
- `package-windows-launched-tree.ts` publishes the human-installable `.zip` built from `retainedUnpackDir` in the launch proof, so it structurally cannot zip a tree nothing launched. This matters more than it reads: the build is **not byte-reproducible** — three runs on sha `068ea95df` produced artifacts 15 bytes apart — so re-extracting the archive would not even yield another run's bytes.

Two files are published per run and they are not interchangeable: the `.tar.zst` the updater downloads (Windows `tar.exe` cannot read zstd, so a human cannot open it), and the `.zip` a human extracts with nothing installed.

**Unsigned, measured on real hardware rather than predicted.** Artifact `windows-app-068ea95df…` (run 31781800859, v1.44.0) on an x64 Windows box: first launch shows a full-screen blue SmartScreen warning, two clicks to pass it (`Подробнее` → `Всё равно запустить` on a ru-RU machine), and the app then runs normally — title bar `[CANARY] dev-3.0 v1.44.0`, board renders, a task ran a Claude Code agent over PowerShell. The dialog's title text was not captured and is not quoted anywhere, because it is localised. The download instructions in the run summary name the click path and say the warning means *unrecognised publisher*, not *virus found*.

Rationale, risks and the two ordering traps found on the way are in `decisions/2026/08/08/windows-ships-through-canary-first.md`.

**Do not enable auto-merge on this PR.** `canary-publish` must be dispatched on `main` immediately after the merge: adding `win-x64` to the list is what unlocks the channel on Windows via `canaryPublishesFor()`, and no `canary-win-x64-update.json` exists until a run publishes one.

### DIFF
--- a/.github/workflows/canary-publish.yml
+++ b/.github/workflows/canary-publish.yml
@@ -52,6 +52,7 @@ jobs:
       macos-x64: ${{ steps.probe.outputs.macos-x64 }}
       linux-x64: ${{ steps.probe.outputs.linux-x64 }}
       linux-arm64: ${{ steps.probe.outputs.linux-arm64 }}
+      win-x64: ${{ steps.probe.outputs.win-x64 }}
     steps:
       - uses: actions/checkout@v5
 
@@ -96,7 +97,8 @@ jobs:
       needs.decide.outputs.macos-arm64 == 'true' ||
       needs.decide.outputs.macos-x64 == 'true' ||
       needs.decide.outputs.linux-x64 == 'true' ||
-      needs.decide.outputs.linux-arm64 == 'true'
+      needs.decide.outputs.linux-arm64 == 'true' ||
+      needs.decide.outputs.win-x64 == 'true'
     uses: ./.github/workflows/windows-conpty-package.yml
 
   # ──────────────────────────────────────────────────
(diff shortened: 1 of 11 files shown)


h0x91b:
Claude here (the AI that wrote this branch) — correcting the record, because the description above now says something that did not happen.

The body says **"Do not enable auto-merge on this PR"**, and auto-merge is exactly what was used. That was Arseny's call, and the reason it was safe is narrower than the original instruction: all 11 checks were already green and the PR was `CLEAN` before auto-merge was enabled, so it merged in seconds. The ban existed for one failure mode only — CI going green while nobody is watching, leaving the channel unlocked on Windows with no manifest in the bucket for an unbounded stretch. That mode was not reachable here.

Merged `52729daea` at 09:39:37Z; `canary-publish` dispatched on `main` at 09:40:35Z (run 31789069907), 58 seconds later. The follow-up that actually matters is still open: `canary-win-x64-update.json` has to answer 200, and then a human has to launch the published `.zip` from the bucket — the CI proof covers the published bytes under CI, not on real hardware.

PR #1381, opened 2026-08-14
Ignore cached agent binary paths that no longer match the base command

### LINKED ISSUE #1380: Editing an agent's base command has no effect — stale cached binary path silently overrides it
## Bug

Changing an agent's **Base Command** in Settings → Agents (e.g. `claude` → `claude-codex`) shows as saved, but new sessions still launch with the old command.

## Root cause

The edit does save to `agents.json`. However, `checkAgentAvailability` auto-caches each agent's resolved binary path into `settings.json` → `agentBinaryPaths` (to survive the minimal PATH of macOS .app bundles, see `decisions/2026/03/16/agent-binary-detection.md`), and `applyBinaryPathOverride` in `src/bun/agents.ts` applied that cached path unconditionally at launch — silently replacing the edited base command with the previously resolved binary (e.g. `/Users/…/.local/bin/claude`).

The cache never self-heals: `resolveBinaryPath` prefers the saved custom path over a fresh PATH lookup, so re-checking availability keeps returning (and re-saving) the stale path forever.

## Reproduction

1. Open Settings → Agents so availability is checked (this caches the resolved path for `claude`).
2. Edit the Claude agent's Base Command to `claude-codex`.
3. Start a new task — the terminal launches `claude`, not `claude-codex`.

Verified end-to-end by running `resolveCommandForAgent` against affected on-disk state: it returned `/Users/…/.local/bin/claude` despite `baseCommand: "claude-codex"`.

## Expected

A cached binary path should only apply while it still points at the binary the current base command names; an edited base command must win.

## Context

Found while pointing the agent at a `claude-codex` wrapper to test CLIProxyAPI with claude-code as the backing path — any workflow that renames/redirects the agent binary hits this.

arditti:
## Problem

Editing an agent's **Base Command** in Settings → Agents (e.g. `claude` → `claude-codex`) saved to `agents.json` but had no effect: new sessions still launched the old command.

Fixes #1380.

## Root cause

`checkAgentAvailability` auto-caches each agent's resolved binary path into `settings.agentBinaryPaths` (to survive the minimal PATH of macOS .app bundles), and `applyBinaryPathOverride` in `src/bun/agents.ts` applied that cached path unconditionally at launch — silently shadowing the edited base command. The cache never self-healed, because `resolveBinaryPath` prefers the saved custom path over a fresh PATH lookup.

## Fix

New pure helper `binaryPathMatchesCommand` (`src/bun/executable.ts`): a cached/custom agent binary path applies only while its file name still matches the agent's current base command (case-insensitive, tolerates Windows launcher extensions like `.exe`/`.cmd`). Enforced at all three consumers of `agentBinaryPaths`:

- `applyBinaryPathOverride` (`src/bun/agents.ts`) — launch command resolution
- `checkAgentAvailability` (`src/bun/rpc-handlers/settings-config.ts`) — availability display + path auto-save (self-heals: once the new command resolves, the fresh path overwrites the stale entry)
- pre-spawn binary check (`src/bun/rpc-handlers/tmux-pty.ts`) — a stale path no longer vouches for a binary that isn't installed

No data migration: stale cache entries become inert, and reverting the base command revalidates the old cache for free. Details in `decisions/2026/08/14/agent-binary-path-cache-staleness.md`.

## Verification

- Reproduced end-to-end against real affected on-disk state: `resolveCommandForAgent` returned `/Users/…/.local/bin/claude` before the fix and `claude-codex` after.
- New regression tests (fail on the old code): `binaryPathMatchesCommand` unit tests, `applyBinaryPathOverride` with a real temp binary, and a `checkAgentAvailability` case for an edited base command.
- `bun run lint` clean; full `bun run test` green (only pre-existing `native-pane-owner-forward` unix-socket failures, identical on a clean tree).

### DIFF
--- a/change-logs/2026/08/14/fix-stale-agent-binary-path-shadowing-base-command.md
+++ b/change-logs/2026/08/14/fix-stale-agent-binary-path-shadowing-base-command.md
@@ -1,0 +1,3 @@
+Short: Edited agent command now takes effect
+
+Fixed editing an agent's base command having no effect on new sessions: the auto-cached resolved binary path (settings.agentBinaryPaths) kept pointing at the previous binary and silently overrode the edited command at every launch. A cached path now only applies while it still names the same binary as the agent's current base command.
--- a/decisions/2026/08/14/agent-binary-path-cache-staleness.md
+++ b/decisions/2026/08/14/agent-binary-path-cache-staleness.md
@@ -1,0 +1,18 @@
+# Agent binary path cache is invalidated by base-command edits
+
+## Context
+
+`checkAgentAvailability` auto-saves each agent's resolved binary path into `settings.agentBinaryPaths` so launches survive the minimal PATH of macOS `.app` bundles (see `decisions/2026/03/16/agent-binary-detection.md`). That cache was applied unconditionally: after a user edited an agent's base command (e.g. `claude` → `claude-codex`), the edit saved to `agents.json` but every launch still ran the old cached binary — and the cache never healed, because `resolveBinaryPath` prefers the custom path over a fresh PATH lookup.
+
+## Decision
+
+A cached/custom agent binary path applies only while its file name still matches the agent's current base command (case-insensitive, tolerating Windows launcher extensions like `.exe`/`.cmd`). Implemented as `binaryPathMatchesCommand` in `src/bun/executable.ts`, enforced at all three consumers of `agentBinaryPaths`: `applyBinaryPathOverride` (`src/bun/agents.ts`, launch command), `checkAgentAvailability` (`src/bun/rpc-handlers/settings-config.ts`, availability + auto-save), and the pre-spawn binary check in `src/bun/rpc-handlers/tmux-pty.ts`.
+
+## Risks
+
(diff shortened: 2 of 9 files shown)


h0x91b:
Hi Eitav — Claude here, the AI assistant working on this repo with Arseny.

First: thank you. You found something genuinely annoying that we had lived with without understanding it — editing an agent's base command quietly doing nothing because the auto-cached resolved binary path kept shadowing it. Your diagnosis was right, and the fix was right.

I owe you an apology for how we handled it from there. I reviewed your branch, found one edge case, and instead of bringing it back to you here, I committed on top and opened #1386. That should have been a review comment on your PR. It was not intentional, and it was not how we want to treat someone who took the time to send a fix.

Where things stand:

- Your commit shipped in #1386, squashed into `main` as `f31601b6b`, and GitHub carried your `Co-authored-by: Eitav Arditti <eitav@amazon.com>` line into it, so you are in the history.
- What I added on top: a user-chosen custom binary path (Settings → Agents → Custom path) is now stored separately in `settings.agentCustomBinaryPaths` and is never name-checked against the base command, so it cannot be silently discarded by the new staleness rule. That was the one case where your check would have thrown away a path the user set deliberately.

Closing this one as shipped, with the credit belonging to you. If you spot anything else — in this area or elsewhere — please do send it; reports like yours are the hard part, the fix is the easy part.

h0x91b:
Shipped in #1386 (squashed to main as f31601b6b) with your co-author credit — closing here.

PR #1397, opened 2026-08-16
feat(telemetry): add a build-time VITE_TELEMETRY opt-out flag

d3p4z:
## Problem

There is currently no way to build dev3 with analytics turned off. GA4 (`src/mainview/analytics.ts`) initializes unconditionally from `main.tsx`, and PostHog (`src/mainview/posthog.ts`) turns itself on whenever `VITE_POSTHOG_KEY` + `VITE_POSTHOG_HOST` are present. There is no setting, no environment variable and no CLI flag for either. People who build from source — and anyone repackaging the app for a team or a distro — occasionally need to be able to say "this build sends nothing", and today they can only do that by patching the source.

## Design

One build-time flag, `VITE_TELEMETRY`, following the existing `VITE_POSTHOG_*` convention (root `.env`, documented in `.env.example`).

- **Default is `on`.** Anything other than the literal string `off` — including the variable being absent — behaves exactly as today. Nothing changes for existing users, source builds, or the release workflows.
- It is read through `telemetryEnabled()` in the new `src/mainview/telemetry.ts`, per call rather than in a module constant, so Vite still constant-folds it in a build while `vi.stubEnv` can flip it in tests.

## What `VITE_TELEMETRY=off` disables

| Gate | Why it is needed |
|---|---|
| `sendToGA()` | The single transport every GA4 event passes through — covers `first_visit`, `session_start`, `app_update`, `page_view`, `heartbeat`, `agent_launched`, `app_exception` without touching any `trackEvent` call site. |
| `initAnalytics()` | Gated **separately**, because three things never reach `sendToGA`: the `api.ipify.org` public-IP lookup that supplies `ip_override`, the 10-minute heartbeat `setInterval`, and the global `error` / `unhandledrejection` listeners installed by `setupErrorTracking()`. |
| `posthog.ts` | Forced onto the pre-existing no-op client **even when a key and host are configured**, so the flag is authoritative rather than "don't set a key". This also suppresses the DEV-mode "key missing" throw — a build that asked for no telemetry is not misconfigured. |

Feature flags keep working: the no-op client reports every flag as unset, which lands on `FEATURE_FLAG_DEFAULTS` in `src/shared/feature-flags.ts` — the same path a keyless source build already takes.

Nothing was removed and nothing else was refactored; this only adds the switch.

## How to verify

```bash
bun install
bun run lint
bun run test

# gate closed
echo 'VITE_TELEMETRY=off' >> .env
bun run build
grep -o 'function [A-Za-z0-9_$]*(){return!1}' dist/assets/index-*.js
grep -o 'if(![A-Za-z0-9_$]*())return;let [a-z]=String(Math.max' dist/assets/index-*.js
```

On my machine the helper compiled down to `function Xl(){return!1}`, with `if(!Xl())return;` sitting in front of both the `fetch(<GA endpoint>)` in `sendToGA` and the `resolvePublicIp()` call in `initAnalytics`. Building without the variable (or with `VITE_TELEMETRY=on`) produces `function Xl(){return!0}` instead, i.e. today's behaviour.

`posthog-js` is still statically imported and therefore still present in the bundle when off; it is simply never initialized (`Ql&&eu?…init(…):{capture:()=>void 0,…}` with `Ql` folded to `false`). Making it disappear entirely would require an async module export, which felt like more churn than this PR is worth — noted in the decision record.

## Tests

- `src/mainview/__tests__/analytics.test.ts` — a new `VITE_TELEMETRY=off` block: no request at all on init (neither the GA hit nor the ipify lookup), no heartbeat after an hour of fake timers, every explicit `trackEvent` / `trackPageView` / `trackDiffView` / `trackAgentLaunched` dropped, and sending resumes once the stub is removed. The rest of the file continues to exercise the live transport.
- `src/mainview/__tests__/posthog-telemetry.test.ts` — new: initializes when the flag is unset and when it is `on`; does **not** initialize with `off` despite a key and host being stubbed; the no-op client still answers the flag APIs.

Local run: `bun run lint` clean; `bun run test` green for mainview (255 files / 3848 tests) and cli (44 / 782). The bun project shows 10 failures in `src/bun/__tests__/native-pane-owner-forward.test.ts` (`listen EINVAL` — my temp dir blows past the macOS 104-char `sun_path` limit); they reproduce identically on a clean checkout of `main`, so they are environmental and unrelated.

## Also in this PR

- `change-logs/2026/08/16/feature-telemetry-opt-out.md`
- `decisions/2026/08/16/telemetry-build-time-opt-out.md` — why build-time and why the default is `on`
- A short **Telemetry** section in `README.md` stating what each channel sends and how to build with it off, plus the flag in `.env.example`. Happy to reword or drop that section if you would rather document it elsewhere.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,7 @@
 VITE_POSTHOG_KEY=your_posthog_project_token
 VITE_POSTHOG_HOST=https://your-posthog-host
 
-# Telemetry switch, read at build time. `on` (the default when unset) keeps
-# Google Analytics and PostHog as they are; `off` compiles both out of the
-# bundle, including the public-IP lookup GA4 uses for geolocation.
+# Telemetry switch, read at build time. The default when unset keeps Google
+# Analytics and PostHog as they are; `off` (also `false`, `0`, `no`) stops both
+# from running at all, including the public-IP lookup GA4 uses for geolocation.
 VITE_TELEMETRY=on
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,7 @@
 VITE_POSTHOG_KEY=your_posthog_project_token
 VITE_POSTHOG_HOST=https://your-posthog-host
+
+# Telemetry switch, read at build time. `on` (the default when unset) keeps
+# Google Analytics and PostHog as they are; `off` compiles both out of the
+# bundle, including the public-IP lookup GA4 uses for geolocation.
+VITE_TELEMETRY=on
--- a/README.md
+++ b/README.md
@@ -326,7 +326,8 @@ The app reports usage analytics so the project can see what is actually used:
 
 Both are compiled in at build time and can be compiled out. Build from source with
 `VITE_TELEMETRY=off` in the repo-root `.env` (or in the environment) and no analytics code runs:
-no GA4 hits, no public-IP lookup, no PostHog client, no error autocapture. Feature flags then
+no GA4 hits, no public-IP lookup, no PostHog client, no error autocapture (`false`, `0` and `no`
+work the same way; crash logging into the app's own local log file is unaffected). Feature flags then
 fall back to their shipped defaults. Leaving the variable unset means `on`, which is what the
 released binaries do.
 
--- a/README.md
+++ b/README.md
@@ -315,6 +315,26 @@ rate-limit tracking and skill directories differ per agent — the full grid is 
 | [Diagnostic logs](docs/diagnostic-logs.md) | What is logged locally, for how long, and what is redacted |
 | [AGENTS.md](AGENTS.md) | Architecture and contributor guide |
 
+## Telemetry
+
+The app reports usage analytics so the project can see what is actually used:
+
+| Channel | What it sends |
+|---|---|
+| Google Analytics 4 | App launch, version, OS, screen resolution, language, screen navigation, a 10-minute heartbeat, agent launches, and unhandled errors. Identified by a random per-install id, geolocated via a public-IP lookup against `api.ipify.org`. Project *names* and file paths are deliberately never sent — only internal ids. |
+| PostHog | Task and project actions (created, moved, merged, pushed …) plus unhandled errors, under a random per-install id. Also delivers feature flags. Live in the official release builds; a build from source has no key and stays off. |
+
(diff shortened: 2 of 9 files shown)


h0x91b:
Hey — Claude here, the AI assistant that reviewed this branch for Arseny.

First off: thank you for this. It is your first contribution to dev3, and it is a genuinely good one — a switch people building from source have been missing, with a decision record, a changelog entry, README coverage and tests all in the same PR. That is more care than most first PRs get, and the write-up in the description made the review fast. Please keep them coming.

The review came back clean on the substance: both gates are in the right places, `initAnalytics` really is the only caller of `resolvePublicIp()` and `setupErrorTracking()`, every `track*` helper goes through `sendToGA`, and the repo-root `.env` genuinely reaches Vite (`envDir: "../.."` was already there). Two small things came out of it, and since they were tiny I pushed them straight onto the branch rather than sending you round another cycle:

**1. `VITE_TELEMETRY` now accepts `off`, `false`, `0` and `no`, trimmed and case-insensitive.** With the strict literal check, someone repackaging the app who writes `VITE_TELEMETRY=false` would ship telemetry on and have no way to notice. For a privacy switch that is the expensive direction to be wrong in, so the spelling is forgiving now. The trade is that the helper may no longer minify to a bare `return !1` — noted in the decision record; the early returns still carry the guarantee.

**2. `setupErrorTracking()` moved in front of the gate in `initAnalytics`.** Those listeners turned out to be two features in one: the GA `app_exception` event *and* `logToBackend` → `logRendererError`, which writes the app's own local log file and sends nothing anywhere. As written, `VITE_TELEMETRY=off` also cost the build its crash diagnostics. Now the listeners always get wired and the GA event they raise is dropped by `sendToGA` like any other, so "no telemetry" no longer means "no local error log".

Plus a `it.each` in each of the two test files for the extra spellings, and one-line touch-ups to the README, `.env.example`, changelog and decision record so the wording matches (the "compiles out of the bundle" phrasing was promising more than the code does — your PR description already had this right, it was just the in-repo copy that overstated it).

`bun run lint` is clean and the full suite is green — mainview 3858, bun 5878, cli 782, zero failures. Enabling auto-merge now. Thanks again!

Issue #1347, opened 2026-08-14
Remove the attention scope from the active-tasks sidebar (replaced by Spaces)


h0x91b:
The active-tasks sidebar has three scopes today: **project**, **global** and **attention** (the bell button with the pulsing count). Spaces (#1295) needs a scope of its own, and the attention scope is the weakest of the three — it duplicates something the search box already does, and the bell/count/popover cluster around it never felt right in daily use.

So: **the attention scope is deprecated as of now and will be removed in favour of a space scope.** The switcher becomes **global / space / project**, in that order.

The rest of the design lives in `decisions/2026/08/11/spaces-replace-the-per-project-board.md` and in #1295; this issue exists so the removal is visible on its own rather than buried in a feature PR.

## What is removed

| Where | What |
|---|---|
| `src/mainview/components/ActiveTasksSidebar.tsx` | `"attention"` from `SidebarScope`, the bell scope button, its pulsing count badge, the flat-list branch and the tier-header suppression that only that mode used |
| `src/mainview/i18n/translations/{en,ru,es}/common.ts` | `sidebar.scopeAttention`, `sidebar.noAttentionTasks` |
| `src/mainview/i18n/translations/{en,ru,es}/tooltips.ts` | `ttip.sidebar.scopeAttention` |
| `src/mainview/components/__tests__/ActiveTasksSidebar.test.tsx` | the four `sidebar-scope-attention` cases (scope contents, PR-review inclusion, count badge, empty state). The "attention bell inside the rail" case is about the per-task bell and stays |
| `localStorage` (`dev3-sidebar-scope`) | a stored `"attention"` value must fall back to the default scope instead of rendering nothing |

## What stays — deliberately

- **`is:attention`** — the search token and the funnel flag, backed by `isAttentionTask()` in `src/mainview/utils/taskFacets.ts`. The same list is one token away in the search box, so no capability is actually lost with the button.
- **`dev3 attention`** — the CLI ping, the red bell on the card, the reasons popover and the Focus Mode queue. Different feature that happens to share the word; it is how an agent calls the user when it is stuck, and it is not going anywhere.

## Note for whoever picks this up

"Deprecated" here means *scheduled for removal and safe to design around* — it does **not** mean a compatibility shim. `AGENTS.md` bans deprecation markers and dead branches: the PR that does this deletes the scope and rewrites every caller in one change.

[...]

Issue #1295, opened 2026-08-07
Proposal: Organize projects and live work with Spaces


arditti:
## Summary

I would like to align on a larger UX concept before starting implementation: an app-level **Spaces** layer for organizing projects and reviewing active work across them.

Spaces would contain project references only. They would not change project, repository, board, task, or worktree ownership.

## Problem

As the number of projects grows, the current project list becomes difficult to scan.

A single manual project order cannot represent contexts such as Personal, Client work, or Labs. Active agent work is also distributed across project boards, making it difficult to understand everything currently running.

## Proposed model

- A Space is an app-level organizational object.
- A project may belong to zero, one, or multiple Spaces.
- `All projects` is a virtual overview, not a Space.
- `No space` is a computed group containing projects without membership.
- Named Spaces and `No space` share a user-defined display order.
- Deleting a Space removes its memberships only. It never deletes projects or tasks.

For v1, I recommend one hierarchy level:

```text
Space
  Project
  Project
```

“Folder-like” describes expansion, collapse, and ordering. Nested Spaces could be considered separately if there is a concrete need.

## All Projects overview

The mock proposes one global overview with three areas:

1. A Space navigation rail.
2. A single project hierarchy pane.
3. A cross-project Live work sidebar.

There is no separate Unassigned view. `No space` appears as another expandable group in the hierarchy.

The hierarchy would support:

- Expandable Space rows with projects beneath them.
- The same project appearing under multiple Spaces.
- Search across Space names, project names, and repository paths.
- Matching a Space showing all its projects.
- Matching a project showing it under its relevant Spaces.
- A `Current work` summary on each Space row.
- Drag ordering for named Spaces and `No space`.
- Selecting a Space to focus the project list.

## Project creation and editing

This should extend the existing rich New Project screen rather than replace it.

- `New` offers `New project` and `New space`.
- New Project gains an `Add to Spaces` multi-select.
- Existing project editing uses the same multi-select component.
- Empty Spaces are allowed.
- Space membership does not affect repository or worktree locations.

## Live work

The overview includes a cross-project active-task sidebar with two grouping modes:

- **All:** group active tasks globally by lifecycle status.
- **Spaces:** group first by Space, then by lifecycle status.

With many-to-many membership, the current mock intentionally shows a project task in every Space containing that project.

## Explicit non-goals

This proposal does not:

- Create multi-repository projects or boards.
- Change the one-project/one-repository ownership model.
- Move tasks between projects.
[...]

PR #1401, opened 2026-08-16
Spaces: group projects without replacing boards

arditti:
Implements #1295, built to @arditti's proposal mock within the rules of `decisions/2026/08/14/spaces-group-projects-without-replacing-boards.md`. The mapping from every mock element onto those rules is recorded in `decisions/2026/08/17/spaces-dashboard-follows-the-proposal-mock.md` and was posted for review on the issue before the rebuild.

## The model

A **Space** is a named, many-to-many grouping of projects — a global tag, not a container. It owns nothing: every project keeps its own board, git never moves, `projectSlug()` and the `~/.dev3.0/` layout are untouched. **`Home` is computed** from the projects with no membership: never stored, not renameable or deletable, absent from every membership picker. A project in several spaces renders under each of them, and therefore never under `Home`.

## Surfaces

| Surface | What it does |
|---|---|
| Dashboard rail | `All projects` + each space with its project count + computed `Home` + `New Space`. Selecting one **filters the overview in place** — no route, no breadcrumb, no space board. Drag reorders spaces |
| Space header | Collapsible (persisted), project count, split `N need you · N working`, `+` to add existing projects, `…` for rename/delete, grip to reorder |
| Project row | Membership chips for its other spaces, plus the ADR's `Spaces…` action in the existing cluster. Reorder controls disappear where nothing can be reordered |
| Dashboard search | One field over project names **and** space names, same matcher as ⌘K |
| Active tasks | The **same `ActiveTasksSidebar` component** mounted with no current project: global scope, switcher hidden, `All / Needs you` presets that toggle only the `is:attention` token so other filters survive. In a project it gains a `space` scope (projects sharing a space with this one) |
| Task funnel | A `SPACES` group: `space:"Client X"` plus `is:home` for the computed group — a flag value, so it can never collide with a space actually named "Home" |
| Membership | Multi-select in Project Settings and New Project; a project created from a space's `+` lands in that space |
| ⌘K | Matches space names; the row label stays the project name |
| Streamer mode | A space is masked when any member project is sensitive — deliberately conservative |
| CLI | `dev3 current` gains `Spaces:` and read-only `Siblings:` paths |

## Data

New additive `~/.dev3.0/spaces.json` — lock + atomic write + identity-checked cache, `version: 1`, reserved `parentId: null`, soft delete, auto-soft-delete when a space's last member leaves (with a toast), dangling project ids skipped silently and never cleaned up. Nothing existing is renamed, moved or migrated; the file is not created until the first space exists.

## Zero spaces

The dashboard renders its pre-Spaces flat list: no rail, no search field, no task panel, no group headers, no chips, no `Spaces…` action — verified by parking `spaces.json` and reloading. **One honest deviation from "byte-identical":** the header carries a `New Space` button, because with no rail there would otherwise be no way to create the first space.

## Verification

Rebased on current `main`. `bun run lint` clean. Renderer **266 files / 3948 tests** green; CLI **46 files** green; backend **349/350 files** green — the single failure is `native-pane-owner-forward.test.ts`, which fails identically on a clean `origin/main` checkout on this machine (unix socket path length under the test root) and touches nothing here. ~90 new tests across persistence, RPC, grouping, rail, header menu, add-to-space dialog, facets, scope and CLI.

Manual browser QA on the rebased build: zero-spaces path as above; rail filter narrows to one group and leaves the route on `dashboard`; dashboard search matches by space name and by project name; collapse persists to `localStorage`; drag reorder persists to `spaces.json` and moves both rail and headers; `space:"Client X"` filters the panel; the `Needs you` preset appends `is:attention` while keeping `space:"Client X"`; rename and delete work with delete leaving all 9 project rows intact; `dev3 current` prints `Spaces:` + `Siblings:`; console clean throughout.

Docs: UX manifest (`PRODUCT_UX_BIBLE.md` §10, `ux-architecture.yaml` object model, one `UX_DECISIONS.md` entry), an `ask-dev3` entry so the feature map does not drift, and one tip (score 3) for the ⌘K space match.

## Known open questions for review

- The **rail** and the **sidebar-on-dashboard** placement are the two things your ADR argued against ("a space is not a place", "no new panel"). The addendum explains why filtering ≠ navigating and why reusing the component ≠ a second panel — but if either crosses a line, say so and I will rewrite that part.
- A project in several spaces still duplicates its triage rows; per your ADR that is the honest reading of many-to-many, and the thing to watch first in real use.
- Drag never changes membership, per the ADR. Dropping a project onto a rail space would only ever *add* an edge — happy to add it if you want it.

Builds on #1400 (#1347); those commits drop out of this diff once that merges.


src/mainview/components/__tests__/ProjectQuickSwitchModal.test.tsx
```diff
@@ -1,142 +1,57 @@
-import { beforeEach, describe, expect, it, vi } from "vitest";
-import { render, screen } from "@testing-library/react";
+import { render, screen, waitFor } from "@testing-library/react";
```
h0x91b, line 2: 🔴 **This file went from 10 test cases to 1.** The new space-matching test is a good one, but it replaced the file rather than joining it: gone are the focus trap, filtering, `Enter`, arrow-key navigation, the empty state, `Escape`, and click-to-select — plus the two cases that assert the `⌘N` and `⌘0` badges.

Those last two are the ones I'd really like back. "`Cmd+1..9` untouched" is an invariant this PR states in its own description and one the ADR leans on, and those tests were what guarded it. I checked: the cases don't exist anywhere else on the branch, and `ProjectQuickSwitchModal.tsx` itself changed only additively (+6/−0), so the behaviour is alive — it's the safety net that's gone.

Suggested fix: restore the original 10 and add the space-matching case alongside them.

src/mainview/components/NewSpaceModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+	projects: Project[];
+	initialProjectIds?: string[];
+	onClose: () => void;
+}
+
+/**
+ * Create a space with its first members in one flow — a space is never empty,
+ * so Create stays disabled until a name and at least one project are chosen.
+ */
+function NewSpaceModal({ projects, initialProjectIds, onClose }: NewSpaceModalProps) {
```
h0x91b, line 19: 🟡 **This component is imported by nothing.** `rg NewSpaceModal src --glob '*.tsx'` finds only the file itself and its 76-line test — creating a space in the running app is only possible through the picker inside Project Settings.

The ADR has the New Space dialog reachable from the dashboard *and* from Project Settings, so either it wants wiring up on the dashboard (which is also the more discoverable entry point — right now a new user has to go into a project's settings to create a grouping of projects), or the component and its test come out and the ADR line gets corrected. Either is fine by me; shipping it unreferenced is the one option I'd avoid.

src/mainview/components/ActiveTasksSidebar.tsx
```diff
@@ -437,6 +432,32 @@ function ActiveTasksSidebar({
...
+						<Tooltip
+							content={t("sidebar.scopeSpace")}
+							detail={siblingIds === null ? t("ttip.sidebar.scopeSpaceDisabled") : t("ttip.sidebar.scopeSpace")}
+							placement="bottom"
+						>
+							<button
+								type="button"
+								onClick={() => siblingIds !== null && setScope("space")}
+								aria-pressed={scope === "space"}
+								aria-disabled={siblingIds === null || undefined}
```
h0x91b, line 501: 🟡 Answering the question from your description: **hide it entirely when there are no spaces at all.**

Rendering it disabled is a permanent reminder of a feature the user hasn't opted into, and it's the one thing that breaks the "zero spaces means the app is exactly what it was yesterday" promise. `spaces.length === 0` → don't render the button.

Worth keeping the disabled state for the other case though: spaces exist, but *this* project is in none. There the button is meaningful and the tooltip explains what to do.

src/mainview/components/SpaceGroupedProjects.tsx
```diff
@@ -0,0 +1,271 @@
...
+									fill="none"
+									stroke="currentColor"
+									viewBox="0 0 24 24"
+								>
+									<path strokeLinecap="round" strokeLinejoin="round" strokeWidth={2} d="M19 9l-7 7-7-7" />
+								</svg>
+								<span className={`text-fg-2 text-sm font-semibold truncate ${masked ? MASK_CLASS : ""}`}>
+									{space.name}
+								</span>
+								<span className="text-fg-muted text-xs tabular-nums flex-shrink-0">
```
h0x91b: ⚪ The space name above gets `MASK_CLASS` when masked, but this counter doesn't — so on camera the client's name blurs while "4 · 11 tasks" stays perfectly readable next to it.

Small, but it's exactly the class of thing streamer mode exists for: how much work a private client has in flight. Wrapping this `<span>` in the same conditional class as the name should be all it takes.

h0x91b (commented):
Hey — Claude here, the AI assistant working with Arseny on the Spaces thread. I took this branch for a full walk rather than just reading the diff: checked out `01fcfaa`, ran the app, and went through every user story the ADR implies — zero spaces, first space, a project in two spaces, ordering, the sidebar ring, the ⌘K switcher, streamer mode, `dev3 current`, and the on-disk file after each step.

Short version: **this lands the ADR remarkably closely.** The "zero spaces = byte-identical app" promise holds in practice, `spaces.json` really isn't created until the first space, the file matches the schema byte for byte, membership survives a project remove/re-add, auto-soft-delete on the last member leaving fires with the toast, drag never touches membership, and `Cmd+1..9` is untouched. `bun run lint` is clean and the full suite is 10,584 green (one `TaskDiffViewer` flake that reproduces on clean `origin/main` too — not yours). #1400 I checked separately: exactly the list from #1347, nothing extra; the 53 deleted lines in `sidebarTiers.test.ts` are the tests for the `showHeader` field that no longer exists, and the rest of that file is intact. That one looks ready to me.

One blocker and a few smaller things, inline where they anchor:

**🔴 The palette test file lost 9 of its 10 cases** — details in the inline comment. That's the only thing I'd call blocking.

**🟡 `NewSpaceModal` is imported by nothing.** Written, tested, dead. The ADR has the New Space dialog reachable from the dashboard as well as from settings, and right now there is no dashboard entry point at all.

**🟡 The UX manifest wasn't touched.** `AGENTS.md` asks for `docs/ux/` to be updated when a surface appears, and three appeared here: space headers on the dashboard, the membership picker (two places), the third scope button.

**🟡 Your own open question — yes, hide it.** At zero spaces the third scope button should not render at all, not render disabled. "Nothing changed for you" should be literally true for someone who never makes a space. One-line condition, as you say.

**⚪ Streamer mode leaks the space's counts.** The name blurs, the `projects · tasks` count right next to it does not — inline comment on the line.

**⚪ Optional:** the `ask-dev3` skill and `tips.ts` are untouched. Matching a project by its space name in ⌘K is completely invisible, which is exactly what the tips registry is for. At most one tip, and only if you think it earns it — zero is a fine answer under our own policy.

One observation rather than a finding: with a project in two spaces, its triage rows really do render twice on the dashboard, and in a realistic list (22 projects) that reads heavier than it does on paper in the ADR. That's the honest consequence of many-to-many and the ADR already flags it as the thing to watch — collapsed headers do soften it. Not asking you to change anything; just confirming the risk is real and visible, so it's worth keeping an eye on once people use it.

Also on me, not you: the decision record both PRs cite is still sitting on an unpushed branch, so the path in your description doesn't resolve on `main` yet. I'll get it in.

Thanks for splitting this into two PRs and for the surgical shape — new modules with single hook points into the existing components made it genuinely quick to review, and the manual-QA notes in the description matched what I actually saw on screen. Nice work.

src/mainview/components/__tests__/ProjectQuickSwitchModal.test.tsx
arditti: Fixed in c9d717a3 — and thank you for catching it, this was the bad kind of mistake: I wrote the file instead of extending it, so ten passing tests were deleted to make room for one.

All ten are restored verbatim from `origin/main`, including both badge guards (`⌘N` from the board index, and the bracketed Operations name with `⌘0`) — you were right that those are the safety net for the "`Cmd+1..9` untouched" claim this PR makes. The space-matching case now sits alongside them in its own describe, so the file is 11 cases.

One necessary addition to the restored file: the component now calls `useSpaces()`, so the file gained a `vi.mock("../../rpc")` stub for `getSpaces` — without it every restored case would log a transport error. The stub also feeds the new case its space data.

src/mainview/components/NewSpaceModal.tsx
arditti: Fixed — it now has two entry points, both on the dashboard: **+ New Space** at the bottom of the spaces rail, and a **New Space** button beside *Add project* in the header (the latter is the one that exists at zero spaces, since with no spaces there is no rail to hold it).

This was part of a larger rebuild you have not seen yet: the review is against `01fcfaa`, and the branch has since been rebuilt to the shape of the original proposal mock — rail, group headers with split counts, membership chips on rows, a `Spaces…` row action, dashboard search over projects + space names, the reused sidebar mounted beside the overview, a `SPACES` group in the task funnel, and rename/delete on the header. The mapping from every mock element onto your ADR is recorded in `decisions/2026/08/17/spaces-dashboard-follows-the-proposal-mock.md` and was posted on #1295 before I started, flagging the rail and the sidebar placement as the two calls I most expected you to contest.

src/mainview/components/ActiveTasksSidebar.tsx
arditti: Done, exactly as you split it: `spaces.length === 0` → the button is not rendered at all, so a user who never opted in sees yesterday's two-button switcher. The disabled state is kept for the case you called meaningful — spaces exist but *this* project is in none — where the tooltip explains how to fix it.

Covered by two tests now: one asserting the third button is absent with zero spaces, one asserting it is present-but-disabled when spaces exist without this project.

src/mainview/components/SpaceGroupedProjects.tsx
arditti: Fixed, and the same leak was in the rail — masked name, perfectly readable count beside it. Both are covered now: the header's project count and the split `N need you · N working` counters, and the rail row's count, all take `MASK_CLASS` from the same conditional as the name.

Verified on the running app rather than just in tests: marked a `Client X` member sensitive, loaded with `&streamer=on`, and asserted there is **no** unmasked numeric node left in either the rail row or the header (`streamer-private` on both name and count). Then reverted the flag. Two unit tests lock it in — they fail if any numeric node in a masked row lacks the class.

arditti:
Thanks — that's a genuinely useful review, and the blocker was the right call. All four findings are addressed in c9d717a3, with replies inline. Summary:

| | Finding | Status |
|---|---|---|
| 🔴 | Palette test file lost 9 of 10 cases | **Fixed.** All ten restored verbatim from `origin/main` (both badge guards included), space case added alongside → 11. My mistake was writing the file instead of extending it |
| 🟡 | `NewSpaceModal` imported by nothing | **Fixed.** Two dashboard entry points: `+ New Space` in the rail, and a `New Space` button beside *Add project* (the one that exists at zero spaces) |
| 🟡 | UX manifest untouched | **Fixed.** `space` object in `ux-architecture.yaml`, two placement rows in `PRODUCT_UX_BIBLE.md` §10, one dated entry in `UX_DECISIONS.md` |
| 🟡 | Hide the scope button at zero spaces | **Fixed** exactly as you split it — absent at `spaces.length === 0`, disabled-with-tooltip when spaces exist but this project is in none |
| ⚪ | Streamer mode leaks the space's counts | **Fixed** in the header *and* the rail (same leak there), verified live with a real sensitive project |
| ⚪ | `ask-dev3` + a tip | **Both added.** Two `ask-dev3` entries (flow + vocabulary) and exactly one tip, score 3, for the ⌘K space match — everything else is self-describing |

**One thing to re-read before you trust your earlier walk:** your review is against `01fcfaa`, and the branch has since been **rebuilt to the shape of the original proposal mock**. @arditti's read of the first pass was that it followed your ADR's *rules* but not the mock's *shape* — headers over the existing rows and nothing else. So the dashboard now has the rail, group headers with split counts, membership chips, a `Spaces…` row action, a search field over projects + space names, the reused sidebar mounted beside the overview, a `SPACES` group in the task funnel (with `is:home` as a flag value, so it can't collide with a space named "Home"), and rename/delete behind one `…` on the header.

Every mock element is mapped onto your rules in `decisions/2026/08/17/spaces-dashboard-follows-the-proposal-mock.md`, posted on #1295 before I built it. The two I most expect you to contest are called out there and in the PR description: **the rail** (your ADR rejected "a space as a place" — the argument is that it filters in place and never routes, so no space ever gains a board) and **the sidebar on the dashboard** (your ADR said "no new panel" — the argument is that this is the *same component*, not a second one). If either crosses the line, say so and I'll rewrite that part rather than defend it.

Also worth knowing: the zero-spaces dashboard is **not** literally byte-identical any more — it carries the `New Space` button, because with no rail there'd be nowhere else to create the first space. Everything else (rail, search, panel, headers, chips, row action) is absent; I verified by parking `spaces.json` and reloading.

On your observation about duplicated triage rows at 22 projects: agreed, and it matches the ADR's own risk note. Collapsed headers soften it and the rail now lets you filter to one space, which helps more — but I've left the behaviour as the ADR specifies rather than quietly deviating.

Gates after the fixes: `bun run lint` clean, renderer **267 files / 3967 tests** green, CLI **46 files** green, backend **349/350** — the one failure is `native-pane-owner-forward.test.ts`, which fails identically on clean `origin/main` here (socket path length under the test root).

src/mainview/components/SpacesRail.tsx
```diff
@@ -0,0 +1,194 @@
...
+				draggedRef.current = null;
+				setDragged(null);
+				setDropTarget(null);
+			},
+		};
+	}
+
+	return (
+		<aside
+			className="hidden lg:flex w-56 flex-shrink-0 flex-col border-r border-edge overflow-y-auto py-4 px-3 gap-4"
```
h0x91b, line 153: 🟡 **The rail hides below `lg`, but the space it selected stays applied.** `selectedSpaceId` is React state in `Dashboard.tsx`; this is a pure CSS hide, so the component stays mounted and keeps filtering.

Reproduced in a browser rather than reasoned about: selected `Client X` at 1440px, then set the viewport to 1000px. The rail goes `display: none`, the dashboard still renders only the `Client X` group, and 19 of 21 projects are gone with no visible control saying why and no way to get them back. State is not persisted, so a reload clears it — but a user resizing their window on a big monitor (where `screen.width` stays wide, so no mobile path) just watches most of their projects vanish.

The layout priority is worth a second look too: at 1000px the 21rem task panel is kept and the 14rem rail is the thing dropped.

Cheapest fix is probably to clear the selection when the rail is not rendered — same `lg` breakpoint via a matchMedia hook, or reset `selectedSpaceId` on the narrow transition. Showing the active filter as a dismissible chip in the header would also work and is arguably nicer, since it fixes the "why is my list short" question at every width.

src/mainview/utils/spaceGroups.ts
```diff
@@ -0,0 +1,101 @@
...
+			? group.projects
+			: group.projects.filter((p) =>
+					projectHaystack(p.name, spaces, p.id).toLowerCase().includes(q),
+				);
+		if (projects.length > 0) matched.push({ space: group.space, projects });
+	}
+	return matched;
+}
+
+/** Local copy of the search haystack rule (name first, then space names). */
```
h0x91b: 🟡 The comment is honest about it, which is why I'd rather fix it than leave it: `utils/projectSearchHaystack.ts` already exports this exact rule, and the palette imports it. Your own record says "One matcher, one search language" — right now there are two.

They agree today (I checked: `spacesOfProject` filters `!deleted` and `projectIds.includes`, same as this local copy), so nothing is broken. It's purely that the next person to change the rule will change one of them.

Importing `projectSearchHaystack` here should be a straight swap.

decisions/2026/08/17/spaces-dashboard-follows-the-proposal-mock.md
```diff
@@ -0,0 +1,92 @@
...
+  memberships, cannot be renamed or deleted, is absent from every membership multi-select, and
+  its rows carry no chip. `spaces.json` still is not created until the user makes a real space,
+  and with zero spaces the dashboard renders its legacy flat list with no rail and no headers.
+- **A project in several spaces renders under each of them**, and therefore never under `Home`
+  (`Home` is the zero-membership group by definition). Chips on the row name every space it
+  belongs to.
+- **`+ New space` lives in the rail** — the dashboard entry point the previous record required.
+- **The dashboard gains one search field** over projects, matching space names through the same
+  haystack the ⌘K palette uses. One matcher, one search language.
+- **A space header carries** a neutral letter badge, its project count, split
```
h0x91b: ⚪ Stale within the same PR: commit `7e355bd5` ("Drop the letter badges from space rows and headers") removed them, so the header now carries the name, counts, grip, `+` and the menu — no badge. Confirmed on screen.

Worth correcting here rather than leaving a record that describes a thing the code deliberately dropped — that is exactly the kind of drift these records exist to prevent.

h0x91b (commented):
Claude again — second pass, this time over the rebuilt dashboard.

**All four threads from the first review are fixed and resolved.** I checked each one in the code rather than taking the reply for it: the palette file is main's version plus your `getSpaces` stub plus one new describe, so all ten cases including both badge guards are back verbatim (11 total); `NewSpaceModal` is mounted from `Dashboard.tsx` with the two entry points you describe; the zero-space scope button is gated on `spaces.length > 0` with a test per branch; and the masking now covers the counts in *both* the header and the rail, with tests that fail if any numeric leaf in a masked row lacks the class. That last one you extended past what I asked — good catch on the rail.

You also closed the two loose ends from my review body without being asked: the UX manifest, the `ask-dev3` map and a tips entry all landed in `5267d208`.

## How I checked this round

Ran the branch in a real browser at 1440×900 and walked it end to end: zero-spaces baseline (no rail, no panel, no `spaces.json` on disk) → created `Client X` from the dashboard button → rail appeared with the right counts → created `Labs` sharing one project → chips rendered on the shared row under each group, correctly omitting the group's own space → `Spaces…` row picker → header menu rename/delete → `space:"Client X"` in the sidebar search (24→18 needs-you, 5→4 waiting) → deleted both spaces → dashboard returned byte-for-byte to its legacy self.

The file on disk matched the schema at every step, the delete confirmation is in-app (not a native dialog) and says the right thing — *"Its projects stay in dev3 and only lose this grouping"* — soft delete set `deleted: true` and dropped the id from `order`, and no project was touched. **Zero console errors** across the whole walk. Rebased on current `main` locally: `bun run lint` clean, **10,768 tests green** (4008 renderer / 5959 backend / 801 CLI).

The shape is genuinely good. Nothing here blocks a merge.

## What I found

One of these I reproduced live; the rest are read from the code.

**🟡 A selected space stays applied after the rail is hidden** — inline comment, this is the one I'd most like fixed.

**🟡 Two copies of the search haystack** — inline comment. Functionally identical today, so it's drift risk rather than a bug, but the record promises one matcher.

**🟡 The in-app search help never learned the new tokens.** `help.ts` spells the vocabulary out in two places — `priority:P0 label:"Bug Fix" agent:Codex status:review is:attention has:port` — and neither gained `space:` or the new `is:home` flag. You updated `ask-dev3` and the tips registry, so this is the one surface left where the new tokens are invisible to a user who opens Help.

**⚪ Two different project counts on one screen.** The dashboard heading says `22 projects`, the rail says `All projects 21` — the heading counts the builtin Operations board and the rail excludes it. Both are defensible alone; side by side they read as a bug.

**⚪ The decision record is already stale.** `decisions/2026/08/17/…mock.md` still says the space header carries "a neutral letter badge" — inline comment; commit `7e355bd5` removed them.

**⚪ Changelog date.** `change-logs/2026/08/17/` — worth moving to whatever day this actually merges, since the changelog UI groups by ship date.

## Two things that are Arseny's call, not findings

You flagged both yourself and you were right to: **the rail as a filter** and **mounting `ActiveTasksSidebar` on the dashboard**. Both work exactly as you describe and neither breaks a rule the ADR settled — but they are the two places the dashboard gains real chrome, and that is a design decision rather than a correctness one. Leaving them for Arseny to sign off rather than ruling on them in review.

## One ask

**Please rebase on `main`** — #1400 merged as a squash, so this branch's three copies of those commits now conflict and the PR shows `CONFLICTING`. I did the rebase locally to check it's clean, and it is: dropping those three leaves 25 commits and produces exactly one conflict, in the tips registry, where `main` and this branch each appended an entry to `ALL_TIPS` (`share-artifact-link` vs `spaceSearch`) — keep both, and the three locale files resolve the same way.

After that I think this is close to done.

src/mainview/components/SpacesRail.tsx
arditti: Fixed with the cheapest option you named: `Dashboard.tsx` clears `selectedSpaceId` when the viewport crosses below `lg`, via the existing `useNarrowViewport` hook against a `SPACES_RAIL_MIN_WIDTH = 1024` constant exported next to the rail's `hidden lg:flex` classes so the two cannot drift apart silently.

Reproduced your exact walk before and after: Client X selected at 1440px → resize to 1000px → the rail goes away **and** all groups return; resize back to 1440px → the rail is back with *All projects* active, the stale selection is not resurrected. A matchMedia-driven regression test locks it in (select a space → fire the narrow transition → the filtered-out projects reappear).

On the layout-priority observation (21rem panel kept, 14rem rail dropped): left as is for now — the panel is the glanceable "what needs me" surface and the rail's filter now can't get stranded, which was the harmful part.

src/mainview/utils/spaceGroups.ts
arditti: Fixed — `spaceGroups.ts` now imports `projectSearchHaystack` and the local copy is deleted. One matcher, one search language, as the record promises.

decisions/2026/08/17/spaces-dashboard-follows-the-proposal-mock.md
arditti: Corrected — the record now says the header carries the name (no badge) and that the mock's letter badges were deliberately dropped during review, so the record and the screen agree again.

arditti:
All six findings from the second pass are addressed, and the branch is rebased on current `main` (the tips conflict resolved exactly as you suggested — both entries kept).

| | Finding | Status |
|---|---|---|
| 🟡 | Selected space survives the hidden rail | **Fixed** — cleared on the narrow transition via a matchMedia hook against a `SPACES_RAIL_MIN_WIDTH` constant exported beside the rail's `hidden lg:flex` classes; regression test included. Reproduced your 1440→1000px walk before and after |
| 🟡 | Two copies of the search haystack | **Fixed** — `spaceGroups.ts` imports `projectSearchHaystack`, local copy deleted |
| 🟡 | `help.ts` never learned the new tokens | **Fixed** — both topics (`board.filterBar`, `filters.dsl`) now list `space:"Client X"` and `is:home`, in all three locales |
| ⚪ | Two project counts on one screen | **Fixed** — the rail's `All projects` count now includes the pinned Operations board, since selecting it actually shows that board; live both read the same number |
| ⚪ | Stale letter-badge line in the decision record | **Corrected** — the record now says the badges were dropped during review |
| ⚪ | Changelog date | **Moved** to `2026/08/19` |

**One addition beyond the findings**, from a UX pass @arditti asked for before shipping: **each rail row (spaces and the computed `Home`) now carries its needs-you / working split** — the same amber/blue dot pair the group headers use — so a glance at the rail says which space is calling. The dashboard feeds the rail its own cross-project pool, kept live via `rpc:taskUpdated` and gated on the rail existing at all (zero spaces still means zero extra work). Masked spaces blur the split with the name, same rule as the counts you caught last round. Covered by unit tests; the §10 bible row and the changelog entry mention it.

Verification on the rebased branch: `bun run lint` clean; renderer **272 files / 4044 tests** and CLI **47 / 801** green; backend green except `native-pane-owner-forward.test.ts`, which is the same machine-local socket-path-length failure you saw on clean `main`. Browser walk at 1440×900 and 1000px with zero console errors: rail split renders (`Client X ●2 · 2`, `Home ●1 · 7`), `space:"Client X"` filters the panel, the *Needs you* preset composes with it, and the narrow transition clears the rail selection.

h0x91b:
Third pass — short one, because there is almost nothing left to say.

**All three findings from the last round are fixed, verified, and the threads are resolved.** I checked each in the code and then in a running app, not from the replies:

- **The stranded filter is gone.** `Dashboard.tsx` clears `selectedSpaceId` through `useNarrowViewport(SPACES_RAIL_MIN_WIDTH)`, and exporting that constant next to the rail's own `hidden lg:flex` is the right call — it is exactly the drift this would otherwise develop. I re-ran my repro: `Client X` selected at 1440px → 1000px, and the rail goes away **and every group comes back**; widening again restores the rail with *All projects* active and no resurrected selection. That was the one real bug in this PR and it is closed.
- **One haystack.** `spaceGroups.ts` imports `projectSearchHaystack`, the local copy is gone.
- **The record matches the screen again** on the letter badges.

You also picked up both nitpicks without being asked — the changelog moved to `08/19`, and the dashboard heading and the rail now agree on the project count.

The new `Show each rail row's needs-you / working split` commit is a good addition: same amber/blue vocabulary as the group headers, masked rows blur the numbers with the name, and it reads well on screen — a glance at the rail now says which space is calling. No notes.

**Verified on the branch rebased onto current `main`:** `bun run lint` clean, **10,963 tests green** (4147 renderer / 6015 backend / 801 CLI). Zero console errors across the walk. CI is green on your head too.

## The only thing left

**The rebase still has not landed** — the PR is `CONFLICTING` against `main`, and the branch still carries the three #1400 commits that merged as a squash. That is the single thing standing between this and a merge.

I rebased locally again to make sure it is still painless, and it is: `--onto origin/main` past the last of those three leaves 29 commits and **one** conflict, now in `docs/ux/UX_DECISIONS.md`, where `main` and this branch each prepended a newest-first entry — keep both, `main`'s `2026-08-19` above your `2026-08-17`. (The tips registry that clashed last time resolves itself now.)

Push that and I think we merge.

h0x91b:
Claude here again — Arseny asked me to push the rebase for you rather than wait, so I've force-pushed it to `spaces-organizational-v1` (`a570c95c8`). Your commits are untouched; only their base moved. Sorry for reaching into your branch — `maintainerCanModify` was on and `main` was moving fast enough that waiting would have meant a third conflict.

**Heads-up: `git pull` will not work on your side.** Reset instead:

```
git fetch origin && git reset --hard origin/spaces-organizational-v1
```

(replace `origin` with whatever remote points at your fork)

`main` moved twice while this was in flight, so there were three conflicts, all of the append-vs-append kind — both sides added an entry at the end of the same list:

| File | Conflict | Resolution |
|---|---|---|
| `docs/ux/UX_DECISIONS.md` | `main`'s 2026-08-19 menu-flyout entry vs your 2026-08-17 Spaces entry, both prepended | Kept both, newest first |
| `src/mainview/i18n/translations/{en,es,ru}/settings.ts` | #1391's `catalog.*` block vs your `spaces.*` block | Kept both |
| `src/mainview/tips.ts` + its three locale files | #1391's `model-catalog-roles` tip vs your `spaceSearch` tip, both appended to `ALL_TIPS` | Kept both as separate entries |

Nothing of yours was dropped or rewritten. Verified after the rebase: `bun run lint` clean, full suite **11,297 green** (4233 renderer / 6256 backend / 808 CLI).

The PR is `MERGEABLE` now and CI is running. Assuming it stays green this goes in as-is — thanks for the three rounds of turnaround, the fixes were fast and every one of them held up when I checked it.

arditti:
Thanks @h0x91b  for taking it to the finish line and being a true teammate discussing and working on this proposal and PR. Happy to see it shipped as part of dev3

PR #1433, opened 2026-08-20
Keep the last known diff stats visible while the branch status refreshes

h0x91b:
Switching tasks blanked the diff summary (`N files +X −Y`) and the ahead/behind counts into a spinner on every switch, then repainted them a second or two later once the refetch landed.

The numbers are now cached per (task, compare ref) and repainted immediately on switch, with the refresh replacing them in place. The spinner is left for the case it was meant for: a task whose status has never been fetched.

Two places blanked them, both fixed:
- `useTaskBranchStatus` cleared its state on `task.id` change — it now keys the state by task + compare ref and falls back to the cached entry (`branchStatusCache.ts`, LRU-capped). The compare ref is derived during render instead of synced by an effect, so the cache is never read under the previous task's ref.
- `TaskInfoPanel` cleared the lifted `metadataBranchState` on `task.id` change, which wiped what the child had just reported (child effects run before parent effects). The payload now carries its `taskId` and is ignored once it no longer matches the task on screen.

Verified in a browser: 350 ms after a switch the header already shows that task's own numbers, and each task keeps its own values (no cross-task bleed).

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=428be54e-97ea-4d3c-8041-e664aedeb97a) · `dev3://task/428be54e-97ea-4d3c-8041-e664aedeb97a`


h0x91b:
Hey, Claude here — the AI working on this branch. Pushed one more commit-worth of change into the same commit: the branch-status poll no longer holds the merge-completion callback as an effect dependency.

That callback's identity follows the `task`/`project` objects, so every replacement of them restarted the poll and fired another `getBranchStatus`. In a harness where those props are recreated per render it degenerates into a tight refetch loop (~24k git status calls/second); in the app it showed up as an extra fetch per task-list update. The callback is now read through a ref, and `useTaskBranchStatus.cache.test.tsx` guards it — I checked the guard actually bites by reintroducing the old dependency, which hangs the test.

PR #1497, opened 2026-08-23
Auto-expand a space's group when its filter is selected

arditti:
## Summary
- Clicking a space in the rail (or the space-filter sheet on narrow widths) now auto-expands that space's group on the dashboard if it was left collapsed from a previous session — filtering to a space previously just showed the collapsed header with no projects visible.
- Other collapsed spaces are left untouched.

## Test plan
- [x] Added unit tests in `SpaceGroupedProjects.test.tsx` covering the auto-expand and that unrelated collapsed spaces stay collapsed
- [x] `bun run lint` passes
- [x] `bun run test` — mainview (4504/4504) and cli (864/864) green; pre-existing unrelated terminal/isolation failures in the bun config are untouched by this change
- [x] Manually reproduced the bug and verified the fix in a running dev build: collapsed "TFC / Edge general", clicked it in the rail, confirmed it expanded and showed its projects


src/mainview/components/SpaceGroupedProjects.tsx
```diff
@@ -106,12 +110,23 @@ function SpaceGroupedProjects({
...
 
+	useEffect(() => {
+		if (!selectedSpaceId || !collapsed.has(selectedSpaceId)) return;
+		setCollapsed((prev) => {
+			const next = new Set(prev);
+			next.delete(selectedSpaceId);
+			writeCollapsed(next);
+			return next;
+		});
+	}, [selectedSpaceId, collapsed]);
```
h0x91b: This is the one thing I'd like changed before merge.

`collapsed` is both a dependency of this effect and something the effect writes, so it fights the user's own toggle: clicking the header calls `toggleCollapsed`, which adds the space id, the effect reruns because `collapsed` changed, sees the id and deletes it again. The group re-expands in the same commit, so while that space is the selected filter its collapse chevron does nothing at all.

How I checked it: on a scoped QA board I collapsed a space (sticks fine with no filter), selected it in the rail (expands, as intended), then clicked the chevron twice — `aria-expanded` stayed `"true"` and the rows never went away. Same at 900px wide with the filter sheet. It also reproduces without a browser: render the component with `selectedSpaceId="sp_a"`, `fireEvent.click(screen.getByTestId("space-header-sp_a"))`, and `aria-expanded` is still `"true"`.

Suggested shape — derive once per *selection change* and keep the membership check inside the updater, so a later manual collapse sticks:

```ts
useEffect(() => {
	if (!selectedSpaceId) return;
	setCollapsed((prev) => {
		if (!prev.has(selectedSpaceId)) return prev;
		const next = new Set(prev);
		next.delete(selectedSpaceId);
		writeCollapsed(next);
		return next;
	});
}, [selectedSpaceId]);
```

Returning `prev` unchanged keeps React from re-rendering when there is nothing to do, and dropping `collapsed` from the deps is what ends the fight. Re-selecting the space later still expands it, which I think is the behaviour you want.

src/mainview/components/SpaceGroupedProjects.tsx
```diff
@@ -106,12 +110,23 @@ function SpaceGroupedProjects({
...
 	const [collapsed, setCollapsed] = useState<Set<string>>(readCollapsed);
 	const [dragged, setDragged] = useState<{ spaceId: string; projectId: string } | null>(null);
 	const [dropTarget, setDropTarget] = useState<{ spaceId: string; projectId: string; side: "before" | "after" } | null>(null);
 
+	useEffect(() => {
+		if (!selectedSpaceId || !collapsed.has(selectedSpaceId)) return;
+		setCollapsed((prev) => {
+			const next = new Set(prev);
+			next.delete(selectedSpaceId);
+			writeCollapsed(next);
```
h0x91b, line 129: A design question rather than a defect, and worth a maintainer's ruling either way.

Because the expand is written to `localStorage`, the user's saved collapse for that space is not just overridden for the duration of the filter — it is gone. Clear the filter, restart the app, and the group is still open (your own test at `__tests__/SpaceGroupedProjects.test.tsx` asserts `dev3-collapsed-spaces` ends up `[]`, which is exactly this). I confirmed it in the running app too: after clearing the filter and reloading, the group stayed expanded.

That may well be what you intended — "you asked to see this space, so stop hiding it" is a defensible reading, and it keeps the code simple. The alternative is a transient expand: remember the space in a ref, restore the collapsed state when the selection moves elsewhere, and never touch `localStorage`.

I'd leave the choice to the maintainer; if the simple version stays, no change is needed here.

src/mainview/components/SpaceGroupedProjects.tsx
```diff
@@ -106,12 +110,23 @@ function SpaceGroupedProjects({
...
 	spaceOrder,
+	selectedSpaceId,
 }: SpaceGroupedProjectsProps) {
 	const t = useT();
 	const [collapsed, setCollapsed] = useState<Set<string>>(readCollapsed);
 	const [dragged, setDragged] = useState<{ spaceId: string; projectId: string } | null>(null);
 	const [dropTarget, setDropTarget] = useState<{ spaceId: string; projectId: string; side: "before" | "after" } | null>(null);
 
+	useEffect(() => {
+		if (!selectedSpaceId || !collapsed.has(selectedSpaceId)) return;
```
h0x91b: Nitpick, take it or leave it: `selectedSpaceId` can also be `HOME_GROUP_ID` (`"home"`, from `utils/spaceGroups.ts`), and the Home group has no collapse state at all, so that branch can never match. `Dashboard.tsx` already guards it the same way (`if (!selectedSpaceId || selectedSpaceId === HOME_GROUP_ID) return;`), so mirroring it here would make the intent read the same in both places.

src/mainview/components/__tests__/SpaceGroupedProjects.test.tsx
```diff
@@ -83,6 +84,21 @@ describe("SpaceGroupedProjects", () => {
 		expect(JSON.parse(localStorage.getItem("dev3-collapsed-spaces")!)).toEqual(["sp_a"]);
 	});
 
+	it("auto-expands a space that the rail/sheet filter just selected, even if it was left collapsed", () => {
```
h0x91b, line 87: These two cover the new behaviour well. If you add one more while you're here, the case above is the one worth having, because a user hits it within seconds of using the feature:

```tsx
it("still lets the user collapse the selected space by hand", () => {
	renderGroups(new Set(), "sp_a");
	fireEvent.click(screen.getByTestId("space-header-sp_a"));
	expect(screen.getByTestId("space-header-sp_a")).toHaveAttribute("aria-expanded", "false");
	expect(screen.queryByTestId("row-p2")).toBeNull();
});
```

It fails on the current head and passes with the effect change suggested in `SpaceGroupedProjects.tsx`.

Separate, very minor: nothing asserts that `ActivityOverview` actually forwards `selectedSpaceId` down, so deleting that one line would leave the suite green. Not worth a new test on its own — just noting it.

change-logs/2026/08/25/fix-space-filter-auto-expand.md
```diff
@@ -0,0 +1,3 @@
+Short: Space filter auto-expands its group
```
h0x91b, line 1: House convention that is easy to miss as an outside contributor, so purely FYI: `change-logs/YYYY/MM/DD/` is dated by the **expected merge day**, not the day the work started, because the in-app changelog groups entries by ship date. Worth `git mv`-ing the file into the day this is likely to land before merge.

The content and naming are otherwise exactly right — one paragraph, `fix-` prefix, and the `Short:` line included.

h0x91b (changes requested):
Hey Eitav — Claude here, the AI assistant doing code review on this repo. Thank you for this one, and sorry for the wait on your open PRs; your patches are genuinely appreciated.

First: the bug you found is real and your diagnosis is spot on — a collapse left over from a previous session hiding the very group the user just asked to see. The patch is well aimed: one call site, a documented prop in the house comment style, two tests including the "leave the neighbours alone" case, a changelog entry with a `Short:` line, and no unrelated churn. Nice.

I checked out the branch and drove the real UI (desktop rail at 1440x900 and the narrow filter sheet at 900x900) as well as reading the diff. Selecting a space does expand its group exactly as advertised. One thing came out of the manual pass that I think is worth a small change before merge:

**While a space is the selected filter, its collapse chevron stops working.** The new effect keeps `collapsed` in its dependency list and also writes `collapsed`, so a manual collapse is undone by the effect's own rerun. I clicked the chevron twice on both layouts and `aria-expanded` never left `true`. Details and a suggested fix are in the inline comment.

There are also two smaller things (a changelog date, and a question about whether the expand should erase the saved collapse preference for good) plus two optional nitpicks inline. None of them are hard to address, and I'd be happy to re-review as soon as you push.

Thanks again — please do keep them coming.

src/mainview/components/SpaceGroupedProjects.tsx
arditti: Fixed — the effect is now keyed on `selectedSpaceId` alone, with the membership check moved inside the `setCollapsed` updater (returns `prev` unchanged when there's nothing to do). Added the exact regression test you suggested; it fails on the old effect and passes now.

src/mainview/components/SpaceGroupedProjects.tsx
arditti: Good catch, done — the effect now also skips `HOME_GROUP_ID`, matching `Dashboard.tsx`'s guard.

src/mainview/components/__tests__/SpaceGroupedProjects.test.tsx
arditti: Added, verbatim — it does fail on the previous effect and passes with the fix.

change-logs/2026/08/25/fix-space-filter-auto-expand.md
arditti: Moved to `change-logs/2026/08/25/` (today, since that's when this should land).

src/mainview/components/SpaceGroupedProjects.tsx
arditti: Keeping the simple version — persisting the expand to `dev3-collapsed-spaces`. Once the user has asked to see a space via the filter, treating it as "no longer collapsed" for good matches the other collapse-state semantics in this file (manual toggle also persists immediately) and avoids adding a second, transient piece of state to track. Happy to switch to the ref-based transient approach if you'd rather it snap back once the filter clears.

arditti:
Pushed a follow-up commit addressing the review:
- Fixed the effect fighting a manual re-collapse while the space is selected (keyed on `selectedSpaceId` only now, membership check moved into the `setCollapsed` updater).
- Guarded `HOME_GROUP_ID` in the same effect, matching `Dashboard.tsx`.
- Added the suggested regression test for manual collapse-while-selected.
- Moved the changelog entry to `2026/08/25`.
- Replied inline on the persisted-vs-transient expand question — keeping the simple persisted version, open to changing it if you'd prefer otherwise.

Full `bun run test` and `bun run lint` green (aside from pre-existing unrelated flakes in native-terminal/prototype isolation tests on `main`).

h0x91b (approved):
Claude again — re-checked `a02c78ee6` and this is good to go. Thank you for the quick turnaround.

I did not take the fix on trust; I re-ran it both ways.

**Own throwaway tests against your new head** (written independently of yours, then deleted): a manual collapse while the space is the selected filter now sticks and gets written to `localStorage`; it survives a re-render that does not change the selection; selecting away and back expands the group again, which is the right call for a fresh request; a stale collapse is still auto-expanded on first selection; and `HOME_GROUP_ID` touches nothing. Your own three tests pass alongside them — 22/22.

**Real UI again**, scoped QA board, desktop rail at 1440x900 and the narrow filter sheet at 900x900:

1. Collapse with no filter — sticks.
2. Select the space in the rail — expands, as intended.
3. Click the chevron with that space still selected — it collapses and **stays** collapsed, and survives a reload. This was the blocker; it is gone.
4. Clear the filter — the collapse is respected. Re-select the space — it expands again.
5. Same behaviour on the narrow layout with the filter sheet. No console errors.

CI is green on this head, including the aggregate `test` check and all four required contexts (`lint`, `test`, `build-check`, `trivy-scan`).

On the open question — **persisted versus transient expand** — keeping the persisted version is the right call, and here's the reasoning so it is on the record rather than just a preference. The only moment a saved collapse is dropped is the moment it was already contradicting the user's own navigation, and now that a manual re-collapse sticks, taking it back costs exactly one click on the group you are already looking at. The transient version would need to remember the previous selection and restore it on the way out, which reintroduces exactly the class of state fight we just removed: a restore has to know not to overwrite a collapse the user set deliberately while filtered. Simpler state wins here, and collapse is a view convenience rather than user data.

The `HOME_GROUP_ID` guard and the extra regression test are nice additions, and the comment above the effect explains *why* the dependency list is what it is — which is precisely what stops someone re-adding `collapsed` in six months. Squash-merging now. Thanks again, and please do keep them coming.

PR #1503, opened 2026-08-23
Bring-your-own-tunnel remote access: provider setting + forwarded-host origin check

### LINKED ISSUE #1498: Remote access: generalize the tunnel layer with a bring-your-own-tunnel provider
## Summary

Remote access and port exposure are currently hardwired to Cloudflare quick tunnels (`cloudflared`). That is a great zero-config default, but it is not usable everywhere: some corporate/enterprise environments prohibit third-party tunnel egress while providing their own approved, SSO-fronted tunneling service (ngrok-like: a CLI that takes a local port and prints a public HTTPS URL). There are many such services — ngrok itself, Tailscale Funnel, corporate-internal equivalents.

Proposal: generalize the tunnel layer behind a small provider abstraction and add a **"bring your own tunnel"** option, so the user can plug in any CLI-driven tunnel while Cloudflare stays the default.

## POC findings

I ran a POC with one such corporate ngrok-style tunnel CLI in front of dev3's actual `remote-access-server` (stub RPC handler, static code auth). Results:

- **Works through the tunnel as-is:** static assets, HTTP streaming, cookies (the `dev3_session` HttpOnly cookie round-trips fine), `Origin` header forwarding, and WebSockets — both a raw echo WS and dev3's `/rpc` WebSocket carried request/response frames end-to-end.
- **The single breakage:** unlike cloudflared (which preserves the public hostname in `Host`), this class of proxies rewrites `Host` to `localhost:<port>` and carries the public hostname in `X-Forwarded-Host`. dev3's `checkOrigin()` in `src/bun/remote-access-server.ts` compares `Origin` against `Host`, so every browser auth POST and WS upgrade fails with 403 — I reproduced exactly that (`/auth/exchange` → 403 with a *correct* token).
- **Verified fix:** accepting `Origin` host == `X-Forwarded-Host` as well makes the entire flow work end-to-end: `/auth/exchange` 200 + session cookie, authenticated `/health` 200, `/rpc` WS upgrade + RPC round-trip.
- **URL discovery is trivially generalizable:** the CLI prints `Public: https://<name>.<domain>` on stdout and stays in the foreground — same lifecycle shape as `cloudflared` (spawn, scrape URL from output, kill to stop).
- **Bonus properties BYO tunnels can bring:** stable named hostnames (the URL survives a tunnel restart — which would simplify the self-update handoff story, where today we deliberately leak a `cloudflared` process because quick-tunnel hostnames are random, see `releaseMainTunnelForHandoff`), and an SSO layer in front of dev3's own QR/cookie auth.

## Where the Cloudflare coupling lives today

`src/bun/cloudflare-tunnel.ts` is already 90% provider-shaped (a generic manager: spawn, parse URL from output, edge-ready gate, health monitor, restart, handoff). The provider-specific parts are small and enumerable:

1. `parseTunnelUrl()` — regex pinned to `https://*.trycloudflare.com`.
2. `isCloudflaredAvailable()` — `which cloudflared`.
3. `buildTunnelArgv()` — cloudflared CLI flags (`--protocol`, `--edge-bind-address`).
4. `parseTunnelMetricsUrl()` / `waitForEdgeReady()` / health checks — cloudflared's local metrics `/ready` endpoint.
5. `checkOrigin()` in `remote-access-server.ts` — assumes the proxy preserves `Host` (true for cloudflared, false for `X-Forwarded-Host`-style proxies).
6. Naming in UI strings (`remote.anywhereToggle`, `tunnel.exposeViaCloudflare`, install hints), CLI help for `dev3 remote`, and the brew `cloudflared` dependency.

Notably `port-tunnels.ts` (per-task quick/shared port exposure) and the `/p/<subtoken>/<port>/` shared proxy are already provider-agnostic — they only talk to the tunnel manager.

## Proposed design

**1. `TunnelProvider` interface** extracted from `cloudflare-tunnel.ts`:

```ts
interface TunnelProvider {
  id: string;                                  // "cloudflare" | "custom"
  isAvailable(): { ok: boolean; installHint?: string };
  buildArgv(targetPort: number): string[];
  parsePublicUrl(line: string): string | null;
  readiness?: ReadinessProbe;                  // cloudflared: metrics /ready; custom: none/HTTP probe
  stableHostname: boolean;                     // custom named tunnels: true → simpler handoff
}
```

The existing manager (start/stop/health/restart/handoff, `TunnelEntry` map) stays as-is and takes the provider as a parameter. The Cloudflare provider keeps today's behavior verbatim and remains the default.

**2. A `custom` provider configured in app settings** (not per-project — the tunnel is a machine/network property):

- `command` template with a `{port}` placeholder (spawned via `spawn.ts`, foreground process, killed to stop — same lifecycle as cloudflared). Concrete examples the same mechanism must cover with zero code changes:
  - `ngrok http {port} --log stdout --log-format json` (URL appears as `url=https://….ngrok-free.app` in the log stream)
  - `tailscale funnel {port}` (prints `https://<machine>.<tailnet>.ts.net`)
  - any corporate-internal equivalent that prints `https://…` and stays in the foreground
- `urlPattern` — optional regex with the default of "first `https://…` URL printed to stdout/stderr". ngrok, Tailscale Funnel, cloudflared, and the corporate CLI from the POC are all matched by that default.
- Readiness: without a provider-specific `/ready`, gate on URL + a short settle delay, optionally an HTTP `HEAD` probe of the public URL. Health: process liveness (`exited` hook already handles this) instead of the metrics poll.

**3. Reverse-proxy-aware origin check.** Providers split into two Host behaviors: Host-preserving (cloudflared, ngrok by default) already work; Host-rewriting proxies (those that send `Host: localhost:<port>` + `X-Forwarded-Host`) need `checkOrigin()` extended to also accept `Origin` host == `X-Forwarded-Host` — ideally only while a custom tunnel is configured, so the default posture is unchanged. (Spoofing risk is limited — a browser cannot attach `X-Forwarded-Host` cross-origin without a CORS preflight, and non-browser clients already skip the origin check by not sending `Origin` — but gating it keeps the reasoning simple.) `clientIp` already reads `x-forwarded-for`, no change needed there.

**4. Surface it generically.** Remote Access dialog gets a tunnel picker (default "Cloudflare Tunnel", "Custom command…"); Cloudflare-specific strings ("Expose via Cloudflare", QR hints) become "public tunnel" phrasing except where the provider is genuinely Cloudflare; `dev3 remote` gets `--tunnel-cmd`/settings-driven equivalent alongside `--no-tunnel`. Missing custom binary degrades exactly like missing `cloudflared` does today (feature-gate + install hint), per the update-channels rule.

## Open questions

- Self-update handoff for custom providers: with `stableHostname: true` the successor can simply restart the tunnel (same URL, cookie survives) instead of adopting the pid — arguably simpler than today's leak-and-adopt. With random-hostname custom tunnels, kill-and-restart with a changed URL seems acceptable for a v1.
- Per-port "Expose" buttons with a custom provider mint one process per port (like quick tunnels today); providers with named tunnels may need a name template (`{port}` in a `--name`) to avoid collisions.
- Whether `--expose-ports` in `dev3 remote install-service` should support the custom provider in v1 or start with the main UI tunnel only.

Happy to contribute the implementation — the POC branch already validates the auth-path change and the URL-scrape lifecycle.

---

## Update (2026-08-24): scope expanded to every tunnel dev3 starts

The PoC artifact in #1503 now generalizes **per-task port tunnels too** — the main remote-access tunnel, per-port quick tunnels, and the shared multi-port tunnel all run through the one configured provider. Findings that answer the open questions above:

- **Per-task tunnels share the main tunnel's setting.** The tunnel is a machine/network property: a network that forbids `cloudflared` for the web UI forbids it for an exposed dev server just the same, and two settings could silently disagree. `port-tunnels.ts` needed no changes — the generalization lives entirely in the tunnel manager's provider resolution.
- **Per-port Expose with a custom provider mints one process per port**, exactly like quick tunnels today. A fixed-hostname command (no `{port}` in a `--name`) cannot serve two concurrent tunnels — the second lands on the same public URL — so dev3 logs a warning naming the colliding tunnels and the settings hint spells the constraint out. Name templates remain the provider-side fix; dev3 does not try to arbitrate the collision.
- **`--expose-ports` in `dev3 remote` supports the custom provider in v1** for free, since it goes through the same manager path.
- **Hostname-stability learning stays main-only**: the learned `(command, url)` memory would be flip-flopped by concurrent port tunnels, and the stability bit only feeds the main tunnel's self-update handoff.

arditti:
Started as a live artifact for the discussion in #1498; after two review rounds (all findings addressed and re-verified by running the code) it is now **ready for review as a real change**.

Closes #1498.

## What this does

Makes **every tunnel dev3 starts** work behind any tunnel tool, not only the built-in Cloudflare quick tunnel: the main remote-access tunnel, per-task exposed-port quick tunnels, and the shared multi-port tunnel all run through one configured provider. Cloudflare stays the default and nothing changes for users who never touch the new setting.

Three pieces:

1. **A tunnel provider setting.** Settings → System → *Tunnel provider*: either the built-in quick tunnel (default) or a **custom command** the user supplies, with `{port}` substituted at launch (`ngrok http {port} --log stdout`, or any equivalent tool). An optional *Public URL pattern* regex overrides how the public URL is scraped from the command's output; the default matches the first `https://` URL printed on stdout or stderr. One setting governs all tunnel kinds — the tunnel is a machine/network property, and a network that forbids `cloudflared` for the UI forbids it for an exposed dev server just the same. Selecting Custom with a blank command **fails closed**: nothing starts, `cloudflared` is never silently spawned, and both the settings field and the Remote Access modal say the command is empty. A command whose binary is missing fails at run time (`sh` exit 127) and the modal names that too — availability is derived from the line that actually runs, not from a string probe.
2. **An origin-check fix.** `checkOrigin` in the remote-access server compared the request `Origin` host against the `Host` header only. Cloudflare's quick tunnel preserves the public `Host`, so this worked; most other tunnels terminate TLS at their own edge and forward `Host: localhost:<port>`, carrying the public hostname in `X-Forwarded-Host` instead. The result was a hard 403 on `/auth/exchange` even with a valid QR token. The check now also accepts the first entry of `X-Forwarded-Host` — but **only while a custom provider is configured** (as #1498 §3 proposed), since the header is client-controlled on a direct connection; the default posture stays `Origin === Host`, and foreign origins are still rejected either way.
3. **Per-task port tunnels on the same provider.** `port-tunnels.ts` was already provider-agnostic — it only talks to the tunnel manager — so the generalization is the manager resolving the provider for every tunnel kind instead of pinning non-main kinds to cloudflared. Each exposed port spawns its own instance of the custom command, and the existing port-scan liveness reaper stops stale processes exactly as it did for cloudflared.

That origin check was the *only* incompatibility found. Auth, the session cookie, the RPC WebSocket and the PTY stream all work unchanged behind a Host-rewriting tunnel once it passes.

## Findings from the proof of concept

Validated end to end against a third-party tunnel that rewrites `Host`:

- Before the fix: `/auth/exchange` returns **403** with a correct token — the QR link is unusable.
- After the one-line change: **200**, `dev3_session` cookie set, RPC WebSocket round-trip succeeds, terminal streams.
- Edge readiness is reported for a custom provider too. Cloudflare's local metrics `/ready` endpoint is provider-specific, so a custom tunnel is instead verified by an HTTP `HEAD` against the public URL it printed: any HTTP status counts as routable (an auth-gated tunnel answering 401/403 still proves the hostname resolves and forwards), and only a transport failure — DNS, refused, TLS, timeout — means the edge is not up yet. The URL is still published best-effort if the probe never answers, logged as `edgeCheck: "unconfirmed"` rather than silently claimed ready. The same probe becomes the ongoing health check, so a custom tunnel whose edge dies while its process lives is caught and restarted after three consecutive failures. The Cloudflare-only "DNS is still propagating" hint is suppressed for custom providers.
- An invalid user regex falls back to the generic URL pattern with a warning rather than failing the tunnel.

### Per-task tunnels — design choices worth reviewing

- **Hostname-stability learning stays main-only.** The learned-stability memory holds a single `(command, url)` pair; concurrent port tunnels producing different URLs would flip-flop the record and destroy the main tunnel's proof. The stability bit exists only to feed the main tunnel's self-update handoff, which port tunnels never take part in, so `startEntry` records the observation only for `kind: "main"`.
- **A fixed-hostname command cannot serve concurrent tunnels** — the second invocation lands on the same public URL and the provider's edge routes it to whichever process won. dev3 cannot arbitrate that; it logs a clear warning naming both tunnel ids when two live tunnels share one URL, and the settings hint says a fixed-hostname command serves one tunnel at a time. Providers with name templates (`{port}` in a `--name`) avoid it entirely.
- The custom public-URL readiness probe and health monitor apply to port tunnels as-is; a missing custom binary fails the expose with the same "Failed" state a missing `cloudflared` produced.

## Manual QA

Validated end-to-end against a **real homegrown, ngrok-like corporate tunnel CLI** (the class of tool the issue was written for; described by shape rather than name): it terminates TLS at its own edge, rewrites `Host` to `localhost:<port>` with the public hostname in `X-Forwarded-Host`, and fronts everything with an SSO layer keeping its own HttpOnly session cookie on the same public hostname. Measured through it against the real remote-access server: `/auth/exchange` 403 before the origin-check fix and 200 after; `Set-Cookie: dev3_session` passing through the SSO edge intact; a real browser request carrying **both** HttpOnly cookies (the SSO layer's and `dev3_session`) in one `Cookie` header with no collision; the `/rpc` WebSocket upgrading through the SSO edge; 401 without the cookie and 403 for a foreign origin still enforced through the tunnel. The auth-gated SSO edge is also why the readiness probe counts any HTTP status as routable — observed, not assumed.

Also driven in a browser against a scoped throwaway board, with a fake tunnel script standing in for a real tool:

- Provider select renders, switches, and persists to `settings.json`. The section now reads *Tunnel provider*, its description covers remote access and exposed ports alike, and the command hint carries the fixed-hostname caveat — verified rendered, zero console errors.
- Custom provider connected: backend spawned the command, scraped the URL from its output, the toggle reads "Accessible from anywhere (custom tunnel)", the access URL uses the public hostname, and **the QR image regenerates for that URL**. No Cloudflare propagation hint. Stopping the tunnel kills the process.
- Custom command pointing at a missing binary: the modal shows "The custom tunnel command was not found" with a *Tunnel settings* link (navigates to Settings → System) and a *Recheck* button.
- The per-task path (custom argv per port, port substitution, no host-memory writes, duplicate-URL warning) is covered by unit tests on the tunnel manager; the Expose button copy is now provider-neutral ("Expose via public tunnel").

## Resolved during the discussion

- *Should a custom provider get a health probe of its own instead of relying on process liveness?* — Yes, and it now does; see the readiness/health bullet above. Verified live: `[tunnel] Tunnel connected {"probe":"public-url","edgeCheck":"ready"}` against a real third-party tunnel CLI.

- *A custom provider with a fixed hostname — can the self-update handoff stop respawning blind?* — Yes. The handoff leaks a live `cloudflared` process to its successor only because a quick tunnel's hostname is random per process; a provider invoked with a fixed name returns the same hostname every time, so the leak buys nothing and risks an orphan. Stability is now **learned by observation**, never declared: the last `(command, url)` pair is remembered, and a command that yields the same URL again is marked stable (logged as `hostname: "stable" | "unproven"`). For a stable provider the handoff **stops** the tunnel and records the URL the successor must reproduce; the successor spawns its own tunnel, compares, and says whether the sessions survived — with a "re-scan the QR" warning if the hostname actually moved. An abandoned restart restarts the tunnel instead of re-adopting a process that no longer exists, and an adopted custom tunnel now gets the public-URL health probe it previously lacked. Verified live: two consecutive starts of a real third-party tunnel CLI on different local ports produced one public hostname, logged `hostname: "unproven"` then `hostname: "stable"`.

- *Should per-task port tunnels follow in the same setting, or get their own?* — The same setting, and this PR now implements it (see "Per-task tunnels" above). The tunnel is a property of the machine and its network, not of one tunnel kind; a network that requires the corporate tunnel for the web UI requires it for an exposed port too, and two settings would let them silently disagree.

## Open questions for the discussion

- Should the URL pattern be exposed at all, or is "first https URL in the output" enough for every realistic tool?
- Is two identical hostnames enough evidence of stability, or should it take three?
- Is a logged warning enough for the fixed-hostname-command collision on concurrent tunnels, or should the second expose be refused outright?


arditti:
Validated end to end against a real third-party tunnel CLI (one that terminates TLS at its own edge and forwards `Host: localhost:<port>` — exactly the case the origin fix is for), driven from a phone:

```
[tunnel] Starting custom tunnel command {"id":"main","targetPort":15541}
[tunnel] tunnel {"line":"Public: https://<public-host>"}
[tunnel] Tunnel connected {"edgeCheck":"skipped"}
[tunnel] tunnel {"line":"→ WS /rpc"}
[remote-access] RPC WS upgrade: authorized {"ua":"Mozilla/5.0 (iPhone; ...)"}
[tunnel] tunnel {"line":"✓ WS connected to local target"}
```

Full lifecycle confirmed on the running app: resolve the provider and substitute `{port}` → spawn under the platform shell → scrape the public URL from the command's output → publish it (state `connected`, QR and access URL regenerated for the public host) → serve a real request (`GET /?token=… → 200`) → **phone scanned the QR, RPC WebSocket upgrade authorized** → stop kills the child and returns to `idle`. The missing-binary branch was exercised separately with a bogus command.

Also pushed a follow-up commit for something the run exposed: the tunnel logger was named `cf-tunnel` and labelled every piped line `cloudflared`, so a bring-your-own-tunnel session read as if cloudflared were running, and it reported `edgeReady: true` for a readiness probe it deliberately skips. Lines are now attributed to the command's own binary name, and the edge check reports `skipped` rather than a fake confirmed edge.

Two more lifecycle gaps for the discussion list, beyond the ones in the description:
- A custom tunnel that dies resets state to `idle` but is not respawned — no restart or backoff.
- The self-update handoff (the deliberately-leaked tunnel process a successor adopts) was designed around cloudflared's semantics; URL stability across a restart depends entirely on the third-party tool, so that path needs its own decision.

arditti:
Follow-up pushed: edge readiness is no longer skipped for a custom provider.

- The launch path waits for the tunnel to be routable by sending an HTTP `HEAD` to the public URL the CLI printed (10s budget, 750ms poll). **Any** HTTP status counts as up — an auth-gated tunnel answering 401/403 still proves the hostname resolves and forwards — while a transport failure (DNS, connection refused, TLS, timeout) means the edge is not there yet.
- Logging is honest about which probe ran: `probe: "metrics-ready"` for the built-in provider, `probe: "public-url"` for a custom one, and `edgeCheck: "ready" | "unconfirmed"`. An unconfirmed edge still publishes the URL best-effort with a warning, rather than blocking a tunnel that simply refuses HEAD.
- The same probe replaces process-liveness as the ongoing health check for custom providers, so a tunnel whose edge dies while the process stays alive is now detected and restarted after three consecutive failures.
- Custom tunnel output lines are attributed to the tool that produced them instead of being labelled as the built-in binary.

Verified live against a real third-party tunnel CLI: `[tunnel] Tunnel connected {"probe":"public-url","edgeCheck":"ready"}`, followed by a successful request through the public hostname. Unit tests cover the ready path, a 403 counting as routable, the unconfirmed best-effort path, and the ongoing health probe.

arditti:
Pushed `838d6cf` — the last remaining tunnel-lifecycle gap for a custom provider: the self-update handoff.

**Why it mattered.** The handoff deliberately leaks a live tunnel process to the successor server, and the only reason is that a Cloudflare quick tunnel hands out a random hostname per process — killing it would change the public URL and invalidate the host-bound `dev3_session` cookie on the phone the user is holding. A custom provider invoked with a fixed name has no such problem, so for those the leak buys nothing and adds an orphan risk.

**How it decides.** Stability is observed, not declared. `src/bun/tunnel-host-memory.ts` remembers the last `(command, url)` pair in `~/.dev3.0/remote/tunnel-hosts.json` (additive file, nothing renamed or migrated); the same command yielding the same URL again marks it stable, and any different URL revokes the mark. A user-facing "my hostname is fixed" checkbox was rejected: a wrong declaration makes dev3 publish a URL that does not route, which is exactly what the readiness probe exists to prevent — a wrong observation can only fall back to the conservative leak. Parsing the command for name/domain flags was rejected as provider-specific guessing.

**What changes at runtime.** For a stable provider, `prepareHandoff` stops the tunnel and records `stableTunnelUrl` instead of a pid; the successor spawns its own tunnel, compares the URL, and logs whether sessions survived, printing a "re-scan the QR" warning when the hostname actually moved. An abandoned restart restarts the tunnel rather than re-adopting a dead pid. Separately, an adopted custom tunnel now gets the public-URL health probe — it previously had neither a metrics endpoint nor an exit promise, so it was trusted forever.

Live-validated against a real third-party tunnel CLI: two consecutive starts on different local ports both resolved to the one public hostname, logging `hostname: "unproven"` then `hostname: "stable"`, `edgeCheck: "ready"` both times. Reasoning recorded in `decisions/2026/08/23/learned-stable-hostname-for-custom-tunnels.md`; 12 new unit tests, all three suites green.

src/bun/tunnel-provider.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export interface ResolvedTunnelProvider {
+	kind: "cloudflare" | "custom";
+	/** Shell command with `{port}` already NOT substituted — see buildCustomTunnelArgv. */
+	command: string | null;
+	urlRegex: RegExp | null;
+}
+
+/** Default URL scrape for custom providers: the first https URL the CLI prints. */
+export const GENERIC_TUNNEL_URL_REGEX = /https:\/\/[a-zA-Z0-9][a-zA-Z0-9._-]*(?::\d+)?(?=[\s"',;)\]]|$)/;
```
h0x91b: **Blocker** — this never matches a URL that is followed by `/`.

The lookahead requires whitespace, a quote, a comma, a semicolon, a bracket or end-of-line right after the hostname. `/` is none of those, and the character class before it can't consume `/` either, so the whole match fails rather than stopping at the hostname.

Measured against the real regex:

| Input line | Scraped |
|---|---|
| `https://machine.tailnet.ts.net/` | `null` |
| `Public: https://x.corp.example.com/app` | `null` |
| `Visit https://x.example.com.` | `https://x.example.com.` (trailing full stop swallowed into the host) |

Tailscale Funnel prints exactly the first shape:

```
Available on the internet:

https://machine.tailnet.ts.net/
|-- proxy http://127.0.0.1:3000
```

I ran that through a fake tunnel script and got `startTunnel -> null (state=idle) in 30005ms` — the user waits the full `URL_WAIT_TIMEOUT_MS` and then gets a generic failure, with the URL visible in the log the whole time.

Suggested shape: allow an optional path, then strip a single trailing `.` or `,` from the match:

```js
/https:\/\/[a-zA-Z0-9][a-zA-Z0-9._-]*(?::\d+)?(?:\/[^\s"',;)\]]*)?/
```

Worth adding both failing shapes to the `GENERIC_TUNNEL_URL_REGEX` test too — every case currently in `tunnel-provider.test.ts` happens to end on whitespace or end-of-line, which is why this slipped through.

src/bun/tunnel-provider.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * A custom command is a free-form shell line (templates like
+ * `ngrok http {port} --log stdout` need shell word splitting anyway), so it
+ * runs under the platform shell — the same trust model as project scripts.
+ */
+export function buildCustomTunnelArgv(command: string, targetPort: number): string[] {
+	const line = command.replaceAll("{port}", String(targetPort));
+	return process.platform === "win32" ? ["cmd", "/c", line] : ["sh", "-c", line];
```
h0x91b: `sh -c` execs into the child for a plain command, so the common case is fine — but it doesn't exec for a pipeline, and then `entry.process.kill()` in `stopEntry` signals `sh` while the tunnel keeps running.

Measured, same script, started → connected → stopped:

```
plain command      → stray tunnel procs after stop: 0
command | cat      → stray tunnel procs after stop: 1
```

An orphan here is a public URL still pointed at a port dev3 no longer owns.

`sh -c "exec <line>"` fixes the simple case; spawning in its own process group and signalling the group covers pipelines too. Worth noting either way because the same pid travels into the self-update handoff — `releaseMainTunnelForHandoff` records the `sh` pid, so the successor can end up adopting a wrapper rather than the tunnel.

src/bun/tunnel-provider.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/** The tool a custom command runs, used to attribute its output lines in logs. */
+export function tunnelCommandName(command: string): string {
+	const binary = command.trim().split(/\s+/)[0] ?? "";
+	return binary.split(/[\\/]/).pop() || "custom-tunnel";
+}
+
+/** Best-effort availability: the command's first word resolves on PATH. */
+export function isCustomTunnelBinaryAvailable(command: string): boolean {
+	const binary = command.trim().split(/\s+/)[0];
```
h0x91b: **Blocker** — splitting the command on whitespace gives a false "not found" for two shapes people will hit immediately.

Measured against the real function:

| Command | Returns | Reality |
|---|---|---|
| `"/opt/my tools/tunnel.sh" {port}` | `false` | Runs fine under `sh -c` |
| `NGROK_AUTHTOKEN=abc ngrok http {port}` | `false` | Runs fine, and it's the form in ngrok's own docs |
| `/opt/tools/tunnel.sh {port}` | `true` | Correct |

This value gates everything downstream: `getRemoteAccessQR` refuses to auto-start the tunnel, and the modal shows "The custom tunnel command was not found — its binary is not on PATH", pointing the user at a problem that doesn't exist.

Two ways out, and I'd lean on the second:

1. Strip a leading run of `VAR=value` assignments and honour shell quoting before probing.
2. Drop the probe for custom providers entirely and derive availability from the outcome. A `sh -c` that exits `127` immediately *is* "command not found" — I confirmed the current code already sees exactly that: `exitCode: 127`, `sh: totally-not-installed-tunnel: command not found`, back in 12 ms. Less code, and it can't go stale relative to what actually runs.

src/bun/remote-access-server.ts
```diff
@@ -77,9 +77,14 @@ export function checkOrigin(req: Request): boolean {
...
-	if (!host) return false;
+	// A Host-rewriting tunnel proxy (ngrok-style BYO tunnels) sends
+	// `Host: localhost:<port>` and carries the public hostname in
+	// X-Forwarded-Host — compare against that too (first entry if proxies chain).
+	const forwardedHost = req.headers.get("x-forwarded-host")?.split(",")[0]?.trim() || null;
+	if (!host && !forwardedHost) return false;
 	try {
-		return new URL(origin).host === host;
+		const originHost = new URL(origin).host;
+		return originHost === host || originHost === forwardedHost;
```
h0x91b, line 94: This is the one finding with a security shape, so two axes rather than one severity — they point in opposite directions.

**How bad the damage would be: high.** `checkOrigin` guards `/auth/exchange` and the `/rpc` WebSocket upgrade. Past it is a session cookie, and a dev3 session is a terminal on the machine.

**How easy the attack is: low.** `X-Forwarded-Host` isn't CORS-safelisted, so a browser has to win a preflight first — and there's no `OPTIONS` handler and no `Access-Control-*` header anywhere in this file, so it never gets permission. A non-browser client bypasses the check anyway by omitting `Origin`, and gains nothing new.

What I measured calling the exported function directly:

| Headers | Result |
|---|---|
| `Host: abc.trycloudflare.com` + matching `Origin` | `true` — correct |
| `Host: localhost:4242`, `X-Forwarded-Host: me.corp.example.com`, matching `Origin` | `true` — the intended fix, correct |
| `Host: abc.trycloudflare.com`, `X-Forwarded-Host: evil.com`, `Origin: https://evil.com` | **`true`** |
| `Host: 192.168.1.5:4242`, `X-Forwarded-Host: evil.com`, `Origin: https://evil.com` | **`true`** |
| `Host: abc.trycloudflare.com`, `Origin: https://evil.com`, no forwarded header | `false` — correct |

The last two rows are the point: when the client controls both headers, the check compares a value against itself. And it applies to every install, not only ones running a custom tunnel — Cloudflare default and plain LAN mode included.

#1498 §3 proposed gating it (*"ideally only while a custom tunnel is configured, so the default posture is unchanged"*), and the PR drops that without saying why. I'd keep it — `resolveRemoteTunnelProvider()` is already imported next door, so it's cheap; just cache it, since `checkOrigin` runs per request and `loadSettingsSync` reads a file.

Also worth a test: `remote-access-auth.test.ts` proves a foreign origin is rejected *when the forwarded host is honest*, but nothing asserts what happens when the attacker sets both. That's the case that would have caught this.

To be explicit about the limits of what I checked: I did not fire a real cross-origin request at a running server. The "browser can't get through" half is inferred from the source, not measured.

src/bun/settings.ts
```diff
@@ -57,6 +57,20 @@ function sanitizeShortcutOverrides(raw: unknown): GlobalSettings["keyboardShortc
 	return Object.keys(out).length > 0 ? out : undefined;
 }
 
+/**
+ * `provider: "custom"` is meaningful only with a non-blank command; anything
+ * else collapses to "no custom tunnel" (undefined ⇒ built-in Cloudflare).
+ */
+function sanitizeRemoteTunnel(raw: unknown): GlobalSettings["remoteTunnel"] {
```
h0x91b, line 66: An empty command doesn't fail closed here — it falls back to Cloudflare, and the UI and the backend disagree in the meantime.

What I did in the browser: picked "Custom command", typed nothing, reloaded.

| Step | State |
|---|---|
| After selecting Custom | `settings.json` holds `{"provider":"custom","command":""}`; the select reads `custom` |
| Backend, same moment | `sanitizeRemoteTunnel` drops it on load, so `resolveRemoteTunnelProvider()` returns `cloudflare`. Confirmed live — with an empty command my harness started a real `cloudflared` quick tunnel |
| After reload | The select reads `cloudflare`. No message, no explanation |

The fallback direction is the awkward part: someone on a network that forbids `cloudflared` egress — the user this feature exists for — selects Custom, fumbles the command, and dev3 quietly spawns `cloudflared` anyway.

Either keep `{provider:"custom", command:""}` through the sanitizer and have `resolveRemoteTunnelProvider` return a third `misconfigured` kind that starts nothing and surfaces "the custom tunnel command is empty" in the modal, or don't let the select move to Custom until a non-blank command exists. Fail closed and say so.

Smaller, related: both inputs are uncontrolled (`defaultValue` + `onBlur`), so the URL-pattern `onBlur` spreads `globalSettings.remoteTunnel!`. It works today because the local state updates immediately, but it's one refactor away from wiping the command field.

src/bun/cloudflare-tunnel.ts
```diff
@@ -221,11 +261,11 @@ function monitorStderr(entry: TunnelEntry, stderr: ReadableStream): Promise<stri
 	});
 
 	function processLine(line: string): void {
-		logCloudflaredLine(entry.id, line);
+		logTunnelLine(entry.id, line, source);
 		const metricsReadyUrl = parseTunnelMetricsUrl(line);
 		if (metricsReadyUrl) entry.metricsReadyUrl = metricsReadyUrl;
 		if (!urlResolved) {
-			const url = parseTunnelUrl(line);
+			const url = line.match(urlRegex)?.[0] ?? null;
```
h0x91b: Taking `[0]` means the whole match becomes the URL, so a capture group in the user's pattern is ignored.

The natural pattern for ngrok's log line is `url=(https://\S+)` — with this code that publishes `url=https://…` as the public URL. The settings hint calls it "a regular expression matching the public URL", which reads as though a group would work.

```js
const m = line.match(urlRegex);
const url = m ? (m[1] ?? m[0]) : null;
```

Prefer group 1 when the pattern has one. One line, and the field then behaves the way anyone would expect.

Adjacent enough to fold in rather than open a second thread: `compileUrlPattern` builds a `RegExp` from arbitrary user text and runs it against every output line, so a pathological pattern can pin a core. Self-inflicted and low-stakes, but a length cap on the field costs nothing.

src/bun/cloudflare-tunnel.ts
```diff
@@ -221,11 +261,11 @@ function monitorStderr(entry: TunnelEntry, stderr: ReadableStream): Promise<stri
 	});
 
 	function processLine(line: string): void {
-		logCloudflaredLine(entry.id, line);
+		logTunnelLine(entry.id, line, source);
 		const metricsReadyUrl = parseTunnelMetricsUrl(line);
```
h0x91b: Small one: `parseTunnelMetricsUrl` still runs on a **custom** provider's output.

A CLI that happens to print `Starting metrics server on …` would set `entry.metricsReadyUrl`, and `checkHealth` prefers that branch over `publicProbeUrl` — silently disabling the public-URL health probe this PR just added, in favour of polling a `/ready` endpoint that isn't cloudflared's.

Guarding it with `if (!isCustom)` closes it.

change-logs/2026/08/25/feature-byo-tunnel-origin-check.md
```diff
@@ -0,0 +1,3 @@
+Short: Remote access behind any tunnel
```
h0x91b, line 1: Two small changelog-policy things:

- The path date should be the **expected merge date**, not the day the entry was written — the changelog UI groups by ship day. Worth renaming the directory right before merge.
- The policy is 1–3 sentences, one paragraph; this is about 120 words in two long sentences. The first sentence plus the origin-check sentence would carry it.

The `Short:` line is correct, and ranking it unnumbered (mid-pack) looks right to me for the popover.

h0x91b (changes requested):
Hey — Claude here, the AI assistant working on this repo with Arseny. I went through #1503 in detail, and rather than just reading the diff I checked the branch out and drove it: ran the tunnel manager against fake tunnel scripts, exercised the failure paths, and put the settings screen and the Remote Access modal through a headless browser.

First, thank you. This is a genuinely well-built contribution — the decision record is clearer than a lot of what lands here, the reasoning about learned-vs-declared hostname stability is right, and the fact that `port-tunnels.ts` needed nothing but a comment change says the seam was put in the correct place. i18n is complete in all three locales and reads like someone wrote it. I verified the stability learning works exactly as documented: two consecutive starts of the same command logged `hostname: "unproven"` then `hostname: "stable"`, and the memory file flipped accordingly.

I'm marking this **request changes** rather than comment, mostly because two things I found stop the feature working for command shapes people will absolutely reach for first. Both are small and local. Everything is inline on the relevant lines.

## Summary

**Two that block it for now**

1. `GENERIC_TUNNEL_URL_REGEX` never matches a URL followed by `/`. Tailscale Funnel — one of the three providers named in #1498 — prints `https://machine.tailnet.ts.net/`, which scrapes to `null`. I ran it: 30 seconds of silence, then a generic failure, with the URL sitting right there in the log.
2. The PATH probe splits the command on whitespace, so a quoted path containing a space and ngrok's own documented `NGROK_AUTHTOKEN=x ngrok http {port}` both report "command not found" for commands that run fine.

**Four worth fixing**

3. The origin check accepts `X-Forwarded-Host` unconditionally, not only under a custom provider. Your own issue proposed gating it; this is the one finding with a security shape, so I've split it two ways inline.
4. Selecting Custom with an empty command silently falls back to Cloudflare, and the UI disagrees with the backend until a reload.
5. `sh -c` orphans the tunnel process when the command is a pipeline (measured: one stray process after stop).
6. A capture group in a user-supplied `urlPattern` is ignored.

**Nitpicks** — three inline, plus three that aren't worth a thread of their own:

- `checkTunnelBinary` has no caller anywhere in the renderer. Not your doing (`checkCloudflared` had none either), but since you're renaming it, deleting it may be tidier than carrying it forward.
- No renderer test for the new Tunnel provider control in `SystemSettingsSection` — the select, the conditional fields, and the persist callback are uncovered. `App.test.tsx` covers the modal side, so this is the one gap.
- `cloudflare-tunnel.ts` is now a provider-agnostic manager under a Cloudflare-shaped filename, and `TUNNEL_EDGE_READY` sits next to `CUSTOM_TUNNEL_EDGE_READY` without either naming its provider. I'd leave the rename to a follow-up — moving a 900-line file would bury this change.

## What I did not check

Saying it plainly so silence doesn't read as verified: I never ran a real third-party tunnel (only fake scripts), never touched Windows (`cmd /c`, `where`), never executed a full self-update handoff cycle, never drove the per-task Expose button through a custom provider, and never fired a real cross-origin request at a running server — the browser-preflight reasoning in the origin thread is inferred from the absence of CORS headers in the source, not measured.

`bun run lint` is clean and every suite you touched passes in isolation (129 backend + 363 renderer).

Thanks again for this — and for the write-up in #1498, which made the review far easier than it would otherwise have been. Happy to re-review whenever you've had a look.

arditti:
Thanks for this review — checking the branch out and measuring each claim instead of reading the diff is exactly the review this artifact needed, and the findings table speaks for itself. Everything below landed in `89634a64`, one round as promised. First a note on what this branch *has* been run against, since your "what I did not check" list calls it out.

**What was tested against a real tunnel.** The whole flow was validated end-to-end against a homegrown, ngrok-like corporate tunnel CLI — the class of tool the issue was written for, so I'll describe its shape rather than name it: takes a local port, prints a public `https://` URL, stays in the foreground, terminates TLS at its own edge, rewrites `Host` to `localhost:<port>` with the public hostname in `X-Forwarded-Host`, and fronts everything with an SSO layer that keeps its own HttpOnly session cookie on the same public hostname. Measured against the real dev3 remote-access server through that tunnel: `/auth/exchange` 403 before the origin-check fix and 200 after; `Set-Cookie: dev3_session` passing through the SSO edge intact; a real browser request carrying **both** HttpOnly cookies — the SSO layer's and `dev3_session` — in one `Cookie` header on one hostname with no collision; the `/rpc` WebSocket upgrading through the SSO edge; 401 without the cookie and 403 for a foreign origin, both still enforced through the tunnel. The SSO front is also why the readiness probe counts any HTTP status as routable — an auth-gated edge answers 401/403 to an anonymous `HEAD` while being perfectly alive; observed, not assumed. The learned-stability path ran against the same CLI: two starts on different local ports, one hostname, `"unproven"` then `"stable"`. Still *not* run against the real tool: the per-task Expose path and a full self-update handoff cycle — those are fake-script and unit-test coverage only, so our untested lists mostly overlap rather than cancel out.

Per thread:

**1. URL regex (blocker)** — Agreed, and embarrassing given Tailscale Funnel is named in the issue. Fixed with an optional path segment plus stripping a single trailing `.`/`,`; your three failing shapes (including the Funnel trailing-slash line) are in the regex tests now.

**2. PATH probe (blocker)** — Agreed, and I took your second option: the pre-flight probe stays for `cloudflared` only, and a custom command's availability is derived from what actually happens — a `sh -c` exit 127/126 before a URL appears is recorded as `command-not-found` and the Remote Access modal renders it as "the command was not found" with a *Tunnel settings* link, instead of a generic failure. The string probe is deleted; as you say, the outcome can't go stale relative to what runs.

**3. Origin check** — You're right, and dropping the gate the issue itself proposed deserved at least a sentence in the PR; it was an oversight, not a decision. Restored: `X-Forwarded-Host` is honored only while a custom provider is configured, with the resolved provider cached (5 s TTL) rather than re-read per request, and the attacker-sets-both-headers case is now an explicit test — both your table rows. Within a custom-provider install the residual trust in the forwarded header is inherent to Host-rewriting proxies; that stays documented rather than "fixed".

**4. Empty command fallback** — Agreed on fail-closed, and your framing of the fallback direction was the convincing part. `provider: "custom"` with a blank command now survives the sanitizer as a `misconfigured` provider kind: nothing starts, `cloudflared` is never spawned, the settings field shows an inline "required — no tunnel starts while this is empty" warning, and the modal says the command is empty with a settings link. The two inputs are controlled now and persist both values together on blur, so the stale-spread hazard is gone too.

**5. `sh -c` orphans** — Agreed on the failure, with one scope note: full process-group semantics aren't portable (macOS ships no `setsid`, Windows is `cmd /c`), so a plain command line is now `exec`-prefixed — which also makes the pid recorded for the self-update handoff the real tunnel rather than a wrapper — with leading `VAR=value` assignments kept in front of `exec` where POSIX sh applies them. A genuine pipeline keeps its shell and `stopEntry` sweeps its direct children best-effort on stop. If that's not tight enough for the handoff path, happy to go further.

**6. Capture group** — Agreed: group 1 wins when it participated (`m[1] ?? m[0]`), the settings hint says so, and the field is capped at 300 chars so a pathological pattern falls back to the generic scrape instead of running per output line.

**7. Metrics-line leak** — Good catch; `parseTunnelMetricsUrl` now runs only for the built-in provider, with a test proving a custom CLI printing a metrics-server line keeps its public-URL probe.

**8. Changelog** — Trimmed to two sentences; the date directory moves to the actual merge day before this leaves draft.

On the summary nitpicks: `checkTunnelBinary` is deleted rather than carried (repo policy agrees with you), `SystemSettingsSection` has a renderer test now (provider select, conditional fields, both-fields persist, the blank-command warning), and the `cloudflare-tunnel.ts` rename stays a follow-up as you suggest.

`bun run lint` clean; backend, renderer and CLI suites green.

src/bun/tunnel-provider.ts
h0x91b: Verified closed. I ran the new `extractTunnelUrl` + regex over nine shapes, including the two that failed before:

- `https://machine.tailnet.ts.net/` -> `https://machine.tailnet.ts.net/`
- `Public: https://x.corp.example.com/app` -> `https://x.corp.example.com/app`
- `Visit https://x.example.com.` -> `https://x.example.com` (prose stop stripped)

and the JSON / `->` / bare-port shapes all still scrape correctly. End to end with a Tailscale-shaped fake tunnel: connected in 421 ms, where it used to sit for the full 30 s timeout. Nice catch on stripping the trailing punctuation too — I hadn't asked for that.

src/bun/remote-access-server.ts
h0x91b: Verified closed, and the gating does exactly what #1498 §3 described. Ran `checkOrigin` under both providers:

| Case | default (Cloudflare) | custom configured |
|---|---|---|
| `Origin` == `Host` | `true` | `true` |
| legit BYO tunnel via `X-Forwarded-Host` | `false` | `true` |
| attacker sets both headers to `evil.com` | **`false`** | `true` |
| attacker on plain LAN sets both | **`false`** | `true` |
| foreign `Origin`, no forwarded header | `false` | `false` |

The default posture is back to `Origin === Host`, so installs that never opt in are unaffected — which was the whole point. The remaining `true`s under `custom` are the accepted trade-off, and they still need a client that can win a CORS preflight. The 5-second cache is a sensible answer to the per-request `loadSettingsSync` concern.

One tiny thing, not worth its own thread: the comment on `_resetCustomProviderCache` says "for tests and settings saves", but nothing calls it on a settings save — only the tests do. The TTL covers it, so it's the comment that's off, not the behaviour.

src/bun/settings.ts
h0x91b: Verified closed, and `misconfigured` as a third kind is a better shape than what I suggested. Ran it: with `provider: custom` and a blank command, `resolveRemoteTunnelProvider()` returns `misconfigured`, `startTunnel` returns `null` in 1 ms with `failureReason: command-empty`, and **zero** `cloudflared` processes get spawned. That's the case that mattered — someone on a network that forbids `cloudflared` no longer gets one handed to them by accident.

Making the settings inputs controlled and persisting both fields together also fixes the UI-disagrees-with-backend half I mentioned at the bottom of the original comment.

src/bun/cloudflare-tunnel.ts
h0x91b: Verified closed. `extractTunnelUrl("t=1 url=https://abc.ngrok-free.app tail", compileUrlPattern("url=(https://\\S+)"))` returns `https://abc.ngrok-free.app` — group 1 wins when it participated, whole match otherwise.

The 300-character cap on the pattern covers the ReDoS aside I folded into this thread; a 400-character pattern falls back to the generic scrape with a warning, confirmed.

src/bun/cloudflare-tunnel.ts
h0x91b: Verified closed — the `parseMetrics` flag threads through both stream monitors, and there's a test asserting a custom CLI printing a metrics-server line keeps its public-URL probe.

src/bun/tunnel-provider.ts
h0x91b: Half closed, and there's a new edge in the fix itself — keeping this open for both.

**The exec prefix works for plain commands.** Measured, fresh settings per run, counting surviving processes after `stopTunnel()`:

| Command | Orphans after stop |
|---|---|
| `tunnel.sh {port}` | 0 |
| `tunnel.sh {port} \| cat` | **1** |

**The pipeline sweep doesn't fire.** The survivor has `ppid=1`. `entry.process.kill()` drops `sh` immediately, its children are reparented to init, and by the time `pkill -P <pid>` runs on line 579 there is nothing left under that pid to match. The sweep is racing something it cannot win — it needs the child list captured *before* the kill, or the process spawned into its own group with the signal sent to the group.

**New, and this one is a small regression:** the assignment-preserving regex on line 128 uses `\S*` for the value, so a quoted value containing a space splits in the wrong place and `exec` lands inside the quotes:

```
user typed:  FOO="a b" ngrok http {port}
line built:  sh -c 'FOO="a exec b" ngrok http 4242'
```

Confirmed against real `sh`: `FOO=a exec b` instead of `FOO=a b`. The variable silently carries the wrong value, and `exec` doesn't apply either. Before this commit that form ran correctly (it just orphaned), so it's a step back for that shape.

Cheapest fix is to not try: bail out of the exec prefix when the line contains a quote, right next to the existing control-operator check.

```js
if (/[|;&<>()]/.test(line) || /["']/.test(line)) return ["sh", "-c", line];
```

You lose the prefix for quoted-path commands, but those already worked — `sh` execs a lone simple command by itself, which is why the plain row above measured 0 orphans both before and after this commit.

src/bun/tunnel-provider.ts
h0x91b: Removing the probe was the right call and the quoted-path case now starts fine — I ran `"/opt/my tools/tunnel.sh" {port}` end to end and it connected.

Keeping this open because the replacement doesn't actually fire. `lastExitCode` is read on line 544, but `proc.exited` resolves *after* the output streams close, so at that moment it's still `null`. Real run against a missing binary:

```
20:10:40.145  WARN  Tunnel URL not found in process output within timeout
              startTunnel -> null   failureReason=no-url        <- expected command-not-found
20:10:40.185  INFO  Tunnel process exited {"exitCode":127}      <- 40 ms too late
```

So `remote.customTunnelNotFound`, its hint and the *Tunnel settings* link are unreachable in practice — a user with a typo in the command lands on the generic branch instead. Which is a shame, because that message is the good part of this change.

The unit test passes because it mocks `exited: Promise.resolve(127)` — already settled, so the `.then` runs on the first microtask and the value is there in time. Worth reshaping so `exited` settles *after* the streams close, otherwise the next version of this will sail past the test the same way.

Fix is to wait for the exit briefly before classifying, once the URL wait has come up empty:

```js
await Promise.race([proc.exited, delay(200)]);
const notFound = isCustom && (lastExitCode === 127 || lastExitCode === 126);
```

change-logs/2026/08/25/feature-byo-tunnel-origin-check.md
h0x91b: Length is good now — thanks.

Still open on the other half: the directory is `change-logs/2026/08/23/`, and the changelog UI groups by ship date, so it wants the expected merge day. Worth a `git mv` right before this goes in.

h0x91b (changes requested):
Round 2 — Claude again. Thanks for turning this around so fast, and for going past what I asked in a couple of places: stripping prose punctuation off the scraped URL, and making `misconfigured` a real third kind rather than the special-case I suggested. That's a better shape than what was in my comment.

I re-checked everything by running it, not by reading the diff. **Five of the eight threads are verified closed and I've resolved them**, each with a note on how I checked. `bun run lint` is clean, the touched suites pass locally (143 backend, 371 renderer), and CI is 11/11 green on `89634a6`.

## Verified closed

| Thread | How I checked |
|---|---|
| URL scrape | 9 shapes through the real function; the Tailscale form connects in 421 ms instead of timing out at 30 s |
| Origin gating | `checkOrigin` under both providers — the attacker-sets-both case is now `false` on the default path |
| Fail-closed config | Blank command yields `misconfigured`, no tunnel, and zero `cloudflared` processes spawned |
| Capture group + pattern cap | Group 1 wins; a 400-char pattern falls back with a warning |
| Metrics line ignored for custom | `parseMetrics` flag threads through both monitors |

## Three still open

None of these is severe, and I want to be clear about that — no security angle, and the default Cloudflare path is untouched. I'm holding the gate only because two of the three were introduced by this commit and each is a couple of lines:

1. **`command-not-found` never fires** — `lastExitCode` is read before `proc.exited` settles, so the nice new "command was not found" message and its settings link are unreachable. Measured, with the 40 ms gap in the log. Details and a fix in the PATH-probe thread.
2. **`VAR="a b" cmd` gets `exec` spliced into the quotes** — the value regex splits on the space inside the quotes, so the variable silently carries `a exec b`. That form worked before this commit. Details in the orphan thread.
3. **The pipeline sweep doesn't fire** — `pkill -P` runs after `sh` is already gone and its children are reparented to init. Measured: plain command 0 orphans, pipeline 1. Not a regression, but not fixed either, and the commit message says it is.

Plus the changelog directory date, which is a one-line `git mv` before merge.

One more that isn't worth a thread: the comment on `_resetCustomProviderCache` says "for tests and settings saves", but only tests call it. The 5-second TTL makes that harmless — it's the comment that's inaccurate.

Also worth saying: this PR is still marked draft, so whatever we conclude here, it needs taking out of draft before it can go anywhere. Happy to do another pass whenever you've pushed.

arditti:
Round 2 addressed in `154b22e4`, and the PR is out of draft — thanks for holding the gate on exactly the right three; two were genuinely my round-1 bugs.

**1. `command-not-found` never fired** — You called the mechanism precisely: the streams close before `proc.exited` settles, so the classification read a `null` exit code every time. It now waits briefly (`Promise.race([proc.exited, delay(200)])`) before classifying — no stall in the URL-timeout case worth caring about, since the interesting path is a process that is already dead. And per your test note, the mock's `exited` now settles on a timer *after* the streams close, like a real process — the old already-resolved promise would have let the next regression sail through the same way.

**2. `VAR="a b"` splicing** — Took your cheapest fix verbatim: any quoted line skips the `exec` prefix entirely, next to the control-operator check. You're right that nothing is lost — a lone quoted command is exec'd by `sh` itself, which is why your quoted-path row measured zero orphans both before and after. Both shapes are pinned in tests now: `FOO="a b" tunnel {port}` passes through untouched, and the quoted-path form stays prefix-free.

**3. Pipeline sweep** — The children are now captured with `pgrep -P` *before* the wrapper shell is killed, then signalled individually — your diagnosis again, the sweep was racing a reparenting it could not win. Measured live the same way you did: pipeline stop now leaves **zero** orphans (was one). Still best-effort and direct-children-only — a pipeline segment that spawns its own grandchildren is beyond it — but the commit message no longer claims more than the code does.

The trivia too: the changelog directory moved to the expected merge day, and the `_resetCustomProviderCache` comment now says what's true (tests only; production rides the TTL).

Lint clean, backend/renderer/CLI suites green on `154b22e4` (rebased on today's main). One unrelated observation from running the full suite repeatedly: `FolderPickerModal.test.tsx` ("refuses a typed path that escapes the root") flakes roughly one run in three on a loaded machine and is untouched by this branch — mentioning it only so a red run there doesn't get attributed to this PR.

src/bun/tunnel-provider.ts
h0x91b: Verified closed, both halves.

**The quoted-value splice is gone.** Bailing out on any quote is exactly right, and the comment explaining *why* a lone quoted command needs no prefix is the part I'd have missed. Straight out of `buildCustomTunnelArgv`:

| Command | argv line |
|---|---|
| `ngrok http {port}` | `exec ngrok http 4242` |
| `FOO=bar ngrok http {port}` | `FOO=bar exec ngrok http 4242` |
| `FOO="a b" ngrok http {port}` | `FOO="a b" ngrok http 4242` |
| `A=1 B="x y" tunnel --port {port}` | `A=1 B="x y" tunnel --port 4242` |

Through real `sh`, `FOO="a b" env` now reports `FOO=a b` — it was `a exec b` before. Unquoted assignments keep the prefix, so nothing was traded away.

**The sweep now fires.** Capturing the children first is the fix. Paired run of the same pipeline on both commits, snapshotting `ppid` around `stopTunnel()`:

| | before stop | after stop |
|---|---|---|
| `0716a18a` (before) | child ppid = wrapper sh | child **ppid 1**, still alive |
| `154b22e4` (after) | child ppid = wrapper sh | **gone** |

Quoted-path command also starts and stops clean: 0 survivors.

One small thing, not worth holding anything up: the new test asserts that `pgrep` and the kill both happened, but not that `pgrep` came *first* — and the order is the whole invariant. I moved the capture back below the kill and all 68 tests still passed. It does catch a revert to `pkill -P`, so it isn't toothless; an `invocationCallOrder` check (or spying the kill to run `pgrep` again and expecting empty) would close the gap if you ever touch this again.

src/bun/tunnel-provider.ts
h0x91b: Verified closed. Same missing-binary run as before, this time it classifies:

```
before  154b22e4:  url null   reason no-url              (exit code arrived 40 ms late)
after   154b22e4:  url null   reason command-not-found   in 31 ms
```

So `remote.customTunnelNotFound`, its hint and the *Tunnel settings* link are reachable now, and the 200 ms race costs nothing on the path that matters — the process is already dead by the time the URL wait comes up empty.

Thanks for reshaping the mock rather than just moving the line. I checked that it earns its keep: deleting `await Promise.race([proc.exited, delay(200)])` from the source turns that test red. It's a real guard now, not a green light.

change-logs/2026/08/25/feature-byo-tunnel-origin-check.md
h0x91b: Verified closed — `change-logs/2026/08/25/`, which is today and the merge day. Thanks.

h0x91b (approved):
Round 3 — Claude here again. All eight threads are verified closed and resolved, and this is an approve.

I re-checked the three that were open by running them, not by reading the diff, and each one landed:

| Was open | How I checked it closed |
|---|---|
| `command-not-found` never fired | Real missing-binary run: `no-url` before, **`command-not-found` in 31 ms** after |
| `exec` spliced into a quoted value | `sh -c 'FOO="a b" env'` now reports `FOO=a b`; it was `a exec b` |
| Pipeline sweep didn't fire | Paired run on both commits: child ends at **ppid 1 and alive** before, **gone** after |

Two things worth saying out loud, because they're the difference between a fix and a fixed *thing*:

**Your reshaped mock actually guards the code now.** I deleted `await Promise.race([proc.exited, delay(200)])` from the source and the test went red. The old `Promise.resolve(127)` version would have sailed straight past that. Reshaping it instead of just moving the line is the right instinct.

**The one gap left is a nitpick, and I've noted it inline rather than held anything up.** The new sweep test asserts `pgrep` and the kill both happened, but not that `pgrep` came first — and the ordering *is* the invariant. I moved the capture back below the kill and all 68 tests still passed. It does catch a revert to `pkill -P`, so it isn't toothless. Worth tightening whenever you're next in there; not worth another round.

`bun run lint` is clean, the touched backend suites pass (133), and CI is green on `154b22e4` across all four required contexts plus both e2e jobs. Branch is one commit behind `main` with zero overlapping files.

Thanks for three careful rounds on this — including the two places you went past what I asked (stripping prose punctuation off the scraped URL, and making `misconfigured` a real third kind).

PR #1505, opened 2026-08-23
Show agent-to-agent traffic as a hub-and-legs composition of task boxes

h0x91b:
A coordinator driving several tasks used to get one identical violet toast per `dev3 message` — five cards saying the same thing, hitting the visible-toast cap, and never answering who was talking to whom.

An agent toast is now a composition of task boxes rather than one notification card:

- The task at the centre of the traffic is a box on the left; every counterpart is a box on the right, split under a `received` / `sent` label.
- The message text lives in the channel between them, on an orthogonal bracket with a slow dot travelling in the message's own direction.
- Every box carries `#seq` and the task title, and opens that task on click — the sending side included, which needed one new field (`fromTaskId`) on the `agentMessage` push.
- Repeat messages from one counterpart fold into a single box with `×N` and the newest text, so a burst of five is one box, not five.
- Grouping is by hub and by toast variant: an error or success toast landing mid-burst is neither folded in nor treated as the end of it.

Capped at two counterparts per direction with a `+N` line — height is this form's price (four legs measure ~314px), and the card is 32rem wide instead of 26rem.

Not implemented, deliberately: the status dot from the mock. The push event carries no task status and the sender may live in another project, so the renderer cannot resolve it honestly.

Reasoning and the five rejected compositions: `decisions/2026/08/23/agent-toast-groups-by-hub-not-sender.md`.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=5ade716a-2b8b-460d-a2a2-d56cf749accd) · `dev3://task/5ade716a-2b8b-460d-a2a2-d56cf749accd`


h0x91b:
Hey, Claude here — the AI that built this branch.

Closing this without merging: the visual direction was rejected by the repo owner, and the follow-up work moved to a separate task exploring several concepts in parallel. Nothing here is broken — CI was fully green on the head commit — it is simply the wrong shape.

Keeping the branch on origin rather than deleting it, because one piece is worth salvaging regardless of what the final visual looks like: the `fromTaskId` field added to the `agentMessage` push payload, which is what lets a notification open the *sending* task and not just the receiving one. The reasoning for the grouping rules is in `decisions/2026/08/23/agent-toast-groups-by-hub-not-sender.md` on this branch.

PR #1530, opened 2026-08-25
Release task ports under the cross-process assignment lock

Tyagiquamar:
## Problem

\
eleasePorts()\ in \src/bun/port-pool.ts\ rewrote \~/.dev3.0/port-assignments.json\ from the in-memory cache without taking the cross-process file lock that \llocatePorts()\ has held since the concurrent-allocation race was fixed (see the comment above \llocatePorts\, and the existing regression test \does not assign overlapping ports to concurrent allocations\).

The same lost-update race therefore still existed on the release path, across two app instances sharing \~/.dev3.0/\ (installed app + dev build — a supported setup per the on-disk invariants):

1. Instance A allocates ports for task X (under the lock) and persists.
2. Instance B concurrently releases task Y's ports. It loaded its cache earlier, deletes Y, and writes the whole stale map back — erasing A's just-persisted record for task X.
3. A later allocation no longer sees task X's ports as assigned and can hand them to another task → two tasks get overlapping \DEV3_PORT0\ values, which is exactly the dev-server bind-clash failure mode the allocate-side fix describes.

Found by direct code inspection while auditing cross-process state mutation paths (this file's allocate path was already hardened; release was missed).

## Fix

Mirror \llocatePorts\: run the read-decide-write section of \
eleasePorts\ inside \withFileLock(ASSIGNMENTS_FILE, ...)\ with a fresh \
eadFromDisk()\ inside the critical section. The function becomes async; its single production caller (the lifecycle executor's \
eleasePorts\ effect) now awaits it.

## Testing

- New deterministic regression test \does not drop a peer allocation that landed while releasing\ in \src/bun/__tests__/port-pool.test.ts\: allocates, mutates the assignment file behind the in-memory cache (simulating a peer process), releases, then asserts the peer's record survives a fresh reload.
- Mutant check: with the old implementation restored, the new test fails (\1 failed | 20 passed\); with the fix, all 21 pass:
  \\\
  bunx vitest run --config vitest.config.bun.ts src/bun/__tests__/port-pool.test.ts
  Test Files  1 passed (1)
  Tests       21 passed (21)
  \\\
- Adjacent suites that mock/exercise \
eleasePorts\ (\cli-socket-task-move-cleanup.test.ts\) pass. \
ative-teardown.test.ts\ fails identically (7) on unmodified \main\ on this Windows machine — POSIX-only cleanup-script session by design, pre-existing and unrelated.
- Type check: \	sc --noEmit\ reports zero errors under \src/\.

### DIFF
--- a/change-logs/2026/08/25/fix-port-release-lost-update.md
+++ b/change-logs/2026/08/25/fix-port-release-lost-update.md
@@ -1,0 +1,1 @@
+Releasing a task's ports no longer races a second app instance allocating ports at the same moment: the release now takes the same cross-process lock as allocation and re-reads the assignment file first, so a peer's just-persisted task can no longer be erased from port-assignments.json and its ports handed out twice.
--- a/decisions/2026/07/04/port-alloc-file-lock.md
+++ b/decisions/2026/07/04/port-alloc-file-lock.md
@@ -1,5 +1,11 @@
 # 102 — Serialize port allocation with a cross-process file lock
 
+> **Superseded in part by
+> [`decisions/2026/08/25/release-ports-under-the-assignment-lock.md`](../../08/25/release-ports-under-the-assignment-lock.md).**
+> The sentence below claiming `releasePorts()` keeps using the in-memory cache is
+> no longer true: release now takes the same lock and re-reads from disk. The
+> sync getters do still read the cache.
+
 ## Context
 
 `allocatePorts()` (`src/bun/port-pool.ts`) is async and `await`s an OS-level
--- a/decisions/2026/08/25/release-ports-under-the-assignment-lock.md
+++ b/decisions/2026/08/25/release-ports-under-the-assignment-lock.md
@@ -1,0 +1,64 @@
+# Release task ports under the assignment lock
+
+Supersedes part of
+[`decisions/2026/07/04/port-alloc-file-lock.md`](../../07/04/port-alloc-file-lock.md),
+which stated that only allocation is contended.
+
+## Context
+
+`allocatePorts()` has taken `withFileLock(ASSIGNMENTS_FILE, ...)` and re-read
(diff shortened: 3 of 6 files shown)


h0x91b (approved):
Thanks a lot for the first contribution — and a good catch. The release path really was the half that got missed when allocation was hardened, and the regression test is the right kind of test: I restored the old implementation locally and it fails, so the guard genuinely guards something.

I pushed two follow-ups onto your branch before merging:

- `releasePorts` now catches `FileLockTimeoutError` instead of letting it escape. The lifecycle effect runs with the `continue` policy, so a rejection would only show up as a generic warning while the assignment stayed on disk forever with nothing to prune it. It is now swallowed with an explicit `Ports left assigned` error line; non-timeout failures still propagate.
- Decision records. `decisions/2026/07/04/port-alloc-file-lock.md` stated outright that release keeps using the in-memory cache, which your change makes false, so it now carries a supersede note pointing at a new record, `decisions/2026/08/25/release-ports-under-the-assignment-lock.md`, covering the rationale, the orphan-record risk, the older-build caveat and the alternatives I turned down (boot-time pruning would have reintroduced the same lost update in a worse form).

Also rebased on main. Merging this — thanks again, and please keep them coming.

Tyagiquamar:
Thanks for the fast review and for landing the follow-ups yourself — appreciated.

The `FileLockTimeoutError` catch makes sense: I hadn't traced what the `continue` policy does with a rejection there, so a swallowed failure would have left the assignment stuck on disk forever — exactly the class of bug this fix targets. And fair point on the decision records; I'd missed `port-alloc-file-lock.md` while tracing why release still used the cache, so having the supersede chain in place helps the next person.

I'll keep going through the lifecycle/executor paths with the same treatment — anything design-level rather than an obvious bug, I'll raise as an issue first.

PR #1540, opened 2026-08-26
Send task notifications to a command, a webhook, or your phone

mcaldas:
## Why

A task notification only reaches you if something is already listening. `postNativeTaskNotification`
wants the desktop in front of you; `pushWebNotification` wants a browser tab that is open and
connected. Close the lid and the event is gone — which is exactly when an agent blocked on a question
costs the most.

`deliverTaskNotification` (`src/bun/rpc-handlers/shared.ts`) already had the hard parts: one funnel,
per-project silencing, foreground suppression, a queue that flushes on background. What was missing
was a destination that survives having no listener.

## What changed

**Two destinations, one call site.** Both hang off `outboundNotify`, called at the end of
`deliverTaskNotification` after the existing silence and suppression gates, so a muted project stays
muted.

**A command or webhook**, configured in `~/.dev3.0/notifications.json`. Works in every mode including
a plain-HTTP LAN address, which is the case Web Push can never serve. Commands are argv-only with the
event as JSON on stdin — a task title is user-authored and must never reach a shell parser. A hook
that hangs, crashes or 500s is logged and dropped; it cannot fail the task transition that triggered
it.

Config deliberately does **not** live in `GlobalSettings`: `saveSettings` is reachable from the RPC
surface `dev3 remote` serves to an authenticated browser, so an `exec` transport stored there would
let a session plant a command dev-3.0 re-executes on every future notification. Not a new capability
for a session that already has `/pty`, but a persistent and invisible one.

**Web Push**, hand-rolled on WebCrypto — VAPID (RFC 8292) and aes128gcm (RFC 8291) — rather than
taking `web-push` and its 17 transitive packages, on the precedent of `src/bun/jwt.ts`. The keypair is
generated once into `$DEV3_HOME/web-push-keys.json` at 0600; rotating it orphans every device.
Registration is gated exactly like `/rpc` (origin, then session), because a subscription is a standing
capability to wake someone's phone and the remote server serves its shell to unauthenticated
browsers. Devices are keyed by endpoint so a re-subscribe replaces rather than double-buzzes, and are
pruned only on 404/410 — a 429 or a 503 must not silently unsubscribe everyone.

**Discovery.** Enrolment lived four steps deep in Settings, and testing on a real phone showed someone
who knew the feature existed and wanted it still did not reach it. The first time a browser could
accept push, a toast offers it and one tap runs permission, subscribe and register. `ux-principal`
keeps durable configuration in Settings, so the toggle stays where it is and only discovery is added;
a toast rather than a modal because this is an offer, not a decision blocking the user.

The gating is the design: offered once and never again, taken or ignored; silent where accepting could
not work (an iOS Safari tab, a plain-http origin), because an offer that cannot succeed is what
teaches people to dismiss prompts unread; silent on an existing browser-level denial; and silent when
storage is unavailable, since without a way to remember a dismissal the alternative is nagging on
every load.

**AGENTS.md gains the boundary this relies on.** The telemetry rules govern egress dev-3.0 chooses the
destination for. Data sent where the *user* pointed it is theirs, with two conditions that keep that
from being a loophole: never a default endpoint, and any relay must find the payload unreadable.

## Verification

Proven end to end on a real iPhone against a real `dev3 remote`, not a prototype: toast → one tap →
iOS permission → `/push/subscribe` behind the session gate → `Push device registered` → `dev3 notify`
→ `deliverTaskNotification` → `outboundNotify` → notification on a locked phone. Both negative cases
too: with the config removed nothing is sent anywhere, and a `"command": "rm -rf /"` string is
rejected by validation while the app keeps delivering its own notifications.

Two bugs that testing on hardware caught and unit tests could not:

- **Apple validates the VAPID `sub` claim; Google does not.** `mailto:…@localhost` gets 201 from FCM
  and `403 BadJwtToken` from Apple. Desktop-Chrome-only testing ships this broken to every iPhone.
- **An accepted push logged nothing**, so a working push path and a silent one looked identical from
  the server.

`web-push.test.ts` implements the *receiving* half of RFC 8291 independently and decrypts what the
sender produced, because a push service returns 201 without ever decrypting the payload — a green send
proves the JWT and nothing about the ciphertext.

## Also here

The renderer test suite was failing wholesale on Node 25.6+: the substitute `localStorage` installed
only when Node's experimental global was `undefined`, but newer Node exposes an object whose methods
are missing, so the shim never ran and 148 of 297 files died on `localStorage.getItem is not a
function`. Unrelated to notifications, kept in this PR to stay within one changelog entry.

## Known limits

- **Stable origin is documented, not validated.** A quick tunnel's hostname changes per `cloudflared`
  process; the old registration then keeps firing notifications for an origin that can no longer load.
  Nothing fails loudly. `docs/remote-access.md` covers Tailscale serve, a named Cloudflare tunnel and
  a reverse proxy — including that a node with Funnel enabled must serve on 443, since Funnel's public
  DNS points at ingress servers that only terminate TLS there. That last one cost an evening to find.
- **Long-idle delivery** is verified to 20 minutes force-quit, not to hours.
- **The exec/webhook transport** was driven end to end in a QA instance, not against a phone hook.
- **Notifications still only fire for watched tasks** or `dev3 notify`; this PR changes where they go,
  not when they happen.


src/mainview/test-setup.ts
```diff
@@ -16,12 +16,13 @@ afterEach(() => {
...
-// setupFiles run per test file, so each file gets a fresh, isolated store, which
-// is the isolation happy-dom's own storage provided.
-if (typeof globalThis.localStorage === "undefined") {
+// Node's experimental global `localStorage` shadows happy-dom's on both
+// globalThis and window (sessionStorage is unaffected). Without
+// --localstorage-file it is unusable, but not always absent: on Node 25.6 it is
+// an object whose methods are undefined, so probe for a working Storage rather
+// than for undefined. Substitute an in-memory one — setupFiles run per test
+// file, so each file gets a fresh, isolated store.
+if (typeof globalThis.localStorage?.getItem !== "function") {
```
h0x91b: Blocker: this checks only whether getItem exists, not whether storage is usable. In happy-dom that leaves the native Storage object in place, so the new stays quiet when storage is unavailable test cannot replace the instance method and fails locally and in CI. The same shape exists in production when reads work but setItem throws: remember() cannot persist dismissal and the invite can return every load. Please use a guarded write/read/delete probe here and make the test mock the property or prototype reliably.

src/bun/web-push.ts
```diff
@@ -0,0 +1,179 @@
...
+	if (existsSync(path)) {
+		try {
+			const parsed = JSON.parse(readFileSync(path, "utf-8")) as VapidKeys;
+			if (parsed?.publicKey && parsed?.privateKey) return parsed;
+			log.warn("Ignoring malformed VAPID key file", { path });
+		} catch (err) {
+			log.warn("Could not read VAPID keys", { path, error: String(err) });
+		}
+	}
+	const keys = await generateVapidKeys();
```
h0x91b, line 90: Blocker: key creation is racy. I reproduced eight concurrent calls on one absent path returning eight different public keys; only one of those keys survived on disk. A browser can therefore subscribe with a key that is immediately orphaned, so pushes silently fail. Please single-flight this in-process and use exclusive or atomic cross-process creation, then re-read and return the winning persisted key.

h0x91b (changes requested):
Requesting changes for two release blockers: the storage-unavailable test is red and dismissal persistence can repeatedly nag; concurrent VAPID creation can orphan subscriptions. Both are reproduced and documented inline. Please fix both before another review pass.

src/mainview/utils/pushInvite.ts
```diff
@@ -0,0 +1,73 @@
...
+	if (alreadyAnswered()) return;
+	if (!pushReadiness().ready) return;
+	// "denied" is a decision the user already made in the browser; re-asking is
+	// noise, since the prompt will never appear again anyway.
+	if (typeof Notification === "undefined" || Notification.permission === "denied") return;
+	if (await isSubscribed()) return;
+
+	toast.info(t("push.inviteBody"), {
+		durationMs: 15_000,
+		onClick: () => {
```
h0x91b: Blocking product-flow issue: this toast currently performs durable, multi-step setup on click (browser permission plus push enrollment). Toasts may point to their owning surface, but they must not own primary or multi-step actions. Please make this click open a registered Settings → System → Notifications/Push entry only, and keep both permission and enrollment behind explicit controls on that settings screen. The nudge can still be shown once; it should only navigate.

src/mainview/test-setup.ts
mcaldas: Fixed in 8f1e6e9.

`test-setup.ts` now decides with a write/read/delete round trip (`localStorageWorks()`) instead of sniffing for a `getItem` method, so a present-but-unusable Storage gets the in-memory substitute. The test no longer patches an instance method — happy-dom’s Storage proxy swallows that — it replaces the `localStorage` property itself on both `globalThis` and `window` and restores the saved descriptor afterwards.

The production shape is fixed at the root rather than in the test: `pushInvite` has one `usableStorage()` probe, and a store it cannot write is treated exactly like one it cannot read (silence), so a dismissal that could never be persisted no longer turns into an offer on every load. Two tests cover it now — read throws, and read-works-but-`setItem`-throws.

src/bun/web-push.ts
mcaldas: Fixed in 8f1e6e9.

`loadOrCreateVapidKeys` is now single-flight per path: the first caller’s promise is cached in a `Map` and every concurrent caller awaits it, so eight calls on an absent path can no longer mint eight keypairs. Cross-process, the write uses `flag: "wx"` — the loser of the race gets `EEXIST` instead of overwriting, re-reads the file, and returns the persisted winner. If the file cannot be created for any other reason the old warn-and-continue behaviour stands.

Two tests in `web-push.test.ts`: eight concurrent calls all return the key that is on disk, and a contended case that deliberately defeats the single flight (two spellings of one path, which is the shape of a second process) asserting both callers end up on the persisted key. I checked the second one goes red if the write stops being exclusive.

src/mainview/utils/pushInvite.ts
mcaldas: Fixed in 8f1e6e9.

The toast no longer owns any setup. Its click dispatches `OPEN_SETTINGS_SECTION_EVENT` with `"system"` and nothing else — no permission request, no enrolment — landing on Settings → System, where allowing notifications and enabling push stay two explicit controls. Push also got its own registered entry (`push-notifications`, category `system`), so "push" now finds it in settings search rather than only "browser notifications". Copy changed to "tap to set it up" in en/ru/es; the now-dead `push.inviteDone` key is gone. The offered-once gating is unchanged.

Verified in a browser against a scoped dev instance: toast → click → Settings → System with `Notification.permission` still `"default"`, no prompt, no subscribe. A test asserts the dispatched detail is `"system"` and that `subscribeToPush` is never called.

h0x91b (approved):
The merge blockers are fixed. Lint, focused regression tests, and manual Settings QA pass after rebasing onto main.

PR #1541, opened 2026-08-26
Import an agent session that ran outside dev3 as a task

mcaldas:
Put an agent session that ran **outside** dev3 onto the board, continuing the actual conversation rather than retyping it.

Two entry points: an **Import** mode in the New Task modal, and `dev3 import` run from the session's own shell.

## The design decision

A session is identified by its **transcript**, never by its live process. Adopting a running agent is impossible on both terminal backends — the native host refuses a foreign pid, and on either backend `set-environment` only seeds *new* processes, so a live foreign agent can never receive `DEV3_TASK_ID` and would have no hooks, no `dev3 message`, no board automation.

The transcript is also the only discovery surface more than one harness has: dev3 already locates stores for claude/codex/gemini, while the Claude-only live-session registry has no counterpart anywhere.

Full reasoning, including the measurements: `decisions/2026/08/26/import-a-session-by-its-transcript.md`.

## The bug this found

Importing exposed a hole in `resolveResumableSessionId`. An imported conversation stays in the store of the directory it started in — `--resume` appends to the original transcript and only stamps the new cwd into new records. But the agent *does* create the new worktree's store directory, empty, holding just `memory/`. So the resolver saw a store that exists, found no ids in it, and healed a perfectly live id to `null` — which means `--continue`, which in a directory with no conversation starts a brand new one.

The import worked; the first recovery silently orphaned it. `PaneSessionEntry.sessionOriginCwd` now records where the conversation lives and the resolver checks there before falling back. Both new tests fail without the fix — one with the orphaning, one with a later local session hijacking the imported id.

## Scoping is an invariant, not a filter

A session is importable **only** into the project whose path contains its cwd. No cross-project import, and `dev3 import --project` is refused. Importing across projects would fork a branch in the wrong repository and open a card describing work that is not there.

Matching is prefix-on-a-path-boundary: a project at `…/playground/dev-3.0` must not claim `…/dev-3.0-scratch`, and its parent `…/playground` must not match it — both are real cases on a real board.

Measured on one: of 30 transcripts carrying a cwd, only 7 sit under a registered project. So **"no dev3 project owns this directory" is the common path**, and it is written as one — naming the directory, explaining the rule, giving the next step and listing what is registered. New exit code `17`.

## Providers

| | Locator | Parser | Resume by id | Ships |
|---|---|---|---|---|
| Claude | yes | yes, plus a free `ai-title` | yes | **yes** |
| Codex | yes | yes (no title, first turn used) | yes | **yes** |
| Gemini | yes | no | yes | needs a parser |
| Cursor / OpenCode | no | no | partial | no |

Uneven provider support is the house norm here — `agent-support-matrix.md` gains a row rather than the feature waiting for the slowest harness.

## Verifying

`bun run test` runs its three configs concurrently; on a loaded machine that produces timeout-shaped failures that look like regressions. Run them sequentially with a short `TMPDIR` (the default `/var/folders/…` path pushes unix socket paths past the macOS 104-char limit):

```
mkdir -p /tmp/d3t && export TMPDIR=/tmp/d3t
bunx vitest run --config vitest.config.bun.ts   # 6993/6993, 406 files
bunx vitest run --config vitest.config.cli.ts   #   889/889,  51 files
bunx vitest run --config vitest.config.ts       #  4685/4686, 298 files
```

The single renderer failure is `BehaviorSettingsSection`, a known timing flake that passes in isolation and is untouched by this branch.

The UI was driven in a browser, which caught two things the tests were happy with: `Use existing branch` still rendered in Import mode (contradicting the session's own branch), and the description placeholder still said "Describe what needs to be done" for a field about to be filled in. Both fixed.

## How to check it

1. **New Task → Import** on a project you have run an agent in — expect a session list with title, branch, turn count and date.
2. **Pick one** — title, description and branch fill in, all editable; `Use existing branch` is absent.
3. **Switch back to Describe** — original placeholder and branch picker return.
4. `cd` into such a repo and run **`dev3 import`** — expect a numbered table.
5. Run **`dev3 import` in a directory dev3 does not know** — expect the directory named, the rule explained, a next step, exit `17`. Not a terse error.
6. **`dev3 import --project <id>`** — expect refusal.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=01f966d1-a8db-47a2-b211-aafd0d44c0b9) · `dev3://task/01f966d1-a8db-47a2-b211-aafd0d44c0b9`

https://claude.ai/code/session_01Qxk78djwJDoFb8ZdHVpsk5


h0x91b:
Hey — Claude here, the AI assistant working on this repo, writing on Arseny's behalf.

First, thank you. This is a genuinely thorough contribution. The decision record measuring what actually happens rather than asserting it, the cwd-scoping rule with its path-boundary check, the exit-code contract followed to the letter — all of that was a pleasure to read. And the `resolveResumableSessionId` bug you uncovered on the way is real: an imported conversation stays in the store of the directory it started in, so the resolver could heal a live session id to `null` and quietly turn `--resume` into a brand-new conversation.

The awkward part is ours, not yours: we were building the same feature in parallel and it landed first — #1587, "Import Claude Code conversations that ran outside dev3 as tasks", merged on 28 Aug. It takes a different route (conversations arrive as ordinary tasks, and discovery is offered per project rather than per session), but it covers the same ground, so I'm closing this as superseded rather than leaving it open against a base that keeps moving under it.

Two pieces of your work we'd genuinely like to keep, if you're up for it:

- **The `sessionOriginCwd` fix in `resolveResumableSessionId`.** #1587 doesn't resume imported sessions, so it never hits this — but the bug bites any pane recovered from a session whose transcript lives elsewhere. A small standalone PR with just that change and its two tests would be very welcome.
- **`dev3 import`, run from the session's own shell.** #1587 ships `dev3 conversations import`, but "import the session I'm sitting in right now" is a nicer entry point and isn't covered.

Sorry for the collision. If you'd like to pick up either piece, say so here and we'll take a look.

PR #1549, opened 2026-08-26
Render .mmd Mermaid files like markdown in preview and diff

arditti:
Standalone Mermaid files (`.mmd`) now get the same rendered-document treatment as `.md`: the file preview modal draws the diagram with a **Rendered / Raw** toggle, and the diff viewer offers the per-file **Preview** toggle and previews the resulting diagram instead of raw diagram source. No new dependency — the markdown renderer already draws `mermaid` fences, so a `.mmd` file is fenced and handed to it.

Two deliberate differences from `.md`: diagram source has no prose blocks, so a modified `.mmd` never gets the rich prose diff (it always previews as the diagram), and the fence shifts line numbers, so the preview carries no per-line anchors for selection comments.

## Checklist

Full rules live in [`AGENTS.md`](https://github.com/h0x91b/dev-3.0/blob/main/AGENTS.md) and [`CONTRIBUTING.md`](https://github.com/h0x91b/dev-3.0/blob/main/CONTRIBUTING.md); this is the short list of things reviews most often bounce on. Delete the lines that do not apply.

- [x] Changelog entry under `change-logs/YYYY/MM/DD/` dated to the **expected merge day**, not the day work started (`feature-` entries need a `Short:` line)
- [x] Rebased on current `origin/main`
- [x] `bun run lint` clean and `bun run test:full` green — the plain `bun run test` skips the slow e2e suites that CI runs
- [x] User-facing strings go through `t()` and exist in all three locales (en / ru / es) — no new strings; the existing `Rendered`/`Raw`/`Preview` keys are reused
- [x] UI change: manual browser QA done; screenshots taken in streamer mode (`&streamer=on`)
- [x] PR is out of draft before requesting review

## Verification

`bun run lint` clean. Renderer 301 files / 4702 tests green, CLI 50 files / 868 tests green, backend 408/409 files green.

One backend file fails on my machine and I could not make it pass: `src/bun/__tests__/native-pane-owner-forward.test.ts` (10 tests), every one with `listen EINVAL` when binding its unix socket. The sandbox `TMPDIR` makes the socket path ~118 characters, over the ~104-byte `sun_path` limit, so it is a path-length limit rather than a code failure — it reproduces on an untouched tree and this change is renderer-only.

Browser QA in headless Chromium, streamer mode on, no console errors: a changed `.mmd` file in the diff viewer renders the diagram with **Preview** pressed and toggles back to the source diff; the file preview modal renders the diagram and switches to numbered source on **Raw**; `.md` and `.ts` files behave as before.

### DIFF
--- a/change-logs/2026/08/26/feature-45-mmd-mermaid-render.md
+++ b/change-logs/2026/08/26/feature-45-mmd-mermaid-render.md
@@ -1,0 +1,3 @@
+Short: Mermaid .mmd files render as diagrams
+
+Standalone Mermaid files (`.mmd`) now render the same way markdown does: the file preview you get from Cmd/Ctrl+Clicking a path opens the drawn diagram with a Raw toggle, and in the diff viewer a changed `.mmd` file gets the Preview toggle and shows the resulting diagram instead of raw diagram source.
--- a/src/mainview/components/FilePreviewModal.tsx
+++ b/src/mainview/components/FilePreviewModal.tsx
@@ -8,6 +8,7 @@ import { formatBytes } from "../utils/formatBytes";
 import { writeClipboardText } from "../utils/clipboard-write";
 import type { FilePreviewResult } from "../../shared/types";
 import { MarkdownDocument } from "./pr-review/markdown";
+import { isRenderableDocPath, toRenderableMarkdown } from "./pr-review/markdown-files";
 
 interface FilePreviewModalProps {
 	path: string;
@@ -69,7 +70,7 @@ export default function FilePreviewModal({ path, line, taskId, onClose }: FilePr
 		}
 	}, [preview]);
 
-	const isMarkdown = /\.(md|markdown)$/i.test(path);
+	const isRenderable = isRenderableDocPath(path);
 	const textContent = preview?.kind === "text" ? preview.content : null;
 	const slash = path.lastIndexOf("/");
 	const fileName = slash >= 0 ? path.slice(slash + 1) : path;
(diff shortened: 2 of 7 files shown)


h0x91b (approved):
Hey — Claude here, the AI assistant reviewing this branch for Arseny. Thanks for the change, and especially for the PR description: naming the two deliberate carve-outs up front (no rich prose diff, no per-line anchors) saved me from filing both of them as bugs.

Nice bit of it: the `.md` extension check now has exactly one home. I grepped the tree and there is no second `/\.(md|markdown)$/i` left anywhere under `src/` — that is the version of this change that does not rot.

Approving. Nothing blocking; two small things worth a follow-up.

**Worth fixing**

1. `markdown-files.ts` — `toRenderableMarkdown` fences with a fixed ```` ``` ````, so a `.mmd` file containing a line that is just ```` ``` ```` closes the fence early. I checked this against the real `remark-parse` rather than from memory: everything after that line parses as ordinary markdown (a heading node and an image node showed up outside the code block). Cheap fix — pick a fence longer than the longest backtick run in the content, or use `~~~`.

2. `TaskDiffViewer.tsx` — on a `.mmd` file the **Show in preview** button on a comment thread does nothing. Threads can exist there, since gutter comments in source-diff mode are unrestricted; in preview mode there are no `data-md-line-start` anchors, so `revealMarkdownLine` returns `false` and that return value is discarded. Either hide the reveal row when the document carries no anchors, or use the `false` to say something.

**Nitpicks, your call**

- `isMarkdownDiffFile` now returns `true` for a file that is not markdown. The new module introduced the right word ("renderable doc") and the predicate did not follow it.
- The toggle's accessible name reads "Toggle Markdown preview for docs/chart.mmd" — a screen reader says "Markdown" for a diagram file. Costs three locales, hence a nitpick.
- An empty `.mmd` in the file-preview modal renders an empty fence and lands in the mermaid error fallback, while the diff viewer has a proper "Nothing to preview" placeholder for the same case.
- Test gap around the two carve-outs: no case for content containing a fence, for empty content, or for an added/deleted `.mmd`.

For the record on my side: `bun run lint` clean, and the three touched suites green locally (8 + 89 tests). I did not run `test:full` or drive the browser myself — taking your word plus the green CI on this head for both.

PR #1550, opened 2026-08-26
Sign the CLI binary with the JIT entitlements bun:ffi needs

mcaldas:
## The bug

`dev3 remote` dies every time a task is created. The browser or phone attached to it
drops into "Reconnecting" — not a dropped WebSocket, the whole server process is gone.

Task creation reaches `clonefile(2)` through `bun:ffi` (`src/bun/cow-clone.ts`), and every
`bun:ffi` `dlopen` compiles a JIT trampoline. `scripts/sign-cli-binaries.sh` signed
`dist/dev3` with `--options runtime` and **no entitlements**, so macOS refuses the JIT
mapping and the process takes a `SIGTRAP`. It is a signal, not a JS throw — the `try/catch`
inside `tryClonefile` never sees it.

Latent since #470 turned on hardened runtime for the CLI binary (2026-04-15); `cow-clone.ts`
has used the FFI since #67.

## Why nobody hit it in development

- The app's own bun binary is signed by Electrobun **with** `allow-jit`, so every task
  created in the desktop app works. Only the CLI binary is affected, and only `dev3 remote`
  runs the clone in-process rather than handing off to the app over the socket.
- The ad-hoc local signature carried no hardened runtime at all, so `bun run dev` could never
  reproduce the configuration that ships. That branch is now hardened too.

## Evidence

A user's crash report (`EXC_BREAKPOINT` / `SIGTRAP`, `pthread_jit_write_protect_np`, main
thread) landed 7s after the server's last log line, which was inside `runCowClones`.

A/B on the real `clonePaths`, compiled and signed both ways:

| Signature | Result |
|---|---|
| hardened, no entitlements (what ships) | exit 133 (SIGTRAP), dead right after "Starting CoW clone" |
| hardened + both entitlements | exit 0, cloned via `clonefile(2)` in 3.1s |

`allow-jit` alone is not enough — `dlopen` then succeeds and the first FFI call is SIGKILLed,
so both keys are required.

## Why the signing layer, not `cow-clone.ts`

The app's own bun binary already carries exactly these two entitlements and runs the same
code; the headless server is the same application without a window. Granting the pair to one
process and denying it to the other is the inconsistency that produced the crash.

Dropping the FFI and letting the existing `cp -cR` rung handle it was measured, not assumed:
`clonefile(2)` clones a 1.2 GB / 92k-file `node_modules` in **3.1s**, `cp -cR` in **14.8s** —
about 12 extra seconds on every task creation.

Scope, stated honestly: `cow-clone.ts` is the only FFI site currently *reachable* in the
headless binary. `native-notifications.ts` returns early on `DEV3_HEADLESS=1`, `hard-exit.ts`
is GUI-only, and `hideAppNative` is a no-op in the browser transport. The others are latent,
not live — an argument for fixing the binary rather than the call site, but not the same as
them being armed today.

## Trade-off

`allow-unsigned-executable-memory` does weaken W^X for a binary that can be exposed over a
public tunnel. It is the same pair the app process already runs with, and neither key is a
restricted entitlement, so notarization is unaffected. If you would rather not grant it to the
server process, deleting the FFI is a legitimate alternative at the cost above — say so and
I will rework it.

## Release gate

Same reasoning as the Mach-O gate in `decisions/2026/07/05/macho-headerpad-codesign-surgery.md`:
this bug class is invisible to signing, notarization and green CI, so only executing the thing
catches it. Two independent checks in `release-build-macos.yml`:

1. Both keys survive into `Resources/app/cli/dev3` — verified to **fail** against the shipped
   v1.48.0 binary.
2. The real `clonePaths` is compiled into a probe, signed with the same recipe, and runs a
   clone — exit 0 with the plist, exit 133 without it.

## Verification

- `bun run lint` clean; full `bun run test` green (300 + 406 + 50 files).
- Both halves of the gate run locally, each confirmed to fail when the protection is removed.
- The newly-signed `dist/dev3` runs normally (`--version`, `current`), so hardening the ad-hoc
  branch does not break local use.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=9c2b6a0d-3581-423b-a312-1fde917a539b) · `dev3://task/9c2b6a0d-3581-423b-a312-1fde917a539b`

https://claude.ai/code/session_01QKB5WVWUWVbM4cX6LdPGvb


h0x91b:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** Prevent the hardened CLI binary from crashing dev3 remote when CoW cloning reaches bun:ffi.

**Current behavior:** The PR adds the two Bun JIT entitlements to the CLI signing recipe and a release gate that checks the shipped binary plus a real clone probe. GitHub reports no checks for the branch and no review has been recorded yet.

**Desired behavior:** The release and local ad-hoc signing paths should remain usable under hardened runtime, while the gate must fail if either entitlement is stripped or no longer sufficient. The security trade-off of allowing unsigned executable memory in the headless CLI must be explicitly accepted by a maintainer.

**Key interfaces:**
- The CLI binary signing recipe and entitlement plist — must stay aligned for Developer ID and ad-hoc signatures.
- The macOS release workflow gate — must exercise both entitlement presence and the real bun:ffi clone path.

**Acceptance criteria:**
- [ ] A maintainer reviews and accepts the hardened-runtime entitlement trade-off for the remote CLI.
- [ ] CI or equivalent macOS verification passes for the signed artifact and clone probe.
- [ ] Existing CLI signing tests remain green for Developer ID and ad-hoc paths.
- [ ] The gate fails under the missing-entitlement mutation and passes with the intended plist.

**Out of scope:**
- Replacing CoW cloning with a slower copy implementation unless the maintainer rejects the entitlement approach.
- Changing unrelated runtime or remote-session behavior.

h0x91b:
Hey — Claude here, the AI assistant working on this with Arseny. First: thanks for this one. The diagnosis is genuinely good work — the SIGTRAP-is-a-signal-not-an-exception detail is exactly why `tryClonefile`'s `try/catch` looked innocent, and I appreciate that the decision record measured `cp -cR` (14.8s vs 3.1s) instead of hand-waving the alternative. The release gate that actually *executes* a signed probe is the right instinct too.

I pushed one commit on top (`a6817c7`) fixing a blocker I hit while reviewing, plus a rebase on current `main`.

**The blocker:** `"${ent_args[@]}"` on an *empty* array under `set -u` is an "unbound variable" error in bash 3.2 — and `/bin/bash` on macOS is 3.2.57. Only `dist/dev3` gets an entitlements plist, so the script signed the CLI and then died on the very next call:

```
[sign-cli-binaries] ad-hoc signed: dist/dev3
scripts/sign-cli-binaries.sh: line 54: ent_args[@]: unbound variable
SCRIPT EXIT=1
```

`bun run build:cli` fails there, and the bundled tmux and the bifrost sidecar never get signed. Whether release CI survives it depends purely on which `bash` is first on PATH (`build-cli.ts` calls a bare `\["bash", script\]`) — brew's 5.x expands an empty array happily, system bash does not. Not something to find out on a release.

The fix is the bash-3.2-safe idiom `${arr[@]+"${arr[@]}"}` at both call sites.

**Why the tests didn't catch it:** both existing cases only put `dev3` in `dist/`, so `sign_one dist/tmux/tmux` hit the `return 0` for a missing file *before* the array was ever expanded — three green tests, broken build. I added a case that stages all three binaries and runs the script through `/bin/bash` explicitly (PATH could hand it a modern bash and hide the whole thing). Verified as a real guard: reverting just the expansion turns it red on the exact line.

Also moved the changelog entry to `2026/08/27` so it lands in the day it actually ships.

Three smaller things I did **not** change, entirely your call, and none of them blocks the merge:

1. In the new gate, `set -euo pipefail` + `CLI_BIN=$(find "$RUNNER_TEMP/eb-payload" ./build … 2>/dev/null | head -1)` — if `eb-payload` was never created (the previous step only emits a `::warning::` when there is no `*.tar.zst`), `find` exits 1, `pipefail` propagates it, and `set -e` kills the step *before* your `::error::no CLI binary found` line, with `2>/dev/null` swallowing the reason. A trailing `|| true` restores the message you wrote.
2. The gate checks the in-bundle copy but not `dist/dev3`, which is what `create-cli-tarball.sh` ships — and the tarball is what a headless box installs, i.e. the machine where this bug actually bit.
3. The probe fails when `method !== "clonefile"`, which is also true on a non-APFS volume — same error text, different cause.

Suite status on the rebased branch: renderer 4791 passed, CLI 900 passed. Backend has 10 failures in `native-pane-owner-forward.test.ts` (`listen EINVAL` — the unix socket path exceeds the 104-byte `sun_path` limit in my worktree); they reproduce identically on clean `origin/main`, so they are pre-existing and unrelated to this PR.

PR #1572, opened 2026-08-27
Turn the agent usage popup into a modal that switches the default account

sworgkh:
dev3 now ships the `low-battery` answer rules itself, on by default, with one toggle to switch them off.

**Where the content comes from.** A build step (`scripts/generate-low-battery.ts`) shallow-clones `h0x91b/toolbelt-for-agents` and bakes `plugins/low-battery/` into `src/shared/low-battery-content.generated.ts` through a pure transform. It runs in `build`, `build:prod`, `package:win-archive` and in `devPlan`, so a release and `bun run dev` carry byte-identical content. Nothing is fetched on the user's machine. A build that cannot reach upstream exits non-zero rather than silently shipping without a feature advertised as on by default. The generated module is checked in, so tests and type-checking never need the network.

**What gets installed.** The skill goes into the same six agent config dirs dev3's own skills use; the Claude Code output style goes into `~/.claude/output-styles/`. Every harness except Claude Code gets an always-on line inside dev3's existing managed block in `~/.agents/AGENTS.md` — no second managed region.

**The output style is never taken from a user.** `applyLowBatteryOutputStyle` (pure, `src/shared/low-battery.ts`) allows three outcomes: key absent or `default` → dev3 selects its style; already any low-battery variant, including the plugin-namespaced `low-battery:Low Battery` → already on, nothing written; anything else → left alone, and the settings row names the style it kept and offers a one-click switch. Turning the toggle off removes only what dev3 wrote.

**Surfacing.** Settings → Agents → Low-battery answer format carries the toggle, the upstream revision, and the "your own style was left alone" state. `dev3 doctor` reports the same. A `feature-` changelog entry plus one clickable first-launch notice announce it.

Verified the style-name question against the installed Claude Code: the user-level loader keys a style by frontmatter `name` with **no** namespace prefix, so `Low Battery` is correct; the plugin loader is the one that prefixes. A test asserts the written name against dev3's own resolver applied to the shipped file.

Rationale and trade-offs: `decisions/2026/08/29/ship-low-battery-from-a-build-time-pull.md`.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=9ece8d22-040f-4a6f-abb1-d9550cb6ebae) · `dev3://task/9ece8d22-040f-4a6f-abb1-d9550cb6ebae`

### DIFF
--- a/change-logs/2026/08/26/feature-agent-usage-modal-default-account.md
+++ b/change-logs/2026/08/26/feature-agent-usage-modal-default-account.md
@@ -1,0 +1,3 @@
+Short: Switch default account from usage popup
+
+The header rate-limit pill now opens an Agent usage modal instead of a hover panel and a jump to Settings. It shows the same per-account quota cards for every Claude and Codex account, and each card is a radio row: click one to make it the default account for new launches, exactly as in Settings → Agent Accounts.
--- a/change-logs/2026/08/26/feature-agent-usage-modal-default-account.md
+++ b/change-logs/2026/08/26/feature-agent-usage-modal-default-account.md
@@ -1,3 +1,0 @@
-Short: Switch default account from usage panel
-
-The header rate-limit pill's panel now lets you switch the default agent account: it shows the same per-account quota cards for every Claude and Codex account, and each card is a radio row — click one to make it the default for new launches, exactly as in Settings → Agent Accounts. The panel still drops below the pill; a click pins it open so the rows can be used.
--- a/change-logs/2026/08/26/feature-agent-usage-modal-default-account.md
+++ b/change-logs/2026/08/26/feature-agent-usage-modal-default-account.md
@@ -1,3 +1,3 @@
-Short: Switch default account from usage popup
+Short: Switch default account from usage panel
 
-The header rate-limit pill now opens an Agent usage modal instead of a hover panel and a jump to Settings. It shows the same per-account quota cards for every Claude and Codex account, and each card is a radio row: click one to make it the default account for new launches, exactly as in Settings → Agent Accounts.
+The header rate-limit pill's panel now lets you switch the default agent account: it shows the same per-account quota cards for every Claude and Codex account, and each card is a radio row — click one to make it the default for new launches, exactly as in Settings → Agent Accounts. The panel still drops below the pill; a click pins it open so the rows can be used.
--- a/src/mainview/components/AgentUsagePanel.tsx
+++ b/src/mainview/components/AgentUsagePanel.tsx
@@ -5,8 +5,6 @@ import { isUnlimitedRateLimitSnapshot } from "../../shared/rate-limits";
 import { api } from "../rpc";
 import { toast } from "../toast";
 import { useT, type TFunction } from "../i18n";
-import { useEscapeKey } from "../hooks/useEscapeKey";
-import { useFocusTrap } from "../utils/useFocusTrap";
 import { notifyAgentAccountsChanged } from "./AgentAccountIndicator";
 import {
 	ACCOUNT_CARD_CLASS,
@@ -191,26 +189,24 @@ function UsageRowCard({
 }
 
 /**
- * Usage details for every agent account, as a modal instead of a hover panel.
- * It is the same quota cards the header pill used to show on hover, plus the
- * settings screen's radio semantics: clicking a card makes that account the
- * default for new launches (running sessions keep their login).
+ * Usage details for every agent account: the quota cards the header pill has
+ * always shown, plus the settings screen's radio semantics — clicking a card
+ * makes that account the default for new launches (running sessions keep their
+ * login). Rendered as the pill's own anchored flyout (or a BottomSheet on
+ * narrow), never as a centred dialog: the readout belongs next to the pill it
+ * explains, not on top of the whole board.
  */
-export default function AgentUsageModal({
+export default function AgentUsagePanel({
 	report,
 	accounts,
-	onClose,
 	onOpenSettings,
 }: {
 	report: AgentRateLimitsReport;
 	accounts: AgentAccountsState | null;
-	onClose: () => void;
 	onOpenSettings: () => void;
 }) {
 	const t = useT();
-	const trapRef = useFocusTrap<HTMLDivElement>();
 	const [busy, setBusy] = useState(false);
-	useEscapeKey(onClose);
 	const now = Date.now();
 
 	const setDefault = useCallback(async (kind: AgentAccountKind, accountId: string | null) => {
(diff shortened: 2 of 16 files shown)


src/mainview/components/RateLimitIndicator.tsx
```diff
@@ -15,26 +19,35 @@ import {
...
  * app-server account read; all other data comes from local files — see
  * rate-limit-monitor.ts.
  */
 function RateLimitIndicator({ compact = false }: { compact?: boolean }) {
 	const t = useT();
 	const [report, setReport] = useState<AgentRateLimitsReport | null>(null);
 	const [accounts, setAccounts] = useState<AgentAccountsState | null>(null);
+	const isNarrow = useNarrowViewport(CAROUSEL_MAX_WIDTH);
+	// Same open/pin/position machinery as the memory-headroom readout: hover drops
+	// the panel below the pill, a click pins it so its rows can be clicked.
```
h0x91b: **Blocker — the comment and the code disagree, and the code is the dangerous one.**

This says "a click pins it so its rows can be clicked", and the docblock above (`:37-38`) repeats it. But `useHeaderFlyout` with `variant: "bar"` calls `setOpen(true)` immediately on `onMouseEnter` (no dwell), and `panelProps.onMouseEnter` cancels the close timer — so once the pointer travels into the panel, every row is live without any pin ever happening.

I verified this in the running app rather than reasoning from the source: with the panel open by hover only, all six switchable rows reported `aria-disabled=null`. One stray click there moves the global default account.

Combined with the silent success path (see my note on `AgentUsagePanel.tsx`), the user can change a durable setting and never learn it. Compare the one mutation the manifest does sanction in this surface class (`ux-architecture.yaml:216`, memory reclaim): confirmation priced by blast radius, and "the row vanishing is the receipt".

Either make a row require the pinned state, or give the switch a receipt — but not neither.

src/mainview/components/AgentUsagePanel.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+	const setDefault = useCallback(async (kind: AgentAccountKind, accountId: string | null) => {
+		setBusy(true);
+		try {
+			// Same contract as Settings → Agent Accounts: the default is only the
+			// preselect for future launches, so no confirmation is needed.
+			await api.request.setActiveAgentAccount({ kind, accountId });
+			notifyAgentAccountsChanged();
+		} catch (err) {
+			toast.error(err instanceof Error ? err.message : String(err), { source: "settings" });
```
h0x91b, line 252: **Blocker (same one, other half): success is silent.**

Only the failure path talks. On success the panel just re-renders with the `Default` chip moved — and because hover-out closes the panel ~120 ms later, that re-render can be gone before the user's eye gets back to it.

A durable global change with no receipt is the part that makes an accidental click unrecoverable in practice (it stays wrong until someone notices weeks later that launches use the wrong account). A `toast.success` naming the account it switched to would close this without touching the interaction model.

src/mainview/components/AgentUsagePanel.tsx
```diff
@@ -0,0 +1,263 @@
...
+	t,
+}: {
+	row: UsageRow;
+	now: number;
+	busy: boolean;
+	onSetDefault: () => void;
+	t: TFunction;
+}) {
+	const inert = busy || !row.selectable || row.isDefault;
+	const name = row.account?.name ?? SOURCE_NAMES[row.kind] ?? row.kind;
```
h0x91b: The accessible name is built from `account.name` only, dropping `organization` and `planLabel` — which are exactly the fields that distinguish accounts sharing one email.

On a real install with several Claude accounts registered under one work email, this produces three rows that all announce the identical name **"Make <that same email> the default account"** (I read them off the live DOM). A screen-reader user cannot tell which one they are about to pick, on a control whose whole job is picking the right one.

The visible card already shows `email · organization` + plan chip — the label should carry the same.

src/mainview/components/AgentUsagePanel.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+	return (
+		<div className="p-3 space-y-3 text-xs">
+			<p className="text-fg-3 leading-snug">{t("rateLimits.panelSubtitle")}</p>
+			{blocks.map((block) => (
+				<div key={block.kind} className="space-y-1.5">
+					<div className="text-fg-2 font-semibold uppercase tracking-wider">
+						{SOURCE_NAMES[block.kind] ?? block.kind}
+					</div>
+					<div role="radiogroup" aria-label={SOURCE_NAMES[block.kind] ?? block.kind} className="space-y-1.5">
```
h0x91b: This is a `radiogroup` in name only: the rows are plain buttons with `role="radio"`, so there's no roving `tabindex` and no arrow-key navigation — the two things a radiogroup's ARIA contract promises.

There's a bigger keyboard problem behind it, and it isn't your doing but this PR is what makes it matter: `HeaderFlyoutPanel` portals to `document.body`, so Tab from the pill doesn't enter the panel at all. These are the first interactive controls ever placed in that flyout, and in natural tab order they're unreachable. (Escape → focus back to the trigger does work.)

If the panel keeps a mutating control, it needs focus moved into it on open, or the controls need a home the keyboard can reach.

src/mainview/components/rate-limit-ui.tsx
```diff
@@ -209,6 +220,24 @@ export function AccountCard({
 				<span className="text-fg-3">{t("rateLimits.credits", { balance: snap.creditsBalance })}</span>
 			)}
 			<CapturedNote capturedAt={snap.capturedAt} now={now} />
+		</>
+	);
+}
+
+/** One account's quota card: identity header + a bar per limit window. */
+export function AccountCard({
```
h0x91b: After this change `AccountCard` has **zero callers and zero tests** — `RateLimitIndicator` was its only consumer and now renders `AgentUsagePanel` instead.

`AGENTS.md` → "No deprecation — ever": the moment something is obsolete, delete it and rewrite every caller in the same change. There are no callers left to rewrite, so this is just a delete.

(`AccountCardHeader` + `AccountQuotaLines` stay, of course — that split is the good part of this PR.)

src/mainview/components/RateLimitIndicator.tsx
```diff
@@ -15,26 +19,35 @@ import {
...
 import { AGENT_ACCOUNTS_CHANGED_EVENT } from "./AgentAccountIndicator";
-import { AccountCard, SOURCE_NAMES, resolveAccount, severityFill } from "./rate-limit-ui";
+import { SOURCE_NAMES, severityFill } from "./rate-limit-ui";
+import AgentUsagePanel from "./AgentUsagePanel";
 
 /** The pill stacks one mini bar per account, capped to keep the header slim. */
 const MAX_PILL_BARS = 4;
 
+/** Panel width in px — wide enough for a "label · bar · % · reset" line. */
+const PANEL_WIDTH = 26 * 16;
```
h0x91b, line 29: Width is fine; height is where it hurts. Measured in the running app with 8 accounts registered (5 Claude, 3 Codex): the panel's content is **864 px inside a 446 px box** (`HeaderFlyoutPanel`'s `maxHeight: 28rem`), with 479 px of window still unused below it.

What ends up below the fold is the worst possible selection: the entire Codex group, the `settings.accountsNewSessionsHint` line ("the default is only a preselect for new launches — running sessions keep their current login") and the **Account settings** button. The one sentence that tells the user this click is safe is the one they can't see, on a panel where scrolling isn't signposted.

The cap is pre-existing, but this is the first content that blows through it — worth sizing to the viewport, or moving the hint above the groups.

change-logs/2026/08/26/feature-agent-usage-modal-default-account.md
```diff
@@ -0,0 +1,3 @@
+Short: Switch default account from usage panel
```
h0x91b: Two nits on this file:

1. **The date is the expected merge date**, not the day the work started (`change-logs/README.md`) — the changelog UI groups by ship day, so this wants moving to whichever day it lands on.
2. The slug still says `modal`, as do several test names and comments (`RateLimitIndicator.test.tsx:570`, `:588`, `:618`) and the PR description. Your second commit turned it into an anchored flyout — the body text here already describes it correctly, so it's just the naming that lagged.

h0x91b (changes requested):
Hey — Claude here, the AI assistant that reviewed this branch for Arseny. I checked the PR out locally rather than just reading the diff: `bun run lint` is clean, the three touched renderer suites pass (71 tests), and I drove the running app in headless Chromium at 1600×1000 with your panel open. Thank you for this one — the `AccountCardHeader` / `AccountQuotaLines` split is exactly the right refactor, reusing `useHeaderFlyout` instead of forking it was the correct instinct, and the informational-row handling (Codex's unmanaged login, readings from removed accounts) is genuinely thoughtful. The code is good.

I'm requesting changes on two things, and only one of them is about the code.

### 1. Placement — this needs a ruling before the code question matters

Switching the default agent account is *durable configuration*, and this repo's UX manifest is unusually explicit about where that class lives:

- `docs/ux/PRODUCT_UX_BIBLE.md:411` / `:588` / `:749` — configuration ("scripts, columns, labels, theme, locale, **gh account**") lives in Global or Project Settings; "durable behavior lives in settings".
- `docs/ux/ux-architecture.yaml:205` — `global_header.allowed` is readouts and destinations only; there is no configuration or mutation entry.
- `docs/ux/ux-architecture.yaml:216` — of the memory-reclaim action, verbatim: *"This is the only mutation allowed in an ambient-status surface."*

So as written, the manifest says **Settings → Agent Accounts** owns this action (with `AgentConfigPicker`'s per-launch picker as the local, non-mutating counterpart). `AGENTS.md` also makes `/ux-principal` mandatory for a change that adds an action to a surface, and `docs/ux` isn't updated here — `agent_rate_limit_indicator` still has no spec block of its own.

The product argument in your description is a good one (you pick an account *while* looking at how much quota each has left), and it may well win. But that's Arseny's call to make on the manifest, not something a PR can settle in passing. If he rules for it, the manifest needs the matching spec entry in the same change.

### 2. The switch fires on hover, silently

Detail is inline, but the short version: the panel opens on hover alone and its rows are already live — I read `aria-disabled=null` on all six switchable rows while merely hovering, never pinned. Success is silent (only errors toast), and the panel closes ~120 ms after the pointer leaves. So a stray click can move a global setting and the user may never find out. Reversibility isn't the problem; noticing is.

### Smaller things

- `AccountCard` (`rate-limit-ui.tsx:228`) now has zero callers and zero tests — the repo's "No deprecation — ever" rule wants it deleted in the same change.
- Two accessible names, a radiogroup without arrow keys, a panel that overflows at 8 accounts, and a couple of naming/date nits — all inline.
- The description still describes the modal design your own second commit reverted ("modal", `AgentUsageModal.test.tsx`, "Escape/backdrop close"). Worth rewriting so a reviewer reads the PR that exists.
- The branch is behind `main` (based on `bc36ca00a`, `main` is at `89b578ef1`) — a rebase plus `bun run test:full` before merge.

Nothing here is a rewrite. Once the placement question has an answer, the rest is an afternoon.

PR #1573, opened 2026-08-27
Make OSC 8 terminal hyperlinks clickable

arditti:
Markdown links that agents print in the terminal (e.g. Claude Code's `#3226` PR references, emitted as OSC 8 hyperlinks) now underline on hover, show their destination URL in a tooltip, and open in the browser on Cmd/Ctrl+Click, in both desktop and remote/browser mode.

Three independent breaks kept them dead; each has its own fix:

1. **CLIs never emitted OSC 8 inside tmux.** `supports-hyperlinks`-style detection (used by Claude Code and friends) gates on `TERM_PROGRAM` allowlists (`ghostty`, `iTerm.app`, `kitty`, …) that never include tmux. Fixed with `FORCE_HYPERLINK=1` in the tmux global session env (`src/bun/tmux/config.ts`) and in `AGENT_ENV_DEFAULTS`, spread below project env at every launch site, so Project Settings / `.dev3` env genuinely overrides it (pinned by a test that assembles the env the way `tmux-pty.ts` does).
2. **tmux stripped OSC 8 before it reached the renderer.** tmux only re-emits hyperlinks to clients whose terminal declares the `hyperlinks` terminal-feature; our attach client's `xterm-256color` never did. Added `terminal-features …:hyperlinks` for `xterm-256color`/`tmux-256color` (requires tmux 3.4+, set with `-gqa` so older servers ignore it). Verified byte-level with a raw pty capture of an attach client: without the feature the escapes vanish, with it they arrive intact.
3. **ghostty-web 0.4.0 cannot click OSC 8 links at all.** Cells carry `hyperlink_id` (hover underline works), but the wasm module exports no hyperlink accessor — `getHyperlinkUri()` is a stub returning `null`, so the built-in `OSC8LinkProvider` always yields zero links, and 0.4.0 is the latest release. New `src/mainview/terminal-osc8-links.ts` captures label→URI pairs from the raw PTY stream in `writeToTerminal` (escape-sequence state machine, chunk-split safe, LRU-bounded, dropped on RIS/CSI 3J) and registers a link provider that maps linked cell ranges (including wrapped rows, gated on `isWrapped` where the buffer knows it) back to their URIs. Activation goes through `window.open`, which the existing `new-window-open` intercept routes to `Utils.openExternal` on desktop.

Because terminal content is untrusted and label and destination are independent in the escape sequence, resolution is fail-closed and visible:

- A label ever seen with two different URIs is poisoned and resolves to nothing — a hostile line printed later can kill an earlier link, never silently retarget it. A URL-shaped label still resolves to its own visible text (what you see is where you go).
- A hover tooltip (`src/mainview/terminal-osc8-hover.ts`) shows each link's resolved destination before the click. ghostty-web never calls `ILink.hover`, so it is our own mousemove overlay.
- URIs are validated with `new URL()`: http(s) only (`mailto:` was a dead click on desktop — the intercept forwards nothing else), control characters and whitespace rejected, so `https://ok.dev\n@evil.example` no longer rides through a prefix check.

Details and trade-offs (same-label collisions, tmux redraw behavior, scrollback wrap-info limits, `FORCE_HYPERLINK` reaching piped output, what happens if ghostty-web ever implements the accessor) are in `decisions/2026/08/27/osc8-hyperlink-uri-from-stream.md`.

**Testing:** 47 unit tests for the tracker/provider/validator/tooltip (chunk splits, BEL/ST termination, SGR inside labels, wrapped-row stitching vs hyperlink-dedup false joins, label poisoning, RIS/3J forgetting, scheme and control-character filtering, tooltip show/hide/row math), plus tmux-config and agent-env precedence assertions; `bun run test:full` green. Manually verified in headless-browser QA: printed an OSC 8 link in a tmux pane — hover shows underline plus a tooltip with the exact URI, Ctrl+Click opened it; printed the same label again with a different URI — every occurrence of that label went dead instead of retargeting.


src/mainview/terminal-osc8-links.ts
```diff
@@ -0,0 +1,275 @@
...
+					let segment = "";
+					while (x < line.length && hyperlinkId(line, x) === id) segment += cellChar(line, x++);
+					const label = stitchedLabel(y, id, segment, x0 === 0, x === line.length);
+					const trimmed = label.trim();
+					if (!trimmed) continue;
+					// The stream tracker is the source of truth; a label that IS a
+					// URI still works when the tracker missed it (e.g. scrolled past
+					// the LRU) because the text itself is the destination.
+					const uri = opts.uriFor(trimmed) ?? (SAFE_URI.test(trimmed) ? trimmed : undefined);
+					if (!uri || !SAFE_URI.test(uri)) continue;
```
h0x91b: **Blocker — the destination is resolved by visible text, and never shown.**

`uriFor(trimmed)` looks the URI up by the label, so the mapping is global to the terminal and the newest OSC 8 with a given label wins for every occurrence of that label on screen. A hostile line printed later — a log entry, a `cat`'d fixture, output of a remote command — silently retargets a link a trusted agent printed earlier. Reproduced against this branch: the original link's own cells resolved to the attacker's URI.

The reason it can't be noticed is that no surface shows the target: `ILink.hover` is never implemented, so hover gives an underline and nothing else.

Two suggestions:
- Key on the hyperlink id instead of the label. The OSC 8 `id=` param is already parsed past in `handleOscPayload` — capturing it would let a link carry its own identity.
- Implement `hover` to show the resolved URI (a small overlay or the existing toast surface), so a text-vs-target mismatch is visible before the click rather than after.

src/mainview/terminal-osc8-links.ts
```diff
@@ -0,0 +1,275 @@
...
+ * (the second reads back as a blank), so the cell-side label carries spaces
+ * the stream-side label never saw.
+ */
+
+const MAX_LABEL_LENGTH = 2048;
+const MAX_OSC_PAYLOAD = 8192;
+/** How many neighbouring rows a wrapped link is stitched across, each way. */
+const MAX_STITCH_ROWS = 4;
+/** Schemes safe to hand to window.open / the desktop external-link intercept. */
+const SAFE_URI = /^(https?:\/\/|mailto:)/i;
```
h0x91b: Two things about this filter.

1. `mailto:` passes here, but the desktop shell's `new-window-open` intercept only accepts `http(s)` (`src/bun/window-manager.ts:269`). So a `mailto:` link is a dead click plus a `Blocked new-window-open with unexpected URL` warning on desktop, while it works in browser remote mode. Worth picking one: drop `mailto` here, or widen the intercept deliberately.

2. The check is a prefix match only, so control characters ride through. Measured on this branch: `https://ok.dev` + a raw `\n` + `@evil.example` passes both this filter and the desktop intercept regex, and browsers strip the newline while parsing — the navigation lands on `evil.example` with `ok.dev` demoted to userinfo. Parsing with `new URL()` and rejecting `[\x00-\x20]` would cover it.

src/mainview/terminal-osc8-links.ts
```diff
@@ -0,0 +1,275 @@
...
+			for (let down = 1; down <= MAX_STITCH_ROWS; down++) {
+				const line = opts.getLine(y + down);
+				const part = edgeSegmentText(line, id, true);
+				if (part === null) break;
+				text += part;
+				if (!line || hyperlinkId(line, line.length - 1) !== id) break;
+			}
+		}
+		return text;
+	}
```
h0x91b, line 300: Stitching decides a neighbouring row is a continuation purely from the hyperlink id on its edge cell — it never asks whether the row is soft-wrapped.

Since ghostty de-duplicates identical hyperlinks to one id (your decision record says exactly this), the same URL printed at the end of one row and again at the start of the next gets concatenated into a label that was never streamed. The lookup then misses and the link goes dead — which is worse than the wrapped case it fixes, because it silently breaks a link that used to work.

`isRowWrapped()` is exported by ghostty-web (`index.d.ts:361`) and already used in this repo at `src/bun/native-terminal-registry/ghostty-live.ts:368` — gating each stitch step on it should be a small change.

src/mainview/terminal-osc8-links.ts
```diff
@@ -0,0 +1,275 @@
...
+				uris.set(key, currentUri);
+				if (uris.size > maxEntries) {
+					const oldest = uris.keys().next().value;
+					if (oldest !== undefined) uris.delete(oldest);
+				}
+			}
+		}
+		currentUri = null;
+		label = "";
+	}
```
h0x91b, line 99: Nitpick: the map is never cleared when the screen or scrollback is (only on `dispose()`), so a label reused much later can resolve to a URI from output the user has long forgotten. Clearing on a screen-clear would also cap the worst case here — 500 entries × (2 KB label + 8 KB URI) is roughly 5 MB per terminal, reachable only with hostile input, but free to bound.

src/bun/rpc-handlers/shared-pure.ts
```diff
@@ -425,7 +425,10 @@ export function buildAgentEnv(extraEnv: Record<string, string>, taskId: string):
 	// `delimiter`, not ":" — Windows separates PATH entries with ";", and one wrong
 	// separator makes the whole variable unparsable for the agent we just launched.
 	const pathWithDev3 = currentPath.includes(dev3Bin) ? currentPath : `${dev3Bin}${delimiter}${currentPath}`;
-	return { ...extraEnv, DEV3_TASK_ID: taskId, PATH: pathWithDev3 };
+	// Agents gate OSC 8 hyperlink emission on supports-hyperlinks heuristics
+	// that recognize neither tmux nor the native backend; ghostty-web renders
+	// OSC 8 clickable, so opt them in (project env may still override).
+	return { FORCE_HYPERLINK: "1", ...extraEnv, DEV3_TASK_ID: taskId, PATH: pathWithDev3 };
```
h0x91b: Two separate things here.

**The comment's promise doesn't hold.** In `tmux-pty.ts:757-765` the env is assembled as `{ ...projectEnv, ...buildTaskLifecycleEnv(...), ...buildAgentEnv(extraEnv, task.id), ... }` — so this hardcoded `FORCE_HYPERLINK: "1"` is spread *after* `projectEnv` and overrides it. A `FORCE_HYPERLINK=0` set in Project Settings or `.dev3` config does nothing; only the per-agent-config `extraEnv` can win. Either move the default below `projectEnv`, or reword the comment (and the PR description) to say per-agent-config env.

**The variable reaches further than the interactive agent.** `supports-hyperlinks` checks `FORCE_HYPERLINK` in its *first* branch, above every colour and TTY check — I read the upstream source to confirm. So every child process now emits OSC 8 into piped, non-terminal output too: an agent's own captured command output, `dev3 pane run` log files, anything a script tees. Scoping it to the interactive process would avoid that; keeping it is also a defensible call, but then it's worth a line in the decision record so the next person isn't surprised by escape sequences in a log.

src/bun/rpc-handlers/__tests__/task-lifecycle-env.test.ts
```diff
@@ -55,3 +55,13 @@ describe("task lifecycle env — DEV3_TASK_SEQ", () => {
 		for (const leak of ["secret", "worktrees", "arsenyp", "feat/", "11111111"]) expect(name).not.toContain(leak);
 	});
 });
+
+describe("agent env — FORCE_HYPERLINK", () => {
+	it("opts agents into OSC 8 hyperlink emission", () => {
+		expect(buildAgentEnv({}, "t1").FORCE_HYPERLINK).toBe("1");
+	});
+
+	it("lets project env override the opt-in", () => {
```
h0x91b, line 74: This asserts the *agent-config* env can override the opt-in, not the project env — `extraEnv` is `resolved.extraEnv` at every call site, while Project Settings env arrives as a separate spread earlier in `tmux-pty.ts`. As written the name reads as proof of a guarantee the code doesn't provide. Renaming it, and adding a test that assembles the env the way `tmux-pty.ts` does, would pin the real precedence.

src/bun/tmux/config.ts
```diff
@@ -111,6 +111,15 @@ set -g focus-events on
...
 set -gqa terminal-features ",tmux-256color:Sync"
 
+# OSC 8 hyperlinks — ghostty-web renders them clickable, but tmux forwards
+# OSC 8 only to clients whose terminal declares the feature. Requires tmux 3.4+.
+set -gqa terminal-features ",xterm-256color:hyperlinks"
+set -gqa terminal-features ",tmux-256color:hyperlinks"
+
+# CLIs (Claude Code & friends) gate OSC 8 emission on supports-hyperlinks
+# heuristics that never recognize tmux; FORCE_HYPERLINK opts them in.
+set-environment -g FORCE_HYPERLINK 1
```
h0x91b, line 121: Both halves look right to me and the `-gqa` + `-q` shape matches the `Sync` lines above, so an older tmux ignores them instead of erroring — which is the behaviour that matters for installs updated in-app rather than through brew. I confirmed on tmux 3.6a that the lines are accepted and that `hyperlinks` is a documented terminal feature ("Supports OSC 8 hyperlinks").

One caveat rather than a change request: tmux does **not** validate feature names at set time (a bogus name is accepted silently, exit 0), so `show -g terminal-features` listing it proves nothing. My own pty captures came back empty in this environment, so I'm relying on your byte-level verification for the forwarding claim rather than having reproduced it.

Separately, `set-environment -g` also puts `FORCE_HYPERLINK` into every pane the user opens by hand, not just agent panes — see the note on `buildAgentEnv` for why that matters for piped output.

h0x91b (changes requested):
Hey — Claude here, the AI assistant doing the review pass on this branch (so you know where the comments are coming from).

First: thank you for this one. Chasing three independent breaks in a row — nothing emitting OSC 8, tmux dropping it, and then the terminal being unable to click it — is the kind of work most people abandon after the first dead end, and the decision record explains the reasoning honestly instead of hiding the heuristic. The premise checks out too: I read the shipped `ghostty-web@0.4.0` bundle and `getHyperlinkUri()` really is `return null`, so the built-in provider genuinely cannot produce a link. The unit tests are the good kind — chunk splits, BEL vs ST, SGR inside labels, wrapped rows.

I'm requesting changes for one thing, and it's a design consequence rather than a mistake in the code.

**The blocker.** The URI is resolved by the link's *visible text*, and nothing on screen ever shows the user where a link actually points. Driving the module on this branch:

```
# trusted agent output, stays on screen as hyperlink id 1
See <osc8 https://github.com/h0x91b/dev-3.0/pull/1573>#1573</osc8>

# any later line — a log, a test fixture, output of a remote command
log: <osc8 https://evil.example/phish>#1573</osc8>

# asking the provider about the ORIGINAL link's own cells:
row 0 resolves to: https://evil.example/phish
```

Terminal content is attacker-controlled in the ordinary case, and there is no tooltip, status line or hover text, so there is no tell. Two changes would close it: key the lookup on the hyperlink id rather than the label (recording the OSC 8 `id=` param when present), and surface the resolved target before activation — `ILink.hover` already exists in the interface and is currently unused.

Five smaller items are inline below, plus one note on a test that asserts something narrower than its name.

**What holds up:** no native dialogs (`window.open` is this repo's own path), no raw tmux spawn, no hardcoded user-facing strings, `~/.dev3.0` untouched, changelog entry with its `Short:` line, decision record with all five sections and an unnumbered slug, and `-gqa` + `-q` means an older tmux ignores the new lines rather than failing — which matters for in-app-updated installs.

**Coverage, so nothing reads as verified when it wasn't.** Checked here: `bun run lint` clean, the three touched test files green (39 tests), the tracker and provider driven directly, the `ghostty-web` bundle read, tracker throughput measured at ~2.2 ms per MB (replay cost is a non-issue). Read but not reproduced: tmux actually forwarding OSC 8 to an `xterm-256color` client — my pty captures came back empty, and tmux doesn't validate feature names at set time, so a green `show` proves nothing there; I'm taking your byte-level capture at face value. Not looked at at all: the app in a browser (I never built or served it, so your headless QA is the only evidence on the rendered surface), the native non-tmux backend, and tmux older than 3.4.

None of this is a big rework — mostly one hover handler and a few narrower checks. Thanks again for digging this deep.

src/mainview/terminal-osc8-links.ts
arditti: Fixed, though with different mechanics than suggested — neither suggestion is implementable against ghostty-web 0.4.0:

- **Keying on the OSC 8 `id=` param:** the cell side only exposes ghostty's internal `hyperlink_id`, which is page-managed and de-duplicated inside the wasm core; the stream-side `id=` param never comes back out, so there is nothing to correlate the two by (this is the "unreconstructible" alternative in the decision record).
- **`ILink.hover`:** the 0.4.0 bundle never invokes it — there is no call site, so implementing the interface member is dead code.

What ships instead:

1. **Fail-closed resolution.** A label ever seen with two different URIs is poisoned and resolves to nothing (`createOsc8Tracker`). A hostile later line can kill an earlier link, never retarget it; a URL-shaped label falls back to its own visible text, which cannot lie. Reproduced your scenario in headless QA — after the second `#9999` with a different URI, every occurrence of that label goes dead.
2. **Visible target.** `terminal-osc8-hover.ts` is our own mousemove overlay showing the resolved URI in a tooltip before the click, since the interface hook does not fire.

src/mainview/terminal-osc8-links.ts
arditti: Both fixed in `safeHttpUri()`: `mailto:` is dropped (dead click on desktop was the deciding factor — the intercept stays as is), and validation now parses with `new URL()`, accepts only `http:`/`https:`, and rejects `[\x00-\x20\x7f]` anywhere in the string. Your `https://ok.dev\n@evil.example` case is a unit test now.

src/mainview/terminal-osc8-links.ts
arditti: Gated on wrap info, with one caveat your suggestion runs into: ghostty-web hardcodes `isWrapped: false` for every scrollback row (the `IBufferLine` wrapper only asks `isRowWrapped` for screen rows), so a hard gate would silently break stitching for any wrapped link that scrolls out. The provider therefore takes `boolean | undefined`: screen rows gate strictly, scrollback rows report unknown (stitching allowed), and a stitched label that misses the map falls back to the row's own segment — which also resolves your dedup-concat case in scrollback. Both behaviors are unit-tested, and the map lookups run before the label-is-a-URI fallback so a false join of two URLs (which parses as one plausible URL) cannot win.

src/mainview/terminal-osc8-links.ts
arditti: The tracker now forgets its map on RIS (`ESC c`) and on erase-scrollback (`CSI 3J`); plain `CSI 2J` keeps it, since those links are still resolvable in scrollback. Covered by tests.

src/bun/rpc-handlers/shared-pure.ts
arditti: Fixed the precedence rather than the comment: `FORCE_HYPERLINK` moved out of `buildAgentEnv` into `AGENT_ENV_DEFAULTS`, spread first at all four launch-site assemblies in `tmux-pty.ts`, so Project Settings / `.dev3` env now genuinely overrides it (and per-agent-config env still outranks both). The piped-output reach is kept — scoping env to only the interactive process is not really possible once the shell tree inherits it — and is now documented in the decision record's Risks section as you asked.

src/bun/rpc-handlers/__tests__/task-lifecycle-env.test.ts
arditti: Replaced with tests that assemble the env in the same spread order as `tmux-pty.ts`: default present, project env overrides it, per-agent-config env overrides both, and `buildAgentEnv` itself no longer carries the variable (so its output can no longer outrank project env by accident).

src/bun/tmux/config.ts
arditti: No change to the tmux lines. The `set-environment -g` reach into hand-opened panes is now covered by the decision record's Risks note on piped output, and the variable is genuinely overridable via project env after the precedence fix on the `buildAgentEnv` thread.

h0x91b (approved):
Claude again (the AI assistant reviewing this branch) — re-checked `c1f6682b0`, and every item from the previous pass is genuinely closed. Nice work, especially picking fail-closed over "most recent wins".

What I re-ran against your code rather than taking on trust:

| Item | Result |
|---|---|
| Label retargeting (the blocker) | A hostile line re-using a label now yields **no link** instead of the attacker's URI, and an uncollided link still resolves. Same label with the *same* URI twice does not poison itself |
| Hover tooltip | `linkAt` returns exactly the URI the click would open, and `undefined` off the label |
| Scheme / URI validation | `javascript:`, `data:`, `file:`, `vscode://`, `mailto:` and leading whitespace all rejected; the `https://ok.dev\n@evil.example` case is now refused; uppercase `HTTPS://` still accepted, which is right |
| Row stitching | Two separate links sharing an id on unwrapped rows no longer fuse; a genuinely wrapped label still stitches |
| Scrollback erase | `CSI 3J` drops the map |
| Env precedence | `AGENT_ENV_DEFAULTS` is spread first at all four assembly sites — I grepped rather than eyeballed — and the new test asserting `buildAgentEnv` does *not* carry the variable is a good guard against it drifting back |

Also: `bun run lint` clean, 79 tests green across the five touched test files locally, all four required checks green on this head, the decision record superseded in place (including the piped-output trade-off spelled out as accepted), and the changelog entry updated rather than duplicated. Every ghostty-web API the tooltip touches exists in the 0.4.0 typings.

One honest gap, unchanged from last time: I never ran the app, so the tooltip's actual appearance and placement are covered by your QA and by unit tests, not by my eyes.

Thanks for the quick and thorough turnaround — squash-merging now.

PR #1576, opened 2026-08-27
Fix Web Push packaging and two remote-access gaps

mcaldas:
Web Push has never worked in a single packaged build. Found during an end-to-end setup of phone access (Cloudflare tunnel + Access + Web Push) against installed 1.48.1, along with two smaller remote-access gaps.

> Revised after review: the static-code change is dropped — see the note at the end.

## 1. Web Push was dead in every packaged build

`src/mainview/public/sw.js` and `manifest.webmanifest` were not in `electrobun.config.ts`'s `build.copy` map, which is an **allow-list** — Vite emits them into `dist/`, but nothing copied them into `views/mainview/`. The remote server's SPA fallback then answered `/sw.js` with `index.html`, and a service worker served as `text/html` cannot register, so push enrolment could never start. On iPhone this surfaced as Safari's generic "could not update push notifications".

The same class of bug shipped once before: the favicons were missing for the same reason and were added by hand in an unrelated PR (42001e0a).

- `publicAssetCopyEntries()` enumerates `src/mainview/public/` from disk and spreads the entries into `copy`; the four hand-written favicon lines are gone.
- `src/bun/__tests__/electrobun-config.test.ts` asserts every file in that directory has an entry, so the next asset dropped in fails the suite rather than a phone. An `index.html`-based test would not have caught `sw.js` — the service worker is registered from JavaScript and never referenced by the HTML.
- The config is also bundled into the app (`src/bun/index.ts` imports it) where `src/` does not exist, so the scan is `existsSync`-guarded — the first version of this crashed the app on boot with ENOENT. The directory is injectable so that guard has its own test.
- `serveStatic` learned the `.webmanifest` MIME type; it was falling through to `application/octet-stream`.

**Verified against a real packaged build**, not just the config: the dev bundle's `views/mainview/` now contains both files, `/sw.js` serves as `application/javascript`, `/manifest.webmanifest` as `application/manifest+json`, and `navigator.serviceWorker.register('/sw.js')` resolves.

## 2. `DEV3_REMOTE_PORT` could not be set for the GUI app

`resolveListenPort` reads the env and *is* used on the GUI path, but `SHELL_ENV_DENIED_PREFIXES` strips every `DEV3_*` key out of the imported shell environment, so `.zshrc`, `.zshenv` and `launchctl setenv` all silently do nothing. Anything hard-coding dev3's port — a tunnel's dashboard ingress, a reverse proxy, a `tailscale serve` entry — broke on every launch.

New `remotePort` global setting, surfaced in Settings → System beside the tunnel provider; the env still wins where it is set.

## 3. Custom tunnel log lines were attributed to the wrapper

`cloudflared tunnel run` never prints its own hostname, so a named-tunnel command must echo the URL itself — `echo https://host; exec cloudflared tunnel run …`. Taking the first word attributed cloudflared's output to `echo`. `tunnelCommandName` now prefers the last exec'd binary; with no `exec` it takes the last command of a sequence (`echo url; ngrok …` → `ngrok`) but the **first** of a pipeline (`ngrok … | tee log` → `ngrok`, the producer rather than the sink), and it no longer mistakes a leading `VAR=value` for the binary.

---

## Review follow-up

- **Static-code change dropped in full**, as requested — the `getStaticCode()` guard, the inherited-env CLI extension, its decision record, both test blocks, the docs paragraph and the changelog sentence are all gone. The point that a static code is a deliberately long-lived, multi-use credential is well taken, and #1585 covers the guessability angle properly.
- **`RemotePortField` freezing its initial value was a real bug** — confirmed and fixed. It now adopts the stored value whenever it arrives, and the new test renders *without* `remotePort` first and supplies it on a second render, so it fails against the old code (verified) rather than passing on the case that already worked.
- **Rebased** on `main` (14 commits, not 3 by the time I got to it). `bun run test:full` is green end-to-end — 12,295 tests, no caveat: #1577 did fix the `FolderPickerModal` failure.
- Nits: the asset scan is `.sort()`ed, and the `existsSync` guard has its own test.

Decision record: `decisions/2026/08/27/public-assets-generated-into-the-copy-map.md`.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=c5bc3344-1359-441d-b7f4-3036e3ca869d) · `dev3://task/c5bc3344-1359-441d-b7f4-3036e3ca869d`

https://claude.ai/code/session_01CSbVqsXDYvyu9ECZaDx1FS


h0x91b (changes requested):
Hi Mathias — this is Claude, the AI assistant working with Arseny on dev-3.0. Thanks for this one, genuinely: the packaging fix is the standout. Web Push had never worked in a single packaged build since it shipped, and it took a full phone setup to even see it. That's a hard bug to find and an easy one to keep missing, so the report and the fix are both appreciated.

We'd like to take most of this. Three things before we merge:

### 1. Please drop the static-code change (§2) entirely

This one is a design difference rather than a defect in your code. A static access code is intentionally a long-lived, multi-use credential: the whole point is that several devices and browsers can enrol without scanning a fresh QR from the host each time. Reuse is the feature, not the weakness — so voiding the code while a tunnel is up disables it precisely in the multi-device remote case it exists to serve.

We do agree there's a real risk here, but we think it's *guessability*, not reuse. The minimum length is currently 4 characters (`src/cli/commands/remote.ts:193`) and `/auth/exchange` has no rate limiting at all, which is the combination that actually matters. We've opened a separate task to raise the minimum length and add attempt throttling, so that ground is covered — it just shouldn't ride along in this PR.

Concretely, please remove: the guard in `getStaticCode()`, the inherited-env extension to the CLI gate, `decisions/2026/08/27/static-access-code-refused-while-a-tunnel-is-up.md`, the `getStaticCode with a public tunnel` tests, the inherited-env CLI test, the static-code paragraph in `docs/remote-access.md`, and that sentence from the changelog entry.

### 2. `RemotePortField` freezes its initial value — `SystemSettingsSection.tsx:631`

`useState(port === undefined ? "" : String(port))` runs once at mount, but settings load asynchronously afterwards (`GlobalSettings.tsx:249`). On any route that opens straight into Settings → System — a deep link, the menu, or a page reload in remote mode — the field renders before the value arrives and shows the "Automatic" placeholder while a port is actually pinned. Interacting with it then persists `undefined` and silently unpins the port.

Worth noting why this is easy to miss: the neighbouring `CustomTunnelFields` uses the identical pattern and is fine, but only because it renders conditionally and so can't mount before the settings land. A test that supplies `remotePort` *after* the first render would lock this down — the current one passes it at mount, which is the case that already works.

### 3. Please rebase on `main`

You're 3 commits behind, and the `FolderPickerModal` failure you flagged as pre-existing has since been fixed upstream in #1577 — so after a rebase that run should be clean and needs no caveat.

### Smaller, entirely optional

- `readdirSync` in `publicAssetCopyEntries()` isn't sorted, so the map key order follows the filesystem.
- The `existsSync` guard — the line that prevents the boot crash you described — has no test of its own.
- `tunnelCommandName` now returns `custom-tunnel` for a piped command like `ngrok http {port} | tee log`, which used to resolve to `ngrok`. Cosmetic, but it's a behaviour change that isn't mentioned in the description.

---

Two things we especially liked, for the record: asserting `public/` ⊆ the copy map instead of parsing `index.html` links (you rejected the obvious version yourself for exactly the right reason — `sw.js` never appears in the HTML), and bucketing the new telemetry to `auto`/`pinned` rather than sending the port. Both are better than what we'd likely have written.

h0x91b:
Claude again (the AI assistant working with Arseny) — following up on my review, because §2 deserves a much better answer than "please drop it".

**First, and I want to be clear about this: your finding here was genuinely valuable.** Arseny went through the whole area after reading it, and it turned up a chain of problems nobody had looked at. The gap you spotted is real, and the discussion it started is going to fix considerably more than the line you touched. Asking you to remove the change is not a verdict on the work — it is that the investigation you triggered changed our mind about what the right shape is.

**He's taking this part himself in a separate PR**, so please just drop §2 from yours; the rest of the review stands as written.

Since you'll want to know where it landed, here is the full picture.

### What we got wrong in our own code, and what you inherited from it

The framing "a static code has no replay protection, therefore it must not ride a tunnel" came from our own comments and docs, and it's the wrong frame. **A static access code is permanent and multi-use by design.** The owner sets one long code, and anyone holding it signs in as often as they like, from any browser and any machine. That's not a weakness in the mechanism — it's the entire point of having it, and it's exactly what you need when several devices reach the same host over a tunnel and there's no one at the host screen to produce a fresh QR.

So voiding the code while a tunnel is up disables it precisely in the scenario it exists for. Your patch is a faithful implementation of what our docs claimed the feature was; the docs were wrong.

### What the actual problems are

Reuse isn't the risk. Two other things are:

1. **It can be guessed.** The minimum length is 4 characters (`src/cli/commands/remote.ts:193`) and `/auth/exchange` has no rate limiting at all — no attempt counter, no delay, no 429. `checkOrigin` isn't a defence either, since it returns `true` when a request carries no `Origin` header, which a non-browser client simply omits. That combination is the real hole, and it's the one your change leaves untouched.
2. **It leaks through the URL.** `getAccessUrl` puts the code into `?token=<code>`. Fine for the rotating token, which burns on use — bad for a permanent credential, which then sits in browser history, the address bar, and any proxy or referrer log.

### Where it's going

Two tasks, both already running:

- **Hardening:** raise the minimum code length to at least 8 characters, and rate-limit `/auth/exchange` so guessing is impractical without letting an attacker lock the owner out.
- **Redesign:** take the code out of the URL entirely and put a sign-in field on the browser side instead; let the QR path and the static code work side by side rather than the code disabling the JWT exchange; turn the CLI's hard refusal into a warning that still starts the server, and make it behave identically whether the code arrived by flag or by inherited environment (right now those two disagree, which is the inconsistency underneath what you found); promote the code from an environment variable to a real setting in Settings → System, so the desktop app can have one at all; and surface its state in the UI instead of only in a log line.

One deliberate non-change, so it doesn't read as an oversight later: changing the code will **not** invalidate existing sessions.

### On the rest of your PR

Unchanged from the review: the packaging fix is the valuable part and we want it. Web Push had never worked in a single packaged build, and the way you tested it — asserting `public/` ⊆ the copy map rather than parsing `index.html` links — catches the next asset too. That reasoning is now quoted in our own notes.

Thanks again. A report that makes the maintainer re-examine an entire subsystem is worth a lot more than a patch that merges quietly.

mcaldas:
Thanks — all three applied, and the branch is rebased and green.

**§2 is gone in full:** the `getStaticCode()` guard, the inherited-env CLI extension, the decision record, both test blocks, the docs paragraph and the changelog sentence. `src/cli/commands/remote.ts` and its test file are byte-identical to the base again.

Your reasoning is right and worth stating plainly: I built faithfully to what the code's own comment said the feature was ("no replay protection"), and I never questioned whether that description was correct. A permanent multi-use credential is a different thing from a token that failed to rotate, and voiding it on a live tunnel disables it in exactly the multi-device case it exists for. Guessability being the real risk — a 4-character minimum against an endpoint with no throttling, and `checkOrigin` returning true when the header is simply absent — is a sharper read than mine, and #1585 is the right home for it.

**`RemotePortField` was a real bug, not a style point.** `getGlobalSettings()` resolves in a `useEffect` after the section mounts and the section renders unconditionally, so the field mounted with `port === undefined`, showed "Automatic" over a pinned port, and persisted `undefined` on blur. It now adopts the stored value whenever it arrives. Your warning about the test was the useful part: mine passed the value at mount, which only exercises the case that already worked. The new one renders without `remotePort` and supplies it on a second render — I deleted the fix to confirm it actually fails (`expected '' to be '41234'`) rather than trusting that it would.

**Rebased** — 14 commits by the time I got to it. `bun run test:full` is green end-to-end, 12,295 tests: #1577 did fix the `FolderPickerModal` run, so the caveat is out of the description.

**On the nits:** the scan is sorted, and the `existsSync` guard has its own test — I made the directory an injectable parameter to get it, because my first attempt used `process.chdir` and proved nothing (the path resolves from `import.meta.url`, so it would have passed regardless).

One place I went further than you asked. You suggested documenting the `tunnelCommandName` pipeline change; it was three lines to make it correct instead, so with no `exec` it now takes the last command of a sequence (`echo url; ngrok …` → `ngrok`) but the **first** of a pipeline (`ngrok … | tee log` → `ngrok`, the producer rather than the sink), and no longer mistakes a leading `VAR=value` for the binary. It is described in the PR body and covered by tests — happy to reduce it back to a plain doc note if you would rather the behaviour change stayed out of this PR.

h0x91b:
Claude here again (the AI assistant working with Arseny) — approving, with one heads-up you should read before you pull.

**We rewrote your branch.** You had been waiting on us for a day, `main` had moved under you twice, and the remaining work was mechanical — so rather than send you round another loop, Arseny asked me to finish it on your fork directly. Your two commits are now one, rebased onto current `main`, and the branch was force-pushed. If you have local work on it, reset to the remote rather than merging.

Here is exactly what changed, so nothing arrives as a surprise:

**1. Rebased over the static-access-code work (#1586).** That PR added a "Remote access code" field to the same System settings section your port field lives in, which is what put your branch into conflict. Seven conflict hunks, all of them "both sides added something here" — the two fields now sit side by side, in the same order in `settings-registry.ts` and in the rendered section. Nothing of yours was dropped.

**2. Took you up on your offer about `tunnelCommandName`.** You wrote that you were happy to reduce it back to a doc note. We kept a bit more than that: the `exec` branch stays, because the bug behind it is real — `cloudflared tunnel run` cannot print its own hostname, so the wrapper shape is forced on the user and `echo` was genuinely the wrong label. What we removed is the sequence/pipeline heuristic, because it regressed on redirection. These three commands returned `"1"` instead of the tool name:

```
ngrok http {port} --log stdout 2>&1
ngrok http {port} > out.log 2>&1
cloudflared tunnel --url http://localhost:{port} 2>&1 | tee tunnel.log
```

The `&` inside `2>&1` was read as a command separator, so `1` became the last segment and then the label. `main` gets these right by taking the first word, so it was a regression rather than existing messiness — and your four new cases could not catch it because none of them contained a redirection. Matching `exec` as a whole word and splitting nothing else fixes all three while keeping your `echo`/`exec` case working. There is a new test covering it.

**3. Small things.** The changelog entry moved to the merge date. Two doc comments that said the desktop app "never sees" `DEV3_REMOTE_PORT` now say a *Finder launch* never sees it — the prefix strip only filters the imported login-shell environment, so a launch from a terminal that exported the variable does see it. Your setting is still exactly as necessary; the sentence was just wider than the truth, and it was our sentence originally, not yours.

**4. The one thing that is not yours and did not block you.** A pinned port that turns out to be busy used to take down half of app startup silently. That is our code, it predates your PR, and your setting simply made it reachable for the first time. It is fixed on `main` in #1590, which landed while this was being rebased.

The commit keeps you as author; Arseny is the committer, and the commit body spells out the maintainer edits.

Thank you — genuinely. Web Push had never once worked in a packaged build, and finding that took a real phone and real patience. The way you tested it is the part we liked most: asserting that `public/` is a subset of the copy map instead of parsing `index.html` for links, because you noticed yourself that `sw.js` never appears in the HTML at all. That reasoning is now quoted in our own notes. You also took a firm review note about the access code without a word of argument and came back with a better patch. Please do send more.

PR #1595, opened 2026-08-29
Select terminal text on a phone the way the phone does it

mcaldas:
## The report

> "Review the mobile version of dev3 and see if it would be possible to select the text (iOS mostly), right now it selects the full terminal, I wanna be able to select specific text"

On a phone there was no way to copy a specific piece of terminal output — a path, an error, a URL. Long-pressing highlighted the whole terminal, and Copy produced nothing useful.

## Why it happened

ghostty-web marks the terminal container `contenteditable="true"`. On iOS, a long press inside an editable block with no text in it — the glyphs live in a `<canvas>` — makes WebKit select **the entire block**. Copy then puts a `.webarchive` of the page on the pasteboard. The reporter's own clipboard, mid-investigation, held 5.2 MB of exactly that.

The app's only real selection trigger was a horizontal drag in raw mode, which is also the pane-swipe gesture — and compose mode, the default on mobile, disabled it outright.

## What was tried first, and why it wasn't enough

The first pass replaced the horizontal drag with a 450 ms long press that anchored a ghostty/tmux selection and copied on release. It worked end to end — the dragged range landed in the tmux paste buffer, verified.

It was still wrong. It produced no grab handles, no magnifier and no Copy menu, because none of that is ours to draw: **iOS renders its selection UI only for genuine text nodes, never over a canvas.** The mechanism was right and the affordance — the half a user actually recognises — was missing. That commit is kept in the history rather than squashed away, because the reasoning is the interesting part.

## What this does instead

Stop implementing selection. Give the platform something real to select.

`src/mainview/terminal-touch-text-layer.ts` lays a **transparent copy of the visible rows over the canvas**. iOS and Android then run their own selection on it — handles, magnifier, Copy / Look Up / Share — and the copy is plain text straight out of the DOM, with no OSC 52 round trip and no insecure-context clipboard problem.

Three things carry the alignment, and all three are load-bearing:

| | |
|---|---|
| **Row indexing** | `buffer.active.getLine(y)` indexes the **whole** buffer, scrollback included — screen row 0 is `wasmTerm.getScrollbackLength()`, not 0. The layer resolves a row the way ghostty resolves a click: `scrollbackLength - max(0, floor(viewportY)) + row`. Reading from 0 laid the session's oldest output over the live screen, pixel-perfect and completely wrong |
| **Columns** | `lineToText` (already in `terminal-file-links.ts`) emits one UTF-16 unit per cell, so string index === screen column |
| **Advance** | `letter-spacing` is retuned every refresh to `charWidth - measuredAdvance`, so one character advances by exactly one cell whatever the font measures |

The layer refreshes from the render guard's `onFrame` — ghostty-web never fires `onRender` — coalesced to one rebuild per animation frame, and skips entirely while a selection is live so incoming output cannot collapse it.

Because the layer is on top, it (not the canvas) is what a finger lands on, so the scroll/tap gestures in `TerminalView.tsx` listen there and stand down whenever it carries `data-selecting`. `MobilePaneCarousel` reads the same attribute, so dragging a selection handle sideways never swipes panes — and its nascent-selection collapse hack goes away with the trigger it existed for.

## Who gets it

Gated on `matchMedia("(pointer: coarse)")`, **not** `navigator.maxTouchPoints > 0`. That distinction is a bug this PR already fixed once: the layer is what a pointer lands on, so gating on mere touch support would install it on a touchscreen laptop driving a mouse and take every terminal click away from ghostty.

## Verification

Driven in a headless browser against a real dev-server, at a 402×874 iPhone viewport:

- **Layer geometry** — flush over the canvas, `0px` delta on all four edges; columns land at exactly `6.0px` pitch (col 10 → 60.27px, col 40 → 240.27px).
- **Alignment, visually** — tinting the layer red over the live canvas shows every glyph crisp, no doubling. This is how the scrollback-indexing bug was found; it is invisible any other way.
- **Desktop not broken** — with a fine pointer, no layer is installed and a mouse drag still reaches tmux: 167 bytes of exactly the dragged range in the paste buffer.
- **The earlier gesture, before it was replaced** — long-press + drag put the exact 164-byte range in the tmux buffer, confirming the copy path end to end.
- **Suites** — `bun run test:full` green after the rebase: 5010 renderer, 7535 backend, 985 CLI.

New tests cover the layer's content, scrollback indexing, viewport scrolling, geometry, live font changes, selection-safety on refresh, the `data-selecting` flag, disposal, and the gesture stand-down — plus the fine-pointer guard.

## What is NOT verified here

**The iOS chrome itself.** Headless Chromium renders no handles and no callout, and Playwright's device emulation still reports `pointer: fine`, so both the selection UI and the gate that admits it were exercised by forcing the media query in both directions. **It was confirmed working by hand on a real iPhone** before this PR was opened.

## Known limits, deliberate for now

- Only the visible viewport is in the DOM, so a selection cannot run off-screen into scrollback — scroll first, then press.
- The layer carries logical text, so a bidi viewport selects in logical rather than visual order.
- Alignment assumes the font measures the same for `measureText` as for ghostty's `charWidth`; a fallback font that does not would drift across a wide row.

## Notes for review

- `user-select: none` on the terminal container applies to **every** browser session, desktop included. It only suppresses DOM selection — ghostty draws its own on the canvas — which is why the desktop drag was driven end to end rather than assumed.
- Font size and family are both live settings since `main` moved under this branch; the layer re-reads them from `term.options` on every refresh rather than capturing them at install. That seam was found by the rebase and is its own commit.
- Full rationale, alternatives and risks: `decisions/2026/08/29/native-touch-selection-over-a-text-layer.md`.

https://claude.ai/code/session_01E741gykZSdmumF7vqujPEs

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=40b9c7f7-84b2-4c62-b273-fa1acef0f44d) · `dev3://task/40b9c7f7-84b2-4c62-b273-fa1acef0f44d`


h0x91b:
Hey @mcaldas — Claude here, the AI assistant working with Arseny on this.

Thank you for this one, genuinely. The write-up did a lot of the reviewing for me, and keeping the earlier long-press commit in history instead of squashing it made the reasoning easy to follow.

The hold-up is on the device side rather than in the code: Arseny tried the branch on Android and it behaves badly there — the selection doesn't hold together in practice, so it isn't really usable yet. We'd rather not merge it in that state, and it isn't obvious yet what the fix looks like, so let's keep the PR open rather than land something that only works on one platform.

Not a "start over" — the approach looks right to us, it's the Android side that needs to catch up. Happy to share more detail from the review if that would be useful.

mcaldas:
Thanks — and yes please, more detail would help a lot. I don't have an Android device here, so I'm working from hypotheses and I'd rather not guess-and-push at your test device.

**What would unblock fastest**

A screen recording of the long press + drag is worth more than any description, but even one line each on these would narrow it a lot:

1. **Which browser and version**, and Android version. Chrome, Samsung Internet and Firefox each drive selection differently, and Samsung Internet in particular diverges from Blink defaults.
2. **Where it breaks.** These fail in very different places, and the fix is different for each:
   - long press selects **nothing** at all
   - it selects, but the **handles can't be dragged** (they appear, then the drag does nothing or the selection snaps back)
   - it selects, but the selection **disappears on its own** a moment later
   - it selects, but the **terminal scrolls** under it while dragging
   - it selects, but dragging sideways **switches pane** instead of extending
   - it selects fine and **Copy gives the wrong text** (or nothing)
3. **Idle vs busy terminal.** Does it misbehave the same on a pane printing nothing, or only while output is flowing? This one is diagnostic on its own — see hypothesis 3 below.
4. Does the **on-screen keyboard** pop up during any of this?

**My three suspects, in order** — if any of them matches what Arseny saw, say which and I can go straight at it:

1. **`touch-action: none` on the terminal container** (`TerminalView.tsx`, set so iOS doesn't rubber-band the page and kill our synthesized scroll). Blink is stricter than WebKit about this: it can leave the long press working while breaking the *handle drag*, because the drag is a touch gesture the page has declared it owns. Symptom would be #2 above. Fix is to relax it to `pan-y`, or drop it to `auto` while a selection is live.

2. **A race between the OS starting the selection and us noticing.** `data-selecting` is set from a `selectionchange` listener, but our `touchmove` handler reads that attribute synchronously. Blink dispatches `selectionchange` on a task, so there's a window where the OS is dragging a handle and we still think we own the gesture — we synthesize wheel scroll, and `MobilePaneCarousel` may `preventDefault()` a horizontal move as a pane swipe, which would cancel the drag outright. Symptoms #4 and #5. This one I can fix blind and probably should regardless: read `document.getSelection()` directly instead of the attribute, so the guard is synchronous and cannot lag.

3. **The layer rebuilding under a live selection.** `refresh()` skips while a selection is live, but it decides that with `element.contains(range.commonAncestorContainer)`. If Blink anchors part of the range outside the layer, that check says "not live", the row text nodes get replaced, and the selection dies — which would look exactly like "it doesn't hold together", and would happen *only on a terminal that is printing*. Hence question 3 above.

If it's #1 I need a device to confirm the fix, and I'd lean on you for a single retest. #2 and #3 I can fix and cover with tests here.

One more thing worth stating plainly: I verified the iOS side by hand on a real iPhone, but the platform's selection UI can't be reproduced headlessly at all — so whatever I do for Android will need a pass on real hardware before it's worth trusting. Happy to keep it open as long as that takes.

mcaldas:
Pushed one fix while waiting — hypothesis 2 turned out to be a real bug, not just a theory, so it's worth having regardless of what Arseny saw.

**Both gesture guards were reading a stale flag.** `data-selecting` was maintained by a `selectionchange` listener, and both `TerminalView`'s touch handler and `MobilePaneCarousel` read that attribute. Blink dispatches `selectionchange` on a task, so for the frames right after the OS starts a selection we still believed we owned the gesture: the touchmove handler synthesized a wheel scroll, and the carousel could `preventDefault()` a horizontal move as a pane swipe — dragging the terminal out from under the handles, or cancelling the drag outright.

A test that mounts a live selection *without* delivering the event reproduces it: a `deltaY: 40` wheel fires straight into the canvas. Both guards now ask `document.getSelection()` synchronously; the attribute and its listener are gone.

**Hypothesis 3 is folded into the same commit.** The range test widened from `contains` to `intersectsNode` — a range Blink anchored at a parent still covers our rows, and reading that as "no selection" let a refresh replace the text nodes under a held selection. On a printing terminal that happens within a frame, which is why I asked whether idle vs busy differs.

Neither is *confirmed* to be what you hit — I'm fixing blind, and I'd rather say so than imply it's solved.

**Still untouched, deliberately: `touch-action: none` on the terminal container.** That's my leading suspect for the handle drag specifically, but it's also load-bearing on iOS (it's what stops Safari rubber-banding the page and cancelling our synthesized scroll), and I can't tell whether relaxing it regresses iOS without a device. If Arseny's symptom is "handles appear but dragging does nothing / snaps back", that points straight at it and I'll do it next — probably by flipping the container to `touch-action: auto` while a selection is live, since Blink latches the value at touchstart and the handle grab is a fresh touch sequence.

`bun run test:full` green, rebased on current main. Take your time — a retest on the device is worth more than me guessing again.

mcaldas:
I got an Android emulator running here, so I can stop guessing — **and my leading hypothesis was wrong.** Chrome 124, Android 15, arm64, real Chrome for Android driven through `adb shell input swipe` so the touches go through the real input pipeline, with CDP attached to read the selection back.

**What actually works on Android, today, on this branch:**

- long press selects the **word** under the finger (`"SwiftShader"`, 11 chars)
- Android's **grab handles render** and sit correctly on the glyphs
- **dragging a handle extends the selection exactly** — `"SwiftShader"` → `"SwiftShader/src/Vulkan/"`, 23 chars
- no pane swipe on a horizontal handle drag, no scroll under the selection

So `touch-action: none` is **not** the problem, and I was wrong to lead with it. It stays as it is.

**What's actually broken:** Chrome never shows the floating action bar — no `Copy`, no `Select all`, no `Share`. You get a perfect selection and no way to copy it. That is almost certainly what "doesn't hold together in practice" looked like from the outside.

**Isolated with a control.** Same emulator, same Chrome, a plain `<pre>` page: long press → `Copy / Share / Select all / Web search` appears immediately. So it's our page, not the browser and not the emulator.

**Cause: the `contenteditable="true"` that ghostty-web puts on the terminal container.** With it, *no* selection-context UI appears at all. Setting `contenteditable="false"` on that container makes Chrome's selection UI come back — reproduced twice, at different screen positions. It's the same attribute that caused the original iOS bug (WebKit selecting the whole editable block), which is a bit poetic.

**What I haven't done yet, deliberately:** actually dropping it. The container is ghostty's focus target, and I want to confirm that removing `contenteditable` in browser mode doesn't break focus, IME or the raw-mode keyboard path before I push it. Typing goes through the hidden textarea there and `inputmode="none"` is already on the container, so I think it's safe — but "I think" isn't good enough for the thing that broke this on iOS in the first place.

Two questions, now that I can test rather than ask:

1. Does the missing Copy bar match what Arseny saw? If he saw something *else* as well, I still want to know — I can reproduce here now.
2. Which Android browser was it? I've only got Chrome in this emulator; Samsung Internet diverges and I can't test it.

I'll follow up with the fix once I've verified the input paths.

h0x91b:
Now the selection itself works fine, but the `copy` button doesn't appear.. So I still can't copy text... maybe auto-copy on each change selection?


https://github.com/user-attachments/assets/f267522f-6fb5-4faa-a3a5-a11a05de30af

mcaldas:
Your read matches mine exactly, and auto-copy is the right call — pushed.

I'd reached the same place independently on the emulator: selection is fine, Chrome never draws the floating toolbar over these rows, so you hold a perfect selection with nothing to do with it. A plain `<pre>` in the same browser gets the toolbar instantly, so it's this page and not the browser.

**One correction on the way there.** I eliminated four candidates by experiment, including the one I was most confident about: `touch-action: none` is **not** the cause — selection and handle-drag both work fine with it, so it stays as-is. Fullscreen isn't it. The canvas repainting every frame isn't it. The container's `contenteditable` is *part* of it: the layer now carries `contenteditable="false"`, a non-editable island inside ghostty's editable container. That's semantically right anyway — these rows are not editable text — and it leaves ghostty's focus and IME alone, which is why I scoped it to the layer rather than stripping the attribute off the container. It restores Chrome's selection-context UI, but still not the toolbar.

**So the fix doesn't depend on the toolbar at all.** dev3 already answers this everywhere else — selecting *is* copying, and `terminal.copyHint` has said so for a long time:

> "Copied! Just select text in any terminal pane — it's on your clipboard automatically, no ⌘C needed."

The touch layer was the one place that rule never reached. It now emits a settled selection 250ms after the last change — your "on each change" with a debounce, because a handle drag fires dozens of changes and each one would otherwise be a clipboard write — and `TerminalView` copies it with `writeClipboardText`. Deliberately **not** `copyTerminalSelection`: that RPC writes the *host's* clipboard, which is the wrong machine when the terminal is being read on a phone.

I did consider a Copy button and talked myself out of it: it would add a control to a surface with a complexity budget, and invent a second model for an action the product already has. `AGENTS.md` is blunt about that being the top UX anti-pattern here, and it's right.

**Verified end to end on the emulator, not just in tests:** long-press a word, then Android's own `KEYCODE_PASTE` into the composer returns that word. Worth knowing — remote mode is plain http, so `navigator.clipboard` doesn't exist and this runs entirely through the documented `execCommand` fallback. That's the path real users on a LAN URL will hit.

`test:full` green, rebased on current main.

Two things I can't cover from here: **Samsung Internet** (only Chrome in the emulator), and whether the one-time `terminal.copyHint` toast reads well on a phone — it was written for desktop. If either bothers you on the device, say so and I'll follow up.

I didn't watch the attached video, since your description already matched what I could reproduce — tell me if it shows something beyond the missing Copy bar and I'll take another look.

mcaldas:
@h0x91b — explicitly asking rather than leaving it implied: **could Arseny retest on the device when he has a moment?**

Head is `9842050`, CI green. What changed since his last run is the auto-copy he suggested, so the thing to check is short: long-press a word in a terminal, lift, then paste anywhere. There is deliberately **no Copy button** — the text should already be on the clipboard, the same way selecting works on the desktop.

If it doesn't land, the most useful detail is whether the page is on `http` or through the tunnel: plain http is not a secure context, so `navigator.clipboard` doesn't exist and the copy runs through the `execCommand` fallback. That fallback is the path I verified on the emulator, but the two contexts take different branches and only one of them is testable here.

I can't request a review through GitHub from a fork, hence asking here.

mcaldas:
Small status update rather than a bare ping: **main had moved 15 commits and this had gone conflicting**, so I've rebased it. Head is now `d4f2ff8`, `MERGEABLE/CLEAN`, CI 11/11.

The rebase pulled in the new touch scroll-to-latest button, which is worth a second pair of eyes since it's the one thing that could collide with this work: it's `z-20` and rendered by the host outside the ghostty container, while the text layer is `z-index: 1` inside it, so the button stays above and tappable. Suite is green, but that's a stacking-order argument, not a device test — if it turns out to misbehave on hardware, that's the first place I'd look.

Otherwise unchanged from Saturday: the auto-copy Arseny suggested is in, verified end to end on a Chrome emulator (long-press a word, then Android's own paste returns it). Whenever he next has the device handy, that's the one thing worth confirming — no rush from my side, and I know a retest is a favour rather than an obligation.

Issue #1608, opened 2026-08-31
Inter-task messages delivered truncated — the head of long messages is lost


yhattav:
## What happens

Messages sent between task agents with `dev3 message --task <id> ...` sometimes arrive in the receiving agent's session with the BEGINNING of the message missing. The injected prompt starts mid-sentence (sometimes mid-word, or with a single stray character before the message body). The tail of the message and the closing `</message>` wrapper are always intact — only the first chunk is dropped.

## Frequency / shape

- Observed repeatedly across two days in one long-running orchestrator session receiving reports from many child tasks.
- Roughly 1 in 3 LONG messages (over ~1.5 KB) arrive truncated; short messages arrive whole.
- The sending side appears fine: the same content is recoverable in full from the sender's own records, so it looks like a delivery/injection issue, not a compose issue.

## Expected

Either deliver long messages whole, or split them into multiple complete turns — never drop the head silently. A truncated delivery is worse than a failed one because the receiver cannot tell what is missing.

## Environment

- dev3 CLI v1.35.1 (3df2b0ab), tmux backend, macOS (Darwin 25.4.0)
- Receiver is a long-lived orchestrator session with frequent incoming agent traffic

## Repro shape

1. Task A (any agent) sends a >1.5 KB message to task B via `dev3 message --task <B> --subject ... "<long text>"`.
2. Repeat with several long messages while task B's pane is active.
3. Compare what B's session received against what A sent — intermittently the first portion is missing in B's prompt.

h0x91b:
Hey, Claude here — the AI assistant working on this repo. Thanks for this one, genuinely: the report came with a repro shape, a frequency *and* a size range, and that is exactly what made it measurable instead of a ghost story. It got two full passes of work.

**Both are shipped in v1.51.0.**

### Pass 1 — #1612 (merged `a681784bc`)

We measured the transport before changing it: 67 sends through the production path and 8 deliveries into a real Claude Code, on a quiet pane, on a pane printing ~1000 lines/s, and with the reader ignoring stdin for 3 s while the payload landed. Every one byte-exact. So the bytes do not die in the CLI socket, the hold queue, the paste, tmux or the pty.

[...]

PR #1634, opened 2026-09-02
Arm the agent-message receipt at the pty chunk boundary

h0x91b:
Hey, Claude here — the AI working on this branch.

Follow-up to #1612 / #1618 on issue #1608. @yhattav found the mechanism behind the lost head, and it is not in dev3's transport: the pty hands the receiving CLI ~1 KiB per read, Claude Code folds the first chunk into a `[Pasted text #N]` attachment and appends the rest inline, and the attachment is intermittently dropped at submit (upstream anthropics/claude-code#90910). The cut lands at byte 1022 every time.

The `<full-copy>` receipt armed at **1 500 bytes of body**, which left every delivery between one read and 1.5 KB unprotected. Now:

- `AGENT_MESSAGE_RECEIPT_THRESHOLD_BYTES` drops to **1 000** — the chunk size, where the exposure actually begins.
- It is measured on the **typed envelope**, not the body: the header is 200–400 bytes of what the pty chunks, so a short body under a long header was exposed all the same. `wrapWithReceipt` measures the bare envelope and re-wraps with the path only when it crosses the line.
- Deliveries that fit one read are unchanged: no receipt line, no file.

Still detection, not prevention — the fix proper is upstream. Decision record supersedes `agent-message-receipt-survives-a-lost-head`; tests pin the envelope-vs-body rule and the constant against the measured 1022.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=48a9aee4-0ffd-49f8-b915-00f9116442ad) · `dev3://task/48a9aee4-0ffd-49f8-b915-00f9116442ad`


sjswerdloff:
please note: https://github.com/anthropics/claude-code/issues/90910#issuecomment-5522087224
we found a 3 line issue as well as the byte boundary.

h0x91b:
@sjswerdloff thanks, checking!

PR #1635, opened 2026-09-02
Title each window with the task or project it shows

### LINKED ISSUE #1632: Multi-window: every window shares one static title — can't tell which task a window shows
## The need

Multi-window (⇧⌘N) is the feature that lets a multi-monitor setup keep several tasks visible at once. My workflow leans on it hard: I keep 3-4 long-running agent sessions (coordinator-style tasks I visit constantly) each pinned to its own window on its own display, and only the main window stays on the board. For that layout to work I must be able to tell **which window is which task at a glance** — in the title bar, in Mission Control, in the Window menu, when clicking through from another app.

## Actual behavior

Every dev-3.0 window carries the identical static title (`dev-3.0 vX.Y.Z`). With several windows open I have to guess who I'm talking to until I read the terminal content itself. Mission Control and the Window menu show N indistinguishable entries.

## Expected

The window title follows what the window shows — e.g. `<task title> · dev-3.0` on a task route, `<project> · dev-3.0` on a board.

## Root cause (from a source read)

The renderer already computes exactly this title per route: the `document.title` effect in `src/mainview/App.tsx` ("Drive document.title from the current route…") builds `"<task/project> · <base>"`, including the streamer-mode masking for sensitive projects. But that only reaches the browser tab in remote mode. On desktop, `src/bun/window-manager.ts` sets `title: opts.title` once at `BrowserWindow` creation and nothing ever calls `setTitle` again — the webview's `document.title` doesn't propagate to the native window in Electrobun.

## Suggested fix

The App.tsx title effect is already the single choke point (route changes, task renames, streamer masking all flow through it). Forward its computed value over RPC to the owning window's `BrowserWindow.setTitle`. Streamer-mode privacy comes along for free since the masked/neutral title is computed in the same place.

## Environment

macOS (Darwin 25.4), dev-3.0 desktop app, multiple windows via File → New Window / ⇧⌘N.

---
*Filed on behalf of the user by their agent after confirming the behavior against the source.*

h0x91b:
Every dev-3.0 window carried the identical static title, so a multi-window setup showed N indistinguishable entries in the title bar, Mission Control and the Window menu (#1632).

The renderer already computed the right string for `document.title`, but a webview's title does not propagate to the native window in Electrobun. It now reports the route **context** (task title, project name, or the streamer-mode placeholder) over a new `setWindowTitleContext` request; the host composes `<context> · <base>` and calls `BrowserWindow.setTitle`, keeping the base (version + build time) it already owned.

The handler is per window — `createAppWindow` overrides it in each window's own RPC instance over a back-reference to that window, so only the window that reported gets retitled (the shared `handlers` object has no idea which window called, and routing through the focused window would let a background window retitle the foreground one). The remote transport keeps a deliberate no-op: a browser client already owns its tab title. Streamer-mode privacy comes along for free, since the masked context is computed in the same effect.

Decision record: `decisions/2026/09/02/per-window-native-title.md`.

Closes #1632

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=2e63ea12-9621-45a2-9773-2e30e1c6d24f) · `dev3://task/2e63ea12-9621-45a2-9773-2e30e1c6d24f`

### DIFF
--- a/change-logs/2026/09/03/feature-per-window-native-title.md
+++ b/change-logs/2026/09/03/feature-per-window-native-title.md
@@ -1,0 +1,5 @@
+Short: Window titles name their task
+
+Each desktop window now shows what it displays in its native title bar — `<task> · dev-3.0` on a task, `<project> · dev-3.0` on a board — so a multi-monitor layout of several windows is tellable apart in the title bar, Mission Control and the Window menu. Streamer mode keeps a sensitive project's name out of the title, exactly as it already did for the browser tab.
+
+Suggested by @yhattav (h0x91b/dev-3.0#1632)
--- a/decisions/2026/09/02/per-window-native-title.md
+++ b/decisions/2026/09/02/per-window-native-title.md
@@ -1,0 +1,44 @@
+# Per-window native title from the renderer's route context
+
+## Context
+
+Every dev-3.0 window carried the identical static title (`dev-3.0 vX.Y.Z [build]`) set once at
+`BrowserWindow` creation. With multi-window (⇧⌘N) on several displays there was no way to tell which
+window showed which task — not in the title bar, not in Mission Control, not in the Window menu
+(issue #1632). The renderer already computed exactly the right string for `document.title`
+(`App.tsx`, the route-title effect), but a webview's `document.title` does not propagate to the
+native window in Electrobun.
+
+## Decision
+
+The renderer reports only the **context** (task title, project name, or the streamer-mode
+placeholder) over a new RPC request `setWindowTitleContext`; the host composes the final title with
+`composeWindowTitle(base, context)` (`src/bun/app-utils.ts`) and calls `BrowserWindow.setTitle`.
+
+Two points worth knowing:
+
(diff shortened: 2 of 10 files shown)


h0x91b:
Hi — Claude here, the AI working on this branch. Linking this to its origin: it implements **#1632** (multi-window: every window shares one static title), and the PR body's `Closes #1632` will close it on merge.

What landed, in short: the renderer reports its route context over a new `setWindowTitleContext` request, and the host composes `<context> · <base>` and calls `BrowserWindow.setTitle`. The handler is per window (each window's own RPC instance, closed over a back-reference to itself), so only the window that reported gets retitled; the remote transport keeps a deliberate no-op, since a browser client already owns its tab title. Rationale and the alternatives I rejected are in `decisions/2026/09/02/per-window-native-title.md`.

Verification beyond the suite: ran the built desktop app and watched the native title bar change live, and confirmed the renderer→host round trip from the running app's log. The one hop I could not drive from outside is a task route in that native window — synthetic clicks and ⌘1 never reach the WKWebView — so that branch is unit-tested rather than eyeballed.

Auto-merge (squash) is enabled; waiting on CI.

PR #1636, opened 2026-09-02
Paste agent prompts through a tmux buffer instead of typing them

mcaldas:
## The bug

A long `dev3 message` between two tasks reached the receiving agent with the **beginning cut off** — it started mid-sentence. Reported four times in one session, always on multi-paragraph bodies of ~1 500 characters or more. Losing the head is the worst half: the ask is at the top, the sign-off at the bottom, so the receiver keeps the pleasantries and loses the instruction. And it does not look like corruption — one receiving agent filled the gap with a plausible, adjacent, wrong mechanism.

## What was actually happening

Measured bottom up on tmux 3.6a and Claude Code 2.1.258:

| Layer | Result |
|---|---|
| tmux → PTY | **Clean.** A raw-mode reader got 512 / 1 500 / 2 500 / 3 500 / 4 500 / 4 900-byte `send-keys -H` payloads byte-identical |
| PTY → JS | **Everything arrives, in pieces.** A Node `setRawMode` reader saw 1 500 bytes as `[1022, 478]`, 4 500 as `[1022, 1022, 1022, 1022, 412]` |
| JS → the agent's input box | **This is the bug.** A 3 163-byte envelope landed as `[Pasted text #1 +10 lines][Pasted text #2 +6 lines][Pasted text #3 +8 lines]` — three independent pastes for one message. Lose one and a whole coherent chunk is gone |
| Same payload, bracketed | **One** `[Pasted text #1 +27 lines]`, head intact |

## The fix

`TmuxClient.sendKeysGuarded` no longer types literal text. Each literal chunk goes into its own `set-buffer -b dev3-paste-<uuid> -- <text>` (argv — no quoting, no hex tripling) and the guarded command list pastes it with `paste-buffer -d -p -b <name> -t <pane>`. `-p` brackets the payload **only when the running app asked for DEC 2004**, so an agent CLI gets one atomic paste and a plain shell still gets plain text. Key steps, the guard, the marker and the verdicts are untouched.

Buffer names are per-send (tmux buffers are server-wide), `-d` drops one on a successful paste, and a non-blocking `finally` clears whatever a refused guard left behind — a message body must not linger where `show-buffer` can read it.

`AGENT_MESSAGE_RECEIPT_THRESHOLD_BYTES` drops 1 500 → 512. The old number came from where the field reports started, not where the risk does: body plus envelope crosses the first ~1 KB boundary near 600 bytes, and two of the four reported losses were under 1 500 — so they arrived mangled with no `<full-copy>` path back to the text.

## Verified by behaviour, not by reading

Driving the **patched client** against a real tmux pane running a real `claude`:

- 3 163-byte envelope delivered whole; the agent echoed its first 40 characters verbatim.
- One `[Pasted text]` placeholder, not three.
- Zero `dev3-paste-*` buffers left on the server afterwards.
- A send with a stale server token (guard refuses): zero bytes into the pane, no buffer left behind.
- `bun run lint` clean; `bun run test` green (5 118 + 7 560 + 1 005).

## Deliberately not done

**The native terminal backend is unchanged.** It writes straight to a PTY it owns and has the same chunking, but `NativeTaskTerminal` exposes no way to ask whether the app enabled DEC 2004 — the mode is parsed in `ghostty-live.ts` and never plumbed to the writer. Wrapping unconditionally would type literal `[200~` into an app that never asked for it. Left alone rather than half-fixed and unverified; called out in the decision record.

**No tail-truncation marker.** dev3 never truncates: past `AGENT_MESSAGE_SPILL_THRESHOLD_BYTES` (4 000) the body is written to a file and the pointer is delivered inside the same envelope. The one report of "the body vanished, only a path arrived" was that spill working as designed.

Full reasoning: `decisions/2026/09/03/paste-agent-prompts-through-a-tmux-buffer.md`.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=c5e26906-64cc-4261-af98-65c3ed52c632) · `dev3://task/c5e26906-64cc-4261-af98-65c3ed52c632`

https://claude.ai/code/session_01F4D9z3BcsBcWm8rdx8CqFD

### DIFF
--- a/change-logs/2026/09/03/fix-agent-message-head-truncation.md
+++ b/change-logs/2026/09/03/fix-agent-message-head-truncation.md
@@ -1,0 +1,9 @@
+Short: Long agent messages arrive whole
+
+Messages sent between tasks with `dev3 message` no longer lose their beginning. Text is now
+pasted into the receiving agent through a tmux paste buffer instead of typed, so a long
+message arrives as one atomic paste rather than as several ~1 KB fragments the receiving
+CLI could drop one of. Messages also keep a `<full-copy>` receipt from 512 bytes up instead
+of 1 500, so the way back to the original text is there whenever anything can go wrong. The
+native terminal backend is unchanged — it has no way to tell whether the running app accepts
+a bracketed paste.
--- a/change-logs/2026/09/03/fix-agent-message-head-truncation.md
+++ b/change-logs/2026/09/03/fix-agent-message-head-truncation.md
@@ -1,9 +1,7 @@
 Short: Long agent messages arrive whole
 
-Messages sent between tasks with `dev3 message` no longer lose their beginning. Text is now
-pasted into the receiving agent through a tmux paste buffer instead of typed, so a long
-message arrives as one atomic paste rather than as several ~1 KB fragments the receiving
-CLI could drop one of. Messages also keep a `<full-copy>` receipt from 512 bytes up instead
-of 1 500, so the way back to the original text is there whenever anything can go wrong. The
-native terminal backend is unchanged — it has no way to tell whether the running app accepts
-a bracketed paste.
+Messages sent between tasks with `dev3 message` no longer lose their beginning: text is pasted
+into the receiving agent through a tmux paste buffer instead of typed, so a long message arrives
+as one atomic paste rather than as several ~1 KB fragments the receiving CLI could drop one of.
+The native terminal backend is unchanged — it cannot tell whether the running app accepts a
+bracketed paste.
--- a/decisions/2026/09/03/paste-agent-prompts-through-a-tmux-buffer.md
+++ b/decisions/2026/09/03/paste-agent-prompts-through-a-tmux-buffer.md
@@ -32,21 +32,24 @@ Measured on tmux 3.6a and Claude Code 2.1.258, bottom up.
 
 `TmuxClient.sendKeysGuarded` no longer types literal text. Each literal chunk is loaded
 into its own `set-buffer -b dev3-paste-<uuid> -- <text>` (argv, so no quoting and no hex)
-and the guarded command list pastes it with `paste-buffer -d -p -b <name> -t <pane>`.
+and the guarded command list pastes it with `paste-buffer -d -p -r -b <name> -t <pane>`.
 `-p` wraps the payload in bracketed paste **only when the running app asked for DEC 2004**,
 so an agent CLI takes one message as one atomic paste while a plain shell still receives
 plain text. Key steps still go through `send-keys`, and the guard, the marker and the
 verdicts are untouched.
 
+`-r` is not optional. Without it tmux replaces every LF in the buffer with a CR — measured on
+3.6a against a raw-mode reader: `send-keys -H 41 0a 42` delivers `A LF B`, `paste-buffer -d -p`
+delivers `A CR B`, and `-r` delivers `A LF B` again. Claude Code does not care (both forms fold
+into one attachment, verified against a live 2.1.259 pane), but `-p` brackets only when the app
+asked for DEC 2004, and an app that did not gets bare CRs, which a TUI reads as submit while LF
+means insert-newline. `-r` keeps the byte stream identical to what `send-keys -H` delivered, so
+Codex, Gemini and Cursor Agent see no change at all.
+
 Buffer names are per-send because tmux buffers are server-wide; `-d` drops the buffer on a
 successful paste and a `finally` deletes anything a refused guard left behind, so a message
 body never lingers where `show-buffer` can read it. That cleanup is deliberately not
 awaited — it must not extend a send that is already hanging.
-
-`AGENT_MESSAGE_RECEIPT_THRESHOLD_BYTES` drops from 1 500 to 512. The old number came from
-where the field reports started, not from where the risk starts: body plus envelope crosses
-the first ~1 KB chunk boundary near 600 bytes, and two of the four reported losses were
-under 1 500, so they arrived mangled with no `<full-copy>` path back to the text.
 
 Verified end to end through the patched client: `sendKeysGuarded` driving a real tmux pane
 running real Claude Code delivered a 3 163-byte envelope whole, the agent echoed its first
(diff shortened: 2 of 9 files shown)


h0x91b:
Hi — Claude here, the AI assistant working on the same defect from the other side. Thank you for this one: the bottom-up measurement table is exactly the shape of evidence this bug needed, and the `paste-buffer -p` trick (bracketing only when the pane actually asked for DEC 2004) is a genuinely nice find — it sidesteps the "raw `[200~` into a plain shell" problem that made us reject bracketed paste before.

Linking the threads, because this problem turned out to be a bit bigger than any one PR sees:

- **The origin report is #1608** (@yhattav). Two passes already shipped there: #1612 (the `<full-copy>` receipt) and #1618 (burst separator).
- **#1634** is the pass in flight, and it overlaps with your receipt change: it moves `AGENT_MESSAGE_RECEIPT_THRESHOLD_BYTES` to 1 000 and measures it on the **typed envelope** rather than the body — the pty chunks what is typed, and the header is 200–400 bytes of that. Your 512-on-body lands in about the same place for a different reason; one of the two should win, not both.
- **The mechanism is upstream, and it is a regression**: anthropics/claude-code#90910. Yesterday's comment there (yhattav) pins the fold-and-drop: the first ~1 KiB chunk becomes a `[Pasted text]` attachment, later chunks go inline, and the attachment is intermittently dropped at submit. Today's comment (sjswerdloff, https://github.com/anthropics/claude-code/issues/90910#issuecomment-5522087224) adds two things that matter for your approach: it does not happen on 2.1.220 / 2.1.227 and does on 2.1.252+, and the fold rule is **800 characters OR more than three lines** — so a dev3 envelope is folded at any size, and a single bracketed paste is still one attachment.

Which leads to the one thing I would want measured before this merges. The reported failure rate on an affected build is ~38% of arrivals. Your end-to-end run delivered one 3 163-byte envelope whole — on a build that fails 38% of the time, a single clean arrival has roughly a 62% chance of happening by luck. And the object that gets dropped is *the attachment itself*, so a whole-message bracketed paste may just move the loss from "head missing, tail and `<full-copy>` path intact" to "everything missing, nothing to detect it with". That is a hypothesis, not a finding — but it is the one your data cannot yet rule out. If you can run the same envelope 30+ times into a 2.1.258 pane and count what the agent actually received (the session transcript under `~/.claude/projects/` is ground truth, not the screen), that number decides it either way, and I would be glad to be wrong here.

Two smaller notes: the native backend has the same chunking and stays exposed, as your record says; and a `set-buffer` that lands but whose paste never runs (app killed in between) leaves the body in the server's buffer list until the next send cleans up.

None of this takes away from the PR — if the bracketed paste survives the drop at n=30, it is the better fix, since it keeps messages as text instead of file pointers. Thanks again for digging this far.

h0x91b:
One more, Claude again — this came from a parallel review pass and is a concrete one-character fix, so posting it separately:

**`paste-buffer` needs `-r`.** Without it tmux replaces every linefeed in the buffer with a carriage return (man tmux, `paste-buffer`: *"any linefeed characters in the paste buffer are replaced with a separator, by default carriage return"*). Measured on tmux 3.6a against a raw-mode reader:

```
send-keys -H 41 0a 42     ->  A LF B    (the stream before this PR)
paste-buffer -d -p        ->  A CR B    (this PR as written)
paste-buffer -d -p -r     ->  A LF B    (with the flag)
```

Harmless for Claude Code — both forms into a live 2.1.259 pane produced one attachment of the same 39 lines, so `-r` changes nothing where you tested. But where the pane's app has **not** enabled DEC 2004 the paste goes out unbracketed and the app sees bare CRs, and in a TUI CR is *submit* while LF is *insert newline*: a multi-line message could submit itself line by line. Codex, Gemini, Cursor Agent and OMP are exactly the receivers nobody has driven through this path yet.

Also worth knowing: no test covers multi-line text through `sendKeysGuarded` — adding `-r` only turns literal command-shape assertions red — which is why the substitution shipped unnoticed. A byte-level assertion on the LF surviving would pin it.

h0x91b:
Hi Mathias — Claude here, the AI assistant working on this from the dev3 side. Arseny asked me to carry your branch the rest of the way rather than leave you doing the merge dance, so I have pushed to it directly. Everything below is what changed and why; your commit is untouched and still yours.

First, thank you for this. The bottom-up measurement table is exactly the evidence this bug needed, and `paste-buffer -p` is a genuinely nice find — taking the bracketing condition from tmux, which already tracks whether the pane asked for DEC 2004, sidesteps the "raw `[200~` into a plain shell" problem that made us pass on bracketed paste before. Three mutants (dropping `-p`, a fixed buffer name, disabling the cleanup) each turned exactly the test named for them red, so your tests are load-bearing.

**Rebased onto #1634.** That PR merged while yours was open and it takes a different route to the same defect: past 1000 bytes of the *typed envelope* the body now spills to a file and the agent gets a short pointer, and the `<full-copy>` receipt is deleted outright — the spill file is the copy. That collided with your receipt-threshold change in three files, so those hunks are gone: the constant they tuned no longer exists. The rest of your branch rebased cleanly, and the two changes are complementary rather than redundant — the spill bounds what is ever typed, yours makes what *is* typed arrive as one paste.

**Added `-r` to the paste.** This is the one real gap. Without it tmux replaces every LF in the buffer with CR — the man page says so, and measured on 3.6a against a raw-mode reader:

```
send-keys -H 41 0a 42     ->  A \n B     (what the old path delivered)
paste-buffer -d -p        ->  A \r B
paste-buffer -d -p -r     ->  A \n B     (parity restored)
```

Claude Code does not care — I drove a 3.2 KB, 40-line payload into a live 2.1.259 pane and both forms folded into a single `[Pasted text +39 lines]`. But `-p` only brackets when the app asked for DEC 2004, and dev3 also drives Codex, Gemini and Cursor Agent through this same seam. Where no brackets are added the app sees bare CRs, and a TUI generally reads CR as submit while LF inserts a newline, so a multi-line prompt could have submitted itself line by line. `-r` costs nothing and keeps the byte stream identical to before. There is now a multi-line test covering it — nothing did.

**One smaller fix:** `timeoutMs` was being handed whole to both the `set-buffer` load and the guarded paste, so one stage could run to roughly twice the deadline its caller computed. It is now a budget for the whole send.

Also trimmed the changelog to the repo's one-to-three-sentence rule, refreshed a stale `-l` comment in `agent-prompt.ts`, and extended your decision record with what the rebase changed. I kept your record as the primary one — the investigation in it is the best write-up any of us has of this bug.

One honest caveat, since it belongs next to your claim rather than hidden in a task note: I built an n=30 rig to measure whether the atomic paste actually defeats the drop, and it returned a null result — the *unpatched* path did not fail either, on a fresh receiver. So there is no measurement either way, and this change stands on the mechanism rather than a measured rate. My working guess is that the drop needs a long-lived loaded session. Not a criticism of your PR; my measurement failed, not yours.

Thanks again for measuring this properly instead of guessing.

PR #1643, opened 2026-09-04
Add a scroll-to-latest button to the terminal on touch devices

arditti:
## The problem

On a phone the terminal runs in compose mode, where a tap on the canvas must not summon the on-screen keyboard. That early-return also swallows the tap that leaves tmux copy-mode on desktop, so once a swipe scrolled the pane into history there was no way back except swiping through the whole history or leaving the task. There was no scroll-to-bottom control at all; the `[280/1699]` counter people take for one is tmux's own copy-mode position readout.

Sending a prompt while scrolled up was broken for the same reason: the keys landed in copy-mode instead of reaching the agent.

## What this adds

A round 48px button over the canvas, bottom-right, above the tmux status line and inside the safe area. It renders only on touch input and only while the pane is scrolled into history, and one tap returns to the live tail.

- Scrolled-up state comes from two sources: the existing copy-mode flag in `TerminalView`, confirmed by a 1.5s poll of a new read-only `tmuxPanesInMode` call, and ghostty's own `viewportY` for the native backend.
- `TerminalHandle.scrollToBottom` covers both backends: leave tmux copy-mode, and scroll ghostty's buffer to the bottom.
- Composer paste and submit now leave copy-mode first, so a prompt is never eaten as copy-mode keys.
- Wired into all terminal layouts: tmux wide and narrow, native tiled and narrow, plus the quick shell.

The poll visits both the agent session and `dev3-dev-<id>`, the same pair the existing exit path walks. An earlier revision asked only the agent session, which made the button vanish a tick after appearing whenever the scrolled pane was the dev server's window.

## Why a floating button

Seven placements were mocked over a real 390px screenshot before anything was built. Rejected: a key in `ExtraKeyBar` and a fifth composer icon, since both rows are at their narrow budget and would spend permanent screen on a state that is usually absent; a button in the pane bar, which owns pane navigation and sits far from the thumb; a full-width strip above the status line, which costs a whole band of screen. The accent fill is a documented exception to one-primary-per-screen, recorded in the manifest: the button only shows while Send is off-target anyway.

Rationale in `decisions/2026/09/03/touch-scroll-to-latest-button.md`; the placement rule is in `docs/ux` (bible sections 10 and 12.3, plus a new `terminal_scrollback_return` surface).

## Verification

Driven in a real browser at 390px with touch emulation: the button is absent on a live tail, appears after scrolling up, survives the poll tick, and after one tap it unmounts while the app logs `Exited copy-mode in panes` for both sessions. It never renders on a pointer device, where a canvas click already does the job.

Eight new tests cover the button, the scrolled-into-history signal, the poll, the paste ordering, the host gating, and the new backend call. Full local suite green: 5132 renderer, 7591 backend, 1005 CLI, no type errors.

Untested: physical phone hardware, so a real finger swipe is the one gap. Known limitation: a wheel-up inside a mouse-tracking program such as vim can show the button for up to 1.5s until the poll clears it.


src/bun/rpc-handlers/tmux-pty.ts
```diff
@@ -2520,6 +2520,37 @@ async function exitCopyModeInSession(socket: string, tmuxSession: string): Promi
...
+ * into history: tmux leaves copy-mode on its own when a swipe reaches the
+ * bottom, and the renderer has no other way to notice.
+ */
+async function tmuxPanesInMode(params: { taskId: string }): Promise<{ inMode: boolean }> {
+	const socket = pty.getSessionSocket(params.taskId);
+	// Both sessions, same reason exitCopyModeAllPanes visits both: the pane the
+	// user scrolled is often the dev-server's (dev3-dev-<id>), and answering for
+	// the agent session alone reports "live" for a pane that is still in copy-mode.
+	const sessions: string[] = [];
+	if (await pty.tmuxSessionExists(params.taskId, socket)) {
```
h0x91b: **Blocker — this gate never matches for the quick shell, so the handler always answers "live" there.**

`pty.tmuxSessionExists()` does not use the session name from the live session map; it recomputes one with the *task* shape (`pty-server.ts:1061` → `computeTmuxSessionName(key, "task")` → `taskSessionName()` → `dev3-<short>`). But `ProjectTerminal.tsx:139` passes `taskId={sessionKey}` where `sessionKey = "project-<projectId>"` (`ProjectTerminal.tsx:46`), and a quick-shell session is actually named `projectTerminalSessionName()` → `dev3-pt-<short>` (`session-names.ts:24`). Those two never agree.

The failure, concretely: on a phone, open Quick Shell and swipe into history. The button appears (that flag is purely renderer-local). One poll tick later `tmuxPanesInMode` returns `{ inMode: false }` — not because tmux left copy-mode, but because `sessions` is empty — so the renderer clears `tmuxCopyModeMayBeActiveRef` and the button disappears while the pane is still scrolled up. That is the exact bug this PR fixes, with no way out. The same cleared flag also disarms the `paste`/`submit` copy-mode exit on that surface, so a prompt typed afterwards is eaten as copy-mode keys.

Worth being fair about where this comes from: the naming mismatch is pre-existing, and `exitCopyModeAllPanes` has always had the identical shape — there it was just a silent no-op, which nobody noticed. What is new is that the poll now *acts* on the wrong answer, and the PR description lists quick shell as a wired surface.

Suggested fix: resolve the name through `pty.getSessionTmuxName(params.taskId)` (it reads the live session map and returns the correct name for either session type) and test *that* name's existence with `tmux.hasSession(...)`, instead of going through `tmuxSessionExists()`. The same change is worth making in `exitCopyModeAllPanes` while you are in here.

src/bun/rpc-handlers/tmux-pty.ts
```diff
@@ -2520,6 +2520,37 @@ async function exitCopyModeInSession(socket: string, tmuxSession: string): Promi
...
+ * scroll-to-latest button polls this while it believes the pane is scrolled
+ * into history: tmux leaves copy-mode on its own when a swipe reaches the
+ * bottom, and the renderer has no other way to notice.
+ */
+async function tmuxPanesInMode(params: { taskId: string }): Promise<{ inMode: boolean }> {
+	const socket = pty.getSessionSocket(params.taskId);
+	// Both sessions, same reason exitCopyModeAllPanes visits both: the pane the
+	// user scrolled is often the dev-server's (dev3-dev-<id>), and answering for
+	// the agent session alone reports "live" for a pane that is still in copy-mode.
+	const sessions: string[] = [];
```
h0x91b: Minor, but it is what makes the blocker above expensive: this whole "which sessions do I visit" block is a line-for-line clone of `exitCopyModeAllPanes` just below (comment included). Two copies means the session-name fix has to land twice, and realistically the next person will fix one of them and move on.

A small shared helper — something like `async function copyModeSessions(taskId, socket): Promise<string[]>` — would let both handlers call it and would give the fix exactly one home.

src/mainview/TerminalView.tsx
```diff
@@ -353,6 +364,49 @@ function TerminalView({ ptyUrl, taskId, projectId, onReady, onNativeStatus, onSe
...
+		copyModePollRef.current = setInterval(() => {
+			api.request.tmuxPanesInMode({ taskId }).then(({ inMode }) => {
+				if (inMode || !tmuxCopyModeMayBeActiveRef.current) return;
+				tmuxCopyModeMayBeActiveRef.current = false;
+				setScrolledIntoHistory(false);
+			}).catch(() => {
+				// Best effort — a missing session just leaves the button until the next tick.
+			});
+		}, COPY_MODE_POLL_MS);
+	}, [setScrolledIntoHistory, taskId]);
```
h0x91b, line 402: **The poll has no touch gate, so it also runs on desktop — where the button can never render.**

`markTmuxCopyModeMayBeActive()` is called from the wheel handler (line 1588) and the mouse-drag handler (line 1442) unconditionally, and there is no touch check here either. So an ordinary desktop wheel-up arms this `setInterval` and it keeps firing `tmuxPanesInMode` every 1.5 s until the user leaves copy-mode. Each tick costs 2–4 tmux process spawns on the bun side (`hasSession` twice, `list-panes` up to twice), and the button it is feeding is gated behind `touchInput` in every host, so a desktop user pays for a control they will never see. Someone parked in copy-mode reading a long log pays it continuously.

Two smaller gaps in the same block, both cheap to close while you are here:

- nothing pauses the interval when the tab is hidden — a `visibilitychange` listener that stops and re-arms would cover a backgrounded remote-mode tab;
- in the native tiled layout each pane has its own `TerminalView`, so N scrolled panes mean N independent intervals.

The host already computes exactly the right boolean: `touchInput` in `TaskTerminal.tsx:61`. Passing it down as a prop, or simply arming the poll only when `onScrolledIntoHistory` is actually provided *and* the host wants it, removes the whole cost from the desktop path without changing the touch behaviour at all.

src/mainview/components/TaskTerminal.tsx
```diff
@@ -825,6 +828,7 @@ function TaskTerminal({ projectId, taskId, tasks, projects, navigate, dispatch,
 							onNativeStatus={makePaneNativeStatusHandler(paneId)}
 							onSessionLost={() => markPaneGone(paneId)}
 							touchComposeMode={touchInput && !rawMode}
+							onScrolledIntoHistory={isFocused ? setScrolledUp : undefined}
```
h0x91b: **`scrolledUp` goes stale when pane focus moves.**

The `isFocused ? … : undefined` means the pane that *loses* focus stops reporting, rather than reporting `false`. So `scrolledUp` keeps whatever value the previously focused pane left behind.

Concretely: pane A is scrolled into history, so `scrolledUp === true`. The user focuses pane B, which is sitting at its live tail. B is mounted with `onScrolledIntoHistory === undefined`, so nothing ever lowers the flag — the button renders over B, and tapping it calls `focusedPaneHandle.scrollToBottom()`, i.e. pane B, which was already live. Meanwhile pane A stays in copy-mode and the one control that would have rescued it is gone.

Two ways out, either is fine: keep the flag per pane id (a `Set<paneId>` or a map, and derive `scrolledUp` from the focused one), or keep the single boolean and reset it to `false` whenever `focusPaneId` changes.

docs/ux/PRODUCT_UX_BIBLE.md
```diff
@@ -670,9 +671,10 @@ Every surface from §5 gets an explicit narrow form. "—" = unchanged.
...
+| Terminal panes | tiled tmux panes | **pane carousel** — one zoomed pane, axis-arbitrated horizontal swipe, a slim top bar, Arrow keys; keep-zoom via `tmuxPaneNavigate` (`MobilePaneCarousel.tsx`) | `Observed` |
+| Terminal windows | tmux windows (workspaces) | **window switcher** — a slim ‹ prev · named dropdown · next › bar above the pane bar; buttons + Arrow-while-focused, no swipe (that is the pane carousel's). Only for window count > 1; `tmuxWindowNavigate` (`MobileWindowCarousel.tsx`) | `Observed` |
+| Terminal text input (touch) | direct typing into the focused terminal | **docked composer** (gate = `!isElectrobun && isTouchDevice`, NOT width — an input-model switch): a terminal tap never summons the OSK; the composer above `ExtraKeyBar` owns text entry, and its sticky `⌨` **raw** toggle restores typing. `UX_DECISIONS.md` (2026-07-02) | `Observed` |
```
h0x91b: **These unrelated `§12.3` rules were compacted to make room for the new one, and some of them lost content that lives nowhere else.**

The file now sits at 131 069 bytes against the 128 KB cap in `src/bun/__tests__/ux-docs-budget.test.ts` — **3 bytes of headroom**. The room came out of neighbouring rules rather than out of fat:

- the docked-composer row (line 676) lost its whole spec: *"Send = mode-2004-aware paste + Enter; Insert = paste only; expand state for long prompts; terminal tail stays visible"* and *"covers Quick Shell too"*;
- the pane-carousel row (line 674) lost **`non-overlapping`** on the dots strip — that word was a constraint on the layout, not decoration;
- further up, `§12` lost the pointer *"Full plans preserved in git history (removed feature-plans/)"*, which was the only record of where that content went.

The budget guard's own header says this in as many words (`ux-docs-budget.test.ts:28`): *"compact those by writing the record first, never by deleting the why."* And the same file shows the sanctioned path — the `122 → 123 → 124 → 125` comments above `BUDGET_KB` are each a deliberate bump with a note explaining what was tried first. Raising `BUDGET_KB["PRODUCT_UX_BIBLE.md"]` from 128 to 129 with a one-paragraph comment, and restoring those clauses, is very much the expected move here rather than a workaround.

There is a practical angle too: 3 bytes of headroom means the next PR that touches this file fails the guard for reasons that have nothing to do with it.

By contrast, the `ux-architecture.yaml` compaction in this PR is fine — `split_resize_divider`'s detail was replaced by a pointer to the bible's `§10` row, and that row genuinely still carries all of it. That is de-duplication rather than deletion.

One last small thing on line 676 while it is in view: this also flips the docked composer from `Proposed` to `Observed`. That is very likely correct — it shipped — but it is a factual claim about a different feature riding inside this PR, so it is worth either calling out in the description or splitting off.

h0x91b (changes requested):
Hey — Claude here, the AI assistant that reviewed this branch for the dev-3.0 repo. Thanks for taking this on, and thank you for the care you put into it: mocking seven placements at 390px before writing a line of code, driving it yourself in a real browser, and writing a decision record that openly admits your own copy-mode flag is a *guess* — that last one is rarer than it should be, and it made the review much easier.

One more thing worth calling out, because it is the rule most PRs here trip over: you respected the `AGENTS.md` "never spawn tmux directly" convention exactly. `tmuxPanesInMode` reuses the existing `tmux.listPanes` client method, the existing `PANE_IN_MODE_FORMAT` from `formats.ts`, and the existing `devServerSessionName` from `session-names.ts` — no raw spawn, no inline `-F` string, no session name recomputed at the call site. Your tests are load-bearing too: I mutation-tested two of them (dropped `touchInput &&` from all four render sites; dropped the dev-server session branch from the handler) and both mutants were caught.

I also drove the feature end to end at 390px with touch emulated on a live tmux pane, and the core loop genuinely works: swipe into history → the button appears → tap → back at the live tail with copy-mode exited and the button unmounted. Sending a prompt while scrolled up works exactly as you describe. No console errors.

So this is close. I am requesting changes for one real blocker plus four smaller things — all left as inline comments on the code:

1. **Blocker — the quick shell never reports copy-mode**, so the button vanishes about 1.5 s after it appears there. `tmuxPanesInMode` gates on `pty.tmuxSessionExists()`, which hardcodes the *task* naming, while `ProjectTerminal` passes a `project-<id>` key whose real session is named differently. Inline on `tmux-pty.ts`.
2. The 1.5 s poll has no touch gate, so it also runs on desktop where the button never renders. Inline on `TerminalView.tsx`.
3. `scrolledUp` goes stale when pane focus moves in the native tiled layout. Inline on `TaskTerminal.tsx`.
4. The session-discovery block is a verbatim clone of `exitCopyModeAllPanes`, which means the blocker has to be fixed twice. Inline on `tmux-pty.ts`.
5. Some unrelated `§12.3` rules were compacted away to fit the docs size cap. Inline on `PRODUCT_UX_BIBLE.md`.

### Smaller nitpicks, entirely your call

- `change-logs/2026/09/03/feature-touch-scroll-to-latest-button.md` — the changelog UI groups by **ship** date, so rename the directory to the actual merge day before this lands. The `Short:` line itself is present and well worded.
- `ScrollToLatestButton.tsx:21` — `title` duplicates `aria-label` on a control that only exists for touch, and a native tooltip never fires on touch. Harmless, just dead weight.
- `TaskTerminal.tsx` lines 893 / 998 / 1078 / 1088 — four copies of the same `touchInput && scrolledUp && <handle> && …` conditional. A tiny local wrapper would make the touch gate impossible to forget in a fifth place later.
- `TerminalView.tsx:400` — `exitCopyModeIfNeeded` clears the flag *before* awaiting the RPC, so a second `submit` arriving inside that round trip takes the synchronous path and its keys can still land in copy-mode. Very narrow and self-healing, so probably not worth code.

### What I checked, and what I did not

Verified: `bun run lint` clean; 178 renderer tests green across `TerminalView`, `TaskTerminal`, `ScrollToLatestButton`, `TerminalComposer`, `ExtraKeyBar`; the desktop click-to-leave-copy-mode path is unchanged (the compose-mode early return is untouched, and the old inline exit became `exitCopyModeIfNeeded()` with the same semantics); design tokens are real (`bg-accent-fill`); 48px clears the 44px touch floor; `env(safe-area-inset-bottom)` is respected; all three locales present; the branch is one commit behind `main` with zero conflicts, and both #1634 and #1636 are already ancestors.

Not checked, so please do not read silence as approval: the native (ghostty, non-tmux) backend was only exercised by unit tests, never driven; the Electrobun desktop app was not opened; Windows, landscape widths and `prefers-reduced-motion` were not tested. And the blocker below is proven from the code rather than driven live — when I clicked "Quick Shell" I landed on a task terminal instead, so I never saw the failure with my own eyes. If you read the chain differently, say so and I will happily be wrong about it.

Genuinely nice piece of work — thanks again, and please keep them coming.

arditti:
All five addressed, no pushback — the blocker chain was right.

1. **Quick shell.** One shared lookup (`copyModeSessions`) now serves both handlers and resolves a session key through the registry, so a quick shell gets its own `dev3-pt-<id>`. That also fixes the desktop click-to-leave-copy-mode path, which was equally blind there.
2. **Poll on desktop.** It arms only when a host passes `onScrolledIntoHistory`, and hosts pass it only on touch. The copy-mode flag still sets unconditionally, so the click path is untouched.
3. **Stale on focus change.** The signal is keyed per pane; the button renders only when the scrolled-up pane is the focused one.
4. **Duplicate discovery.** Gone — item 1 is the shared helper, so the blocker had one place to die.
5. **Docs.** Reverted every compaction: `PRODUCT_UX_BIBLE.md` is untouched in this branch. The rule now lives in the yaml surface `terminal_scrollback_return` plus a pointer entry in the decisions index, and the room came from folding one entry whose reasoning already sits in a `decisions/` record. Worth flagging: the bible had 63 bytes of headroom, so a §12.3 row would not fit without compaction. I left that call to you rather than trimming someone else's rules.

I also found one you did not: I had replaced the yaml's `split_resize_divider` spec with a pointer to the bible, against the rule that those two files are independently authored. Reverted.

Nitpicks: all taken. Dead `title` removed, the four gates collapsed into one helper, the send race fixed (a send inside the cancel round trip now awaits that same request), changelog moved to 09-04.

New tests for each: quick-shell resolution, no-poll-on-pointer while the click path still exits, per-pane keying across two native panes, and the queued second send. Rebased on `main`, `bun run lint` clean, full local suite green.

Still untested, same as before: the native backend driven for real, the desktop app, Windows, landscape, reduced motion.

h0x91b (approved):
Claude again (the AI assistant reviewing this branch) — re-reviewed `0baeb8877`, and this is a genuinely excellent round of fixes. Approving and merging.

All five things I flagged are closed, and three of them you took further than I asked:

| | What I asked | What you did |
|---|---|---|
| Blocker | Resolve the quick-shell session name through the registry | `copyModeSessions()` does exactly that — **and** you noticed `exitCopyModeAllPanes` was equally blind and fixed the desktop click-to-leave-copy-mode path with it. That was not in my comment; it is the same bug and you found it |
| Touch gate on the poll | Gate it on touch | Better: the *presence* of `onScrolledIntoHistory` is what arms the poll, so the gate cannot be forgotten at a call site. And critically, you kept `tmuxCopyModeMayBeActiveRef` armed **before** the early return, so the desktop click path still works — that was the regression I was most worried about, and you guarded it with a test |
| Stale `scrolledUp` | Reset on focus change | Keyed per pane instead, with `prev === key ? null : prev` so only the pane that scrolled can clear its own signal. Strictly better than what I suggested |
| Docs | Raise `BUDGET_KB` to 129 | You reverted the compaction entirely — `PRODUCT_UX_BIBLE.md` is now byte-identical to `main` — and found the room by folding one `UX_DECISIONS.md` entry whose reasoning already lives in a `decisions/` record. That is the path the guard's own header calls sanctioned, and it is the better answer |
| Duplicated session lookup | One helper | `copyModeSessions()`, called by both |

The `copyModeExitRef` / `copyModePending()` work on the double-send race is also more careful than the nitpick deserved — holding the in-flight promise so the second send awaits the *same* request is the right shape, not just a wider flag check.

### What I verified this round

- **Mutation-tested the three fixes that matter**, because a fix without a guard comes back silently. Deleting the `if (!onScrolledIntoHistoryRef.current) return;` touch gate → your test fails. Reverting `copyModeSessions` to the old `tmuxSessionExists` gate → two tests fail. Making the button ignore *which* pane is scrolled (the original bug) → your new `TaskTerminalNative` test fails. All three guards are real.
- `bun run lint` clean; **210 renderer tests** green across the six touched suites; the `ux-docs-budget` and `decision-record-names` suites green.
- CI green on all four required contexts for this head (`lint`, `test`, `build-check`, `trivy-scan`), including the aggregate `test` job and both `terminal e2e` runs.
- Drove it again at 390px with touch emulated on a live tmux pane: no button at the tail, button on scroll-up, still there after 6 s (the poll no longer clears it falsely), tap returns to the live tail, zero console errors.
- Branch is rebased on current `main`, no conflicts.

### Three leftovers — none blocking, noted so they are not lost

1. **`decisions/2026/09/03/touch-scroll-to-latest-button.md` now drifts from its own code.** The record still says the poll reads `pane_in_mode` *"over the task session"* (it is registry-aware now, and a quick shell resolves to `dev3-pt-`), says `paste`/`submit` exit *"when the flag is up"* (it is `copyModePending()` now), and says nothing about the mechanism that ended up carrying the whole touch gate — that the poll arms only when a host passes `onScrolledIntoHistory`. `AGENTS.md` asks for the record to be corrected in the same commit, precisely because the next agent will trust it. Not worth another round trip from you, so we will fix it on our side.
2. `src/bun/pty-server.ts:233` — the old `/** Compute the tmux session name … */` line now sits above `PROJECT_SESSION_KEY_PREFIX`, leaving two JSDoc blocks stacked and the function itself uncommented. Pure leftover from the edit.
3. Small coverage gap: collapsing `prev === key ? null : prev` to just `null` in `scrollSignalFor` survives the whole suite. Your test proves the important half (the button targets the *right* pane — it asserts `pane-2` is untouched); the unproven half is "another pane reporting `false` must not clear my key". Worth one more assertion if you are ever back in that file.

Thank you — for the feature, and for how you took the review. This was a pleasure to read the second time around. Merging now.

PR #1738, opened 2026-09-11
fix(agents): hand Claude its protocol body as a file on every platform

### LINKED ISSUE #1734: Agent system prompt is inlined into argv on Unix, so one agent's pkill -f SIGTERMs every other agent
## Summary

On macOS and Linux, dev3 launches each Claude agent with the entire task-lifecycle protocol inlined into `--append-system-prompt`. That makes every agent's command line roughly 29 KB of prose.

Because `pkill -f` / `pgrep -f` match against the full command line, **any pattern that happens to appear as an ordinary word inside that injected prose matches every running agent on the machine**. A single agent running a routine cleanup command silently SIGTERMs every sibling agent.

dev3 already has the safe transport implemented (`--append-system-prompt-file`), but it is gated to Windows PowerShell only.

## Impact

- Blast radius is every agent process on the machine, across all projects and all tasks.
- The agent that runs the command is spared, so the operator sees "every session except the one I was watching died". This points the investigation at the platform, the network or the OS rather than at a sibling agent.
- Nothing in dev3's own logs records it. The pane shows only a bare red `✗ Process exited with code 143`, and dev3's run script then hands the pane back to an interactive shell, so the pane looks healthy.
- It has happened twice on one machine in two days, from two unrelated agents doing two unrelated cleanups.

## Actual vs expected

**Actual.** An agent runs `pkill -f 'agent-browser'` to restart a browser daemon it owns. Every other agent on the machine receives SIGTERM, runs its shutdown hooks and exits 143.

**Expected.** Killing a process by command-line pattern should match the intended target only. An agent's injected system prompt should not be part of the text that other tools pattern-match against.

## Root cause

In `app/bun/index.js`, the Claude launch arguments are built as:

```js
if (!options?.skipSystemPrompt) {
  if (options?.systemPromptFile) {
    args.push("--append-system-prompt-file", quoteIfUnsafe(options.systemPromptFile));
  } else {
    args.push("--append-system-prompt", shellEscape(CLAUDE_SKILL_BODY));
  }
}
```

`systemPromptFile` comes from `systemPromptFileFor(adapter)`, which only ever returns a path when:

```js
function systemPromptNeedsFile(platform2 = process.platform) {
  return launchDialectId(platform2) === "windows-powershell";
}
```

So `ensureAgentSystemPromptFile()` already writes the prompt to a file and passes `--append-system-prompt-file`, complete with change detection. It is switched on only because PowerShell cannot carry a 29 KB quoted argument. The gate is about shell quoting, not about safety.

On every other platform the full protocol stays in argv.

## Minimal reproduction

On macOS or Linux, with at least two dev3 tasks running a Claude agent:

```bash
# Harmless proof: these are sibling agent PIDs, not browsers or hibernating anything.
pgrep -lf agent-browser
pgrep -lf read-only
pgrep -lf hibernate

# Confirm the argv size of any matched agent.
ps -p <pid> -ww -o command= | wc -c
```

On the machine this was found on, right now:

| Pattern                      | Matching processes |
| ---------------------------- | ------------------ |
| `agent-browser`              | 12                 |
| `read-only`                  | 7                  |
| `hibernate`                  | 7                  |
| argv size of a matched agent | 28,899 bytes       |

Every matched PID is a `claude` agent. Replacing `pgrep` with `pkill` in any of those lines terminates all of them.

## Environment

| Item             | Value                       |
| ---------------- | --------------------------- |
| dev3             | v1.52.0, canary channel     |
| dev3 CLI build   | 37491387, 2026-09-11        |
| Claude Code      | 2.1.268                     |
| macOS            | 15.7.9, build 24G830, arm64 |
| Terminal backend | tmux                        |

## Additional note: a second, independent trap

One of the two incidents used this command:

```bash
pkill -f "port-forward -n <ns> svc/<svc> 8443:80" -o
```

The author intended `-o` as "oldest match". macOS does not permute options, so the trailing `-o` is parsed as a **second pattern** and matches every process whose command line contains `-o`. Verified with `pgrep`: 34 processes matched on this machine.

That is a mistake in the agent's command rather than a dev3 defect, but it is worth knowing that it interacts badly with the argv exposure above.

## Diagnostic signature, for anyone who hits this

- Pane ends with `Resume this session with: claude --resume <id>` followed by `✗ Process exited with code 143`.
- Claude Code's SIGTERM handler is the only path that produces 143, so the process was signalled, it did not crash.
- Per-session death time is the mtime of `~/.claude/projects/<slug>/<session-id>.jsonl`, which gets a final flush on shutdown.

L4XB:
Fixes #1734

## Problem

On macOS and Linux dev3 launched Claude with the whole task-lifecycle protocol inlined into `--append-system-prompt`, about 29 KB of prose in every agent's argv. `pkill -f` and `pgrep -f` match the full command line, so any pattern that occurs as an ordinary word in that prose (`agent-browser`, `read-only`, `hibernate`, …) matched every running agent, and one agent's routine cleanup SIGTERMed all of its siblings. The safe transport, `--append-system-prompt-file`, already existed but was gated to Windows, where only the command-line ceiling forced it.

## Change

`src/bun/agents.ts`

- `systemPromptFileFor()` no longer checks `systemPromptNeedsFile()`: Claude gets the file on every platform. The inline body remains the fallback when `ensureAgentSystemPromptFile()` cannot write it (unchanged behaviour, now documented as such). The doc comment explains why argv is the wrong place for the body.

`src/bun/agent-system-prompt-file.ts`

- Comments only: `systemPromptNeedsFile()` keeps its Windows semantics (the platform that cannot launch inline at all) and is now described as the floor rather than the switch.

Nothing changes for Codex, Cursor and the others: they have no file flag, and the description quoting is untouched.

## Tests

- `agents.test.ts`: new case pins that on POSIX the command carries `--append-system-prompt-file`, contains neither the inline flag nor the first 40 characters of the body, and that the file under `AGENT_PROMPTS_DIR` holds exactly `CLAUDE_SKILL_BODY`. It fails on the previous commit.
- `agent-launch-args.test.ts`: "resolveAgentCommand still emits the POSIX escape on POSIX" asserted the inline body's `'\''` sequence and is replaced by a case that expects the file flag, no body escape, and the task description still single-quoted the POSIX way. "without a file the body is still inline" is retitled to say it is the write-failure fallback.
- `agent-command-line-budget.test.ts` (which mocks the writer away to measure the inline worst case) is unchanged and still passes.

`bunx vitest run --config vitest.config.bun.ts` on the three suites: 191 passed. `bun scripts/lint.ts` is clean.

### DIFF
--- a/change-logs/2026/09/12/fix-system-prompt-file-transport.md
+++ b/change-logs/2026/09/12/fix-system-prompt-file-transport.md
@@ -1,0 +1,5 @@
+Short: Agents stop killing each other
+
+Claude's dev3 protocol now reaches the agent as a file on macOS and Linux too, not only on Windows. Inline it sat in every agent's command line, so `pkill -f` and `pgrep -f` matched ordinary English words from that prose and one agent's cleanup terminated every other agent on the machine.
+
+Suggested by @L4XB (h0x91b/dev-3.0#1734)
--- a/decisions/2026/08/28/agent-command-lines-quote-in-the-launch-dialect.md
+++ b/decisions/2026/08/28/agent-command-lines-quote-in-the-launch-dialect.md
@@ -1,5 +1,11 @@
 # Agent command lines quote in the launch dialect
 
+> Partially superseded on 2026-09-12 by
+> `decisions/2026/09/12/claude-protocol-always-travels-as-a-file.md`: the file channel is no
+> longer a Windows-only answer to the command-line ceiling — every platform uses it, because
+> the inline body in `argv` is what `pkill -f` matches (h0x91b/dev-3.0#1734). Everything below
+> about quoting dialects still holds.
+
 ## Context
 
 Starting or resuming any Claude session on Windows died before the binary was
--- a/decisions/2026/09/12/claude-protocol-always-travels-as-a-file.md
+++ b/decisions/2026/09/12/claude-protocol-always-travels-as-a-file.md
@@ -1,0 +1,62 @@
+# Claude's protocol always travels as a file, and a failed write is fatal
+
+## Context
+
+The dev3 protocol was injected into Claude's launch as `--append-system-prompt <body>`,
+about 29 KB of ordinary English prose sitting in every agent's `argv`. `pkill -f` and
+`pgrep -f` match the whole command line, so any pattern that occurs as a word in that prose
+matched every running agent: one agent restarting its own browser daemon SIGTERMed every
+sibling on the machine, twice in two days (h0x91b/dev-3.0#1734). The victim's pane showed only
+`✗ Process exited with code 143`, which points the investigation at anything but the real cause.
+
+The file channel already existed — written for Windows, where the body plus a task description
+does not fit the 32 767-character command line
+(`decisions/2026/08/28/agent-command-lines-quote-in-the-launch-dialect.md`). It was gated to the
+PowerShell dialect, so no POSIX launch ever used it.
+
+## Investigation
+
+The gate turned out to be the whole bug: `systemPromptNeedsFile()` answered a question about
+shell quoting, and the call site used it as if it answered a question about safety. Nothing else
+had to change to close the exposure — ungating `systemPromptFileFor()` is the fix, and the
+contributor's PR #1738 did exactly that.
+
+Two follow-ups were needed to make it hold. First, the write was not atomic: `~/.dev3.0` is
+shared by every installed version of the app, different versions carry different protocol text,
+so two versions running side by side fail the "has the body changed" check in both directions and
+rewrite the file continuously. The reader is another process — the agent being launched — so a
+truncate-then-write could hand it half a protocol with no error anywhere. Second, a failed write
+returned `null` and the launch silently fell back to the inline body, which is the exposure this
+change exists to remove.
+
+## Decision
+
+`src/bun/agents.ts` — `systemPromptFileFor()` supplies the file for every Claude launch on every
+platform. `src/bun/agent-system-prompt-file.ts` — `ensureAgentSystemPromptFile()` writes through a
+temp name in the same directory and renames it into place, and **throws** when it cannot write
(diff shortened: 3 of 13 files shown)


L4XB:
CI was right and I had missed two suites that still describe the inline body. Fixed at `81f9624`, rebased onto `main`.

- `src/bun/__tests__/agent-command-golden.test.ts`: all 16 claude cases compared against `--append-system-prompt <CLAUDE_BODY>`. `redact()` now also folds the prompt file path to `<CLAUDE_BODY_FILE>`, since it is machine specific, and the cases expect `--append-system-prompt-file <CLAUDE_BODY_FILE>`. The header comment that names the sentinels says so too. 108 passed.
- `src/mainview/__tests__/agents.test.ts`: `injects --append-system-prompt for claude base command` asserted `cmd` contains `dev-3.0`, which only passed because the protocol prose was inlined into argv, the exact thing this PR removes. It now pins the file flag, the path under `agent-prompts/`, and that the body is absent. 106 passed.

The delivery-channel coverage is unchanged otherwise: the golden matrix still fails on a wrong flag or a lost quote, and `agent-command-line-budget.test.ts` still measures the inline worst case through its mocked writer. `agents.test.ts`, `agent-launch-args.test.ts` and the budget suite are at 191 passed. The only `tsc` error in the tree is `three` missing types inside `node_modules/electrobun`, which is there on `main` as well.

h0x91b:
Thanks a lot for this, and for the issue that found it — a really valuable catch. The argv exposure had been sitting there since the file channel was first added, and it took someone actually hitting it to see it.

Nice follow-up on the two stale suites too. I pushed one commit on top of your branch (`1094135`) with the rest: the prompt file is now written through a temp name and renamed into place, a failed write throws instead of quietly falling back to the inline body, the dead `systemPromptNeedsFile` helper is gone, the Settings launch preview no longer advertises the old flag, plus a changelog entry and a decision record.

CI is green — merging. Great first contribution, hope to see more of them.

PR #1741, opened 2026-09-11
Deliver Claude's system prompt through a file on every platform

### LINKED ISSUE #1734: Agent system prompt is inlined into argv on Unix, so one agent's pkill -f SIGTERMs every other agent

h0x91b:
Fixes #1734.

dev3 launched every Claude agent with `--append-system-prompt <~27 KB of protocol>`, so the whole lifecycle protocol sat in the agent's argv. `pgrep -f` / `pkill -f` match the full command line, so an ordinary word inside the protocol (`agent-browser`, `read-only`, `hibernate`) matched every running agent on the machine — one agent restarting its own browser daemon SIGTERMed every sibling. The safe channel already existed but was gated to Windows PowerShell, where the reason was the command-line ceiling rather than safety.

Claude now always gets `--append-system-prompt-file`. Prompt files are content-addressed (`claude-<sha256:16>.md`) and written through a temp file + rename, so two app versions running side by side cannot overwrite each other's body between dev3's write and the child's read. A failed write throws instead of silently restoring the inline form — that fallback was the exposure itself.

Verified by running it, because the flag is not printed in `claude --help` on 2.1.236: a prompt file with a random sentinel comes back from `claude -p` (control without the flag answers NONE), a live launch shows 214 bytes of argv and zero `pgrep -f` matches on the prompt text, while the old inline shape shows 14 645 bytes and four matches. Read-only evidence only: `ps`/`pgrep`, never `pkill`, no sibling process touched.

Note: already-running agents keep their old argv until they are restarted. A real Windows run was not done; the file channel there is unchanged in shape, only the file name became content-addressed.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=b99fb169-2b28-48fb-9cfc-12e99b5e0843) · `dev3://task/b99fb169-2b28-48fb-9cfc-12e99b5e0843` _(dev3 added this footer — turn it off in Settings → Tasks.)_

### DIFF
--- a/change-logs/2026/09/11/fix-system-prompt-file-transport.md
+++ b/change-logs/2026/09/11/fix-system-prompt-file-transport.md
@@ -1,0 +1,5 @@
+Short: Agent prompts no longer sit in argv
+
+Claude agents now receive the dev3 lifecycle protocol through a file on every platform instead of a ~27 KB `--append-system-prompt` argument. The protocol used to sit in each agent's command line, so a `pkill -f` on an ordinary word like `agent-browser` matched and terminated every other agent on the machine. Existing sessions keep their old command line until they are restarted.
+
+Suggested by @vit-pavlenko (h0x91b/dev-3.0#1734)
--- a/decisions/2026/08/28/agent-command-lines-quote-in-the-launch-dialect.md
+++ b/decisions/2026/08/28/agent-command-lines-quote-in-the-launch-dialect.md
@@ -84,6 +84,12 @@ On Windows the protocol also stops travelling on the command line at all:
 2.1.112 that the flag exists and applies the file's content. POSIX keeps the
 inline form, so nothing about a macOS or Linux launch changes.
 
+> Partly superseded on 2026-09-11 by
+> `decisions/2026/09/11/claude-system-prompt-always-travels-as-a-file.md`: POSIX
+> no longer keeps the inline form, and the file name is now content-addressed
+> (`claude-<sha256:16>.md`, not `claude.md`). The inline body also sat in every
+> agent's `argv`, where `pkill -f` matched it and killed sibling agents.
+
 **The file only covers Claude**, because only Claude has a flag that takes a
 path. Codex carries the protocol in `-c developer_instructions=`, and Cursor and
 OpenCode concatenate it onto the prompt, so for them the body is on the command
--- a/decisions/2026/09/11/claude-system-prompt-always-travels-as-a-file.md
+++ b/decisions/2026/09/11/claude-system-prompt-always-travels-as-a-file.md
@@ -1,0 +1,66 @@
+# Claude's system prompt always travels as a file
+
+## Context
+
+dev3 injected the ~27 KB task-lifecycle protocol into every Claude launch with
+`--append-system-prompt <body>`, so the whole protocol sat in each agent's
+`argv`. `pgrep -f` / `pkill -f` match against the full command line, so ordinary
+English words inside the protocol — `agent-browser`, `read-only`, `hibernate` —
+matched every running agent on the machine. One agent restarting "its own"
+browser daemon SIGTERMed every sibling agent, twice on one machine in two days
+(h0x91b/dev-3.0#1734). The file channel already existed but was gated to Windows
+PowerShell, where the reason was the 32 767-character command-line ceiling.
+
+## Investigation
+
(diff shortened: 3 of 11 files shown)


h0x91b:
#1738 were first

PR #1755, opened 2026-09-12
Gate the mobile terminal input stack on a coarse primary pointer

vit-pavlenko:
## Problem

`TerminalComposer`, `ExtraKeyBar`, touch compose mode and the copy-mode scroll poll armed whenever `navigator.maxTouchPoints > 0` in browser remote mode. A Windows laptop with a touchscreen reports touch points even when driven entirely by a mouse, so a 1900px desktop session over `dev3 remote` got the phone composer and a vw-sized key bar — and, since compose mode blocks `mousedown`/`mouseup`/`click` on the terminal (by design, to keep taps from summoning the OSK), it also killed mouse text-selection and click-to-switch-pane.

## Fix

New `isTouchPrimary()` in `src/mainview/utils/platform.ts`: `maxTouchPoints > 0` stays as the cheap precondition, `matchMedia("(pointer: coarse)")` discriminates a phone/tablet from a touchscreen laptop, falling back to the old touch-points-only behaviour where `matchMedia` is absent. Both call sites (`TaskTerminal.tsx`, `ProjectTerminal.tsx`) use it.

Rationale and alternatives in `decisions/2026/09/12/touch-primary-gates-mobile-terminal-input.md`.

## Tests

- `platform.test.ts` — direct `isTouchPrimary()` coverage (phone, touchscreen laptop, no touch, no matchMedia fallback)
- `TaskTerminal.test.tsx` — `setTouchDevice` now stubs the pointer query too; new touchscreen-laptop (touch + `pointer: fine`) case
- `ProjectTerminal.test.tsx` — mirrored touch visibility cases
- `TaskTerminalNative.test.tsx` — the scroll-to-latest test now stubs a coarse pointer alongside `maxTouchPoints`

## Validation

`bun run lint` and the full local `bun run test` green; CI green. Verified on the affected hardware (Windows touchscreen laptop, browser remote mode): the device probes `[navigator.maxTouchPoints, matchMedia('(pointer: coarse)').matches]` → `[5, false]`, and a headless run of this branch's build confirmed the composer/key bar are gone while mouse selection and pane-click focus work again. A coarse-pointer device keeps the mobile input stack (unit-tested; no phone hardware check).

Origin: dev3 task 4.

https://claude.ai/code/session_01PvDm5ywmh8nwZPchFaepew

### DIFF
--- a/change-logs/2026/09/12/fix-touch-ui-coarse-pointer-gate.md
+++ b/change-logs/2026/09/12/fix-touch-ui-coarse-pointer-gate.md
@@ -1,0 +1,3 @@
+Short: No phone keyboard on touch laptops
+
+Fixed the browser remote mode showing the phone terminal composer and extra-key bar on touchscreen laptops and desktops: the mobile input stack now requires touch to be the device's primary pointer (a coarse-pointer media query), not merely a touchscreen being present.
--- a/decisions/2026/09/12/touch-primary-gates-mobile-terminal-input.md
+++ b/decisions/2026/09/12/touch-primary-gates-mobile-terminal-input.md
@@ -1,0 +1,17 @@
+# Touch-primary detection gates the mobile terminal input stack
+
+## Context
+
+`TaskTerminal` and `ProjectTerminal` switched into the mobile input stack (TerminalComposer, ExtraKeyBar, touch compose mode, the copy-mode scroll poll) whenever `navigator.maxTouchPoints > 0` in browser remote mode. A Windows laptop with a touchscreen reports touch points even when driven entirely by a mouse, so a 1900px desktop session over `dev3 remote` got the phone composer and a vw-sized key bar.
+
+## Decision
+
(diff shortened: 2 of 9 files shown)


h0x91b (approved):
Hey, Claude here — the AI assistant that reviewed this branch.

Approving. The gate is in the right place: one named helper (`isTouchPrimary()`) instead of the check being duplicated across both terminal components, and the decision record is honest about what it trades away (a convertible flipped into tablet mode keeps its laptop verdict until remount).

I checked the tests actually guard the behaviour rather than just passing: replacing the `matchMedia` line with `return true` fails exactly three of them — `platform.test.ts:84` plus both "stays hidden on a touchscreen laptop" component tests. `bun run lint` is clean and the four touched suites are green.

Two things I'd pick up as follow-ups, neither blocking:

1. `ProjectTerminal.test.tsx` — the `afterEach` restore is a no-op. `setTouchPoints` defines an **own** property on `navigator`, while the restore writes the descriptor back onto `Navigator.prototype`, so the own property keeps shadowing it (I verified: `navigator.maxTouchPoints` is still `10` after the "restore"). It's harmless today only because that `describe` is last in the file. The same pattern already exists in `TaskTerminal.test.tsx`, so it's inherited rather than introduced here.
2. Worth one pass on a real phone, and on an iPad with a trackpad if you have one around. The failure modes aren't symmetric: a false positive is the old annoyance (phone UI on a laptop), a false negative takes the on-screen keyboard away from someone who has no other way to type.

Nice fix — thanks for writing the decision record along with it.

PR #1759, opened 2026-09-12
Reserve the completion dialog's loss warning for unreachable commits

### LINKED ISSUE #1684: Completion dialog says merged-but-unpushed commits "will be lost"
## What happened

Completing a task whose branch tip had already been fast-forwarded into the local base branch showed:

```
25 commits never pushed — will be lost
The worktree and branch will be deleted. Continue?
```

The commits are reachable from local `main` after the task branch ref is deleted, so completing loses nothing; only the push is missing. Read literally, the dialog stopped me from completing a finished task until I verified with git that the work was safe.

## Why this matters beyond the wording

The check appears to treat "on a GitHub remote" as the only definition of saved. That assumption doesn't hold for every repository dev3 manages. Some projects are tracked in git but deliberately never pushed, or are pushed only occasionally, and for them the local base branch *is* the destination — personal tooling, dotfiles, machine-specific automation, anything with a remote that is a backup rather than the source of truth. For those repositories, every completion will show this warning, and a warning that fires on every task trains people to click through it. That is the worst outcome for a safety dialog, because the day the commits really are unreachable, the message looks identical.

## Expected

Two distinct messages, since the two situations carry different risk:

- When `git merge-base --is-ancestor <branch-tip> <local-base>` holds: `merged into <base> locally but not pushed`. Still worth a warning, because a local-only copy is not durable, but it should not claim the commits will be lost.
- Reserve `will be lost` for commits that are unreachable from every other ref, which is the only case where deleting the branch actually destroys them.

A small "why?" expander listing which refs contain the commits would make either message self-explaining. Optionally, a per-project setting declaring that the base branch is the destination (no remote expected) would let those projects skip the unpushed warning entirely while keeping the real unreachable check.

## Repro

1. Create a task; make a few commits on its branch.
2. In the main checkout, `git merge --ff-only <task-branch>` into the base branch. Do not push anything.
3. Move the task to Completed.
4. Dialog reports every commit as "never pushed — will be lost", although all of them are now on the local base branch.

## Environment

- dev3 CLI v1.46.0 (77929c08)
- macOS 15.6 (Darwin 24.6.0)
- tmux backend

Prasad8830:
Fixes #1684.

The completion dialogue claimed N commits never pushed - will be lost" for a task branch that had been fast-forwarded into the local base and simply not pushed. Those commits stay reachable from `main` after the branch ref is deleted, so completing lost nothing.

### Cause

`getUnpreservedCount` asks `git rev-list HEAD --not --remotes`, and a local branch is not a remote ref, so a fully merged branch keeps the `-1` "nothing preserved" sentinel. `getUnsavedWork` then falls back to `ahead`, measured against `origin/<base>`, which a pending push leaves non-zero. Only repos that have an `origin` are affected: with no remote, the compare ref falls back to the local base, `ahead` becomes 0, and the dialogue is already silent.

### Fix

New `isPreservedOutsideBranch` (`src/bun/git.ts`) runs one sweep:

    git rev-list --count HEAD --not --exclude=<branch> --branches --remotes --tags

It surfaces as `preservedOutsideBranch` on `BranchStatus`/`UnsavedWork`, and `unsavedWorkWarnings` picks `task.warnUnpushedButKept*` over the loss wording. The commits are still reported as unpushed, because a local-only copy is not durable; they are just no longer described as about to be lost. `unpushed` keeps its exact meaning and its `-1` sentinel, so merge detection and PR polling are untouched.

Remote-tracking refs are deliberately not excluded: `removeWorktree` deletes only the local branch and dev3 never runs `push --delete`, so `origin/<branch>` outlives the task.

### Unproven never reads as safe

The sweep is bounded at 2s through the `timeoutMs` that `run()` already supports, so no new infrastructure. A timeout lands in the same branch as any other git failure and returns false, as do an empty branch name (a detached HEAD leaves nothing to exclude) and a repo with no other refs. False keeps the loud warning.

### A trap worth flagging

`--exclude` takes the name with `refs/heads/` already stripped, and applies only to the `--branches` that follows it. Spelling it `--exclude=refs/heads/<branch>` matches nothing, leaves the task branch in the negative set, and makes HEAD answer for itself, so every branch reads as preserved and every real loss warning goes quiet. A mocked-spawn test cannot see this, which is why the new suite runs against real repositories. `git-preserved-outside-branch.test.ts` fails on that spelling.

### Performance

Synthetic fixture, built with `git fast-import`: 500,000 commits on a linear `main`; 35,003 refs (20,000 branches and 10,000 tags spread evenly over the history, 5,001 remote-tracking refs); task branch 5 commits ahead. Median of 7 runs, git 2.43.0.

| check | no commit-graph | with commit-graph |
| --- | --- | --- |
| new reachability sweep | 86-93 ms | 38 ms |
| existing `--not --remotes` (already on this path) | 24 ms | 13 ms |

The sweep costs the same in all four states (unpreserved, preserved by the local main, by another branch, or by a tag), so the answer does not affect the price. The 2s bound is roughly 20x the measured worst-case and falls within the dialogue's own 4s confirm gate.

`git for-each-ref --contains HEAD --count=1` expresses the same question and was rejected: it exceeded a 180s bound on the same fixture, because `--contains` is evaluated per ref.

### Tests

`src/bun/__tests__/git-preserved-outside-branch.test.ts`, against real repositories:

- commits only on the task branch: not preserved (the guard for the `--exclude` spelling)
- fast-forwarded into the local base: preserved
- another local branch holds them: preserved
- a tag holds them: preserved
- a remote-tracking ref under another name: preserved
- merged locally, then one more commit on top: not preserved
- a branch whose name merely starts with the task branch's is not the task branch
- empty branch name, and a failing git: not preserved

Plus the wording split in `confirmTaskCompletion.test.ts` (including that unproven preservation keeps the loss warning) and the payload wiring in `rpc-handlers.test.ts`.


h0x91b:
Thank you, Prasad, for the fix, the thorough investigation, and the regression tests! All checks passed, and I've merged this.

I'm sorry for the slow response — I missed your notification in my email about five days ago, and it slipped through the cracks. Thanks for your patience and for contributing!

Prasad8830:
> Thank you, Prasad, for the fix, the thorough investigation, and the regression tests! All checks passed, and I've merged this.
> 
> I'm sorry for the slow response — I missed your notification in my email about five days ago, and it slipped through the cracks. Thanks for your patience and for contributing!

No worries, @h0x91b, thanks for merging my PR, this was my first contribution here, and I will keep contributing further

PR #1763, opened 2026-09-12
Fix orphaned model proxies across app restarts

mcaldas:
## Summary

- Replace deferred quit cleanup with a statically imported synchronous SIGTERM path, covering running, starting, and asynchronously stopping proxies. Prevent delayed startup or retries after quit.
- Add optional proxy/owner PIDs to the existing endpoint file. Reclaim only the default or remembered address the replacement will bind, after fresh executable/open-database identity and owner-liveness checks.
- Keep the port-release wait for RPC stop/restart. Preserve other live instances, unrelated listeners, and orphaned routes on unselected ports.
- Add regression coverage, a user-visible changelog, and a decision record superseding the earlier squatter assumption.

## Verification

- `bun run lint`: passed.
- `bun run test:full`: passed on the final rebased commit: 14,963 passed, 3 skipped across 890 files.
- Packaged macOS app with the real pinned Bifrost binary and hardened Bun runtime, using an isolated QA home:
  - Confirmed normal quit exits both app and proxy.
  - Force-killed only the main process, confirmed the proxy became orphaned, then confirmed startup reclaimed it and reused its port.
  - Five consecutive quit/relaunch cycles kept the same port and session key; authenticated requests through a local test provider returned HTTP 200 after each.
  - Removed the optional PID fields to exercise legacy listener/open-file discovery; the packaged app reclaimed the orphan.
  - Installed app continued owning 32123 throughout; the QA app used remembered port 59269 and never killed the installed proxy.
- QA app and proxy stopped after verification.

## Limits and Follow-up

- A real routed Codex conversation across restart and legacy takeover specifically on 32123 remain manual acceptance checks. QA used authenticated chat-completion requests, not a real provider account or agent turn.
- Windows fails closed when identity cannot be verified: optional PID metadata alone cannot safely authorize a kill. No Windows process-enumeration scope was added.
- Existing orphans on other ports remain untouched because they may still serve running sessions.
- The separate deferred task-tunnel cleanup risk was filed as dev3 task #22. The unrelated provider `/v1` URL issue was not changed.

---

🔗 **Origin task in dev3:** [open in dev3](https://dev3.h0x91b.com/open.html?task=688e7e58-d232-49ca-b9d1-7007cab4a5c3) · `dev3://task/688e7e58-d232-49ca-b9d1-7007cab4a5c3` _(dev3 added this footer — turn it off in Settings → Tasks.)_

### DIFF
--- a/change-logs/2026/09/13/fix-model-proxy-orphans.md
+++ b/change-logs/2026/09/13/fix-model-proxy-orphans.md
@@ -1,0 +1,3 @@
+Short: Keep model routes through restarts
+
+Model proxies are signalled immediately when dev3 quits, including during startup. On supported systems, restarting reclaims a verified orphan only on the address the replacement will use, preserving the session key while leaving other live instances and legacy routes alone.
--- a/decisions/2026/08/15/model-catalog-and-model-roles.md
+++ b/decisions/2026/08/15/model-catalog-and-model-roles.md
@@ -136,6 +136,8 @@ is written literally — it is a URL, not a credential.
 
 ## Follow-up: the endpoint is remembered, not fresh per start
 
+Superseded on 2026-09-13 by `decisions/2026/09/13/model-proxy-dies-with-the-app.md`: a squatter may be our own orphan, so verify ownership and reclaim only the address the replacement will take before falling back.
+
 The first implementation picked a fresh ephemeral port and a fresh session key on every start,
 described as a per-run secret. That is wrong for a sidecar other processes talk to: an agent's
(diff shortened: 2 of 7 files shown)


mcaldas:
@h0x91b could you review this PR? All 11 CI checks passed on commit 51599fff, including lint, all test shards, Linux/macOS terminal E2E, build-check, and the security scan. Packaged macOS shutdown, crash recovery, legacy recovery, and five restart cycles also passed. The remaining manual acceptance checks are listed in the PR description: a real routed Codex conversation across restart and legacy takeover specifically on port 32123.

PR #1769, opened 2026-09-14
Add a Simplify View toggle to declutter advanced UI for casual use

lidord-wix:
## Why

dev-3.0 has accumulated a lot of power — agent traffic, stats, automations, bug hunters, hibernate, tmux controls, scripts, custom columns, spawn-agent/spawn-variant, and more. That's great for a power user, but for day-to-day work (create task → Kanban → terminal → basic git) it adds up to a lot of always-visible chrome. This PR adds an opt-in way to turn that down without losing anything.

## What

A **"Simplify View"** toggle in Settings → System → Advanced Experience. Off by default. When on, it hides a closed list of advanced/chrome surfaces at their existing conditional seams (no new wrapper component, no data model changes):

- Agent traffic (screen + header pill/readout)
- Stats nav entry (dashboard card, header kebab, native/browser menu, command palette)
- Automations tab (with a redirect so nobody's stranded on a hidden tab)
- Spawn-variant creation and spawn-extra-agent ("Add agent") — existing variants/sessions stay usable
- Bug hunters, hibernate, scheduled-message creation, in-panel tmux pane controls, the scripts runner
- The "Setup Dev Server" prompt for a project with no dev script configured yet (the *running* dev-server control is untouched)
- tmux session manager, the keyboard-shortcut *editor* (the ⌘/ reference overlay stays)
- Header ambient resource + agent rate-limit readouts
- Custom-column and label *creation* (existing ones keep rendering)

**Never hidden:** notes, image/artifact viewers, help mode/guided tours, diff-review inline comments and the GitHub review layer, the running dev-server control, open-in, ports — these are core outputs/review work, not clutter.

**Nothing is destroyed.** Hiding a surface only removes its entry points; existing labels, columns, variants, and automations keep working underneath, and turning the toggle back off restores everything instantly.

## Design process

This went through the repo's `/ux-principal` skill before any code was written — three sub-agents (IA/placement, interaction/states, consistency/anti-pattern) analyzed the manifest and codebase, converged on: one global boolean (not a per-feature checklist), off by default, subsuming the existing "Advanced Experience" beta-flag pattern rather than fighting it. `docs/ux/PRODUCT_UX_BIBLE.md` §5.10, `docs/ux/ux-architecture.yaml`, and `docs/ux/UX_DECISIONS.md` are updated with the closed hide-list and the reasoning, including a couple of same-day corrections after real usage feedback (restored-then-re-hidden "Add agent"; swapped in the "Setup Dev Server" prompt).

## Testing

- `bun run lint` — clean.
- `bun run test` — 362 mainview / 470 backend / 55 CLI test files, all green except one pre-existing, unrelated flaky fs-watch test (`agent-message-log-watch.test.ts`) that passes cleanly in isolation and touches no file this PR changes.
- Added unit tests for the new toggle (`AdvancedExperienceSection.test.tsx`) and the settings drift-guard/registry-completeness fixtures.
- Manual walkthrough: toggled on/off in the running app, confirmed every listed surface hides/reappears with no data loss.

## Not implemented

- Per-feature granular toggles (deliberately one master switch — see UX_DECISIONS.md for why).
- Any change to the underlying capabilities themselves — this is a visibility filter only.


h0x91b (changes requested):
Hi Lidor — Claude here, the AI assistant working with Arseny on dev-3.0. I read the whole branch line by line, type-checked it and ran the suites it touches. First: thank you. Not for the code (though it's clean) — for naming the problem. "The toolbar is overwhelming" is the kind of thing everyone feels and nobody writes down, and you showed up with a working branch and before/after screenshots instead of a complaint. That's a gift.

I'm requesting changes, but **not because the branch is bad**. It's because the design moved forward in the discussion after you opened it, and the new shape is bigger than what's here. As agreed in the thread, treat this as a new, larger task rather than a list of nits — the spec below is written so your agent can pick it up directly.

The good news: your conditional seams are the right seams. Almost none of this work gets thrown away — the `!simplifyModeEnabled` checks you placed become `isHidden("<id>")` checks in the same spots.

---

## 1. What changes conceptually

Today Simplify View is a boolean that a hardcoded list of components consult. The new model turns that inside out:

**The real feature is per-control hiding.** Any one control can be hidden by the user. Simplify View survives, but it becomes a *preset* — turning it on is exactly as if the user had gone around and hidden those controls by hand. Nothing more special than that.

That single move buys three things:

- **It answers "how do we tell users about a new feature?"** A new control's id is simply not in the hidden set, so it's visible by default. A user on Simplify View still sees new things appear and hides them when they feel like it.
- **It kills the double bookkeeping.** In the current branch, `agent-traffic-flag.ts` grew an `effectiveEnabled()` that ANDs the user's own agent-traffic setting with simplify mode, plus a cross-flag event re-dispatch. With one hidden set there is one answer per surface and that whole layer disappears.
- **The overflow list can never lie.** There's one place that knows what's hidden, so what the dropdown shows is what's hidden.

---

## 2. The spec

### 2.1 One registry, one source of truth

- Every hideable surface gets a stable string id.
- State is a set of hidden ids, stored in `GlobalSettings` — **not `localStorage`**. It has to follow the user into a `dev3 remote` browser session, which is exactly where a calmer UI matters most.
- **Absent id = visible.** New controls are always visible on first appearance, for every user, including one who has Simplify View on.
- **Simplify View = a named preset** that writes the recommended ids into that set. Off by default, same as now.
- **"Show all" clears the set.** This is the escape hatch and the undo for the preset; without it the preset is a one-way door.

### 2.2 Hiding a control

- **Desktop: right click on the control → a context menu with a single "Hide" item.**
- **Touch and remote have no right click.** dev3 runs in a phone browser via `dev3 remote`, and long-press fights with scrolling, so pick a reachable path there (a "Hide" affordance inside the overflow dropdown is fine). The requirement is that hiding is reachable with a finger — not that it's implemented in any particular way.

### 2.3 Restoring — the task info panel

The inspector header is a documented 2×2 grid of bars (`docs/ux/PRODUCT_UX_BIBLE.md` §5.1): Context (row 1 left), Session/Agent (row 1 right), Git (row 2 left), Runtime & access (row 2 right). Each bar owns one action domain, and each already has a "≤ 4 visible, then overflow" budget — so overflow is an idiom that exists, not one you invent.

- **Exactly one restore control for the whole panel**, and it lives in the **panel chrome** — the far right of row 1, next to the ⚙ and the collapse control. §5.1 states that chrome is not a bar and does not count against any bar's budget, which is precisely why it belongs there. Four little eyes (one per bar) would be noisy; putting the single one into the bottom-right bar would park a general control inside the Runtime domain, which §5.1 explicitly warns against.
- It **appears only when this panel has something hidden**, and disappears when the set is empty again.
- The dropdown **groups rows by bar** with the bar's name as the group heading, so the user can see where a control will return to.
- **Each row is dual-purpose:** left click on the row runs the control's action (a rarely used button stays usable without un-hiding it), and a small **"return" icon on the right of the row, on hover** puts it back into its own bar. On touch that return icon is always visible, never hover-only.
- **Last row, permanent: "Show all".**

### 2.4 Restoring — the global header

The header gets **no new icon at all.** Exactly three controls may be hidden there, and they fold into the **existing kebab `⋯`** (`src/mainview/components/GlobalHeader.tsx:1135`), which is already where the header folds things on narrow viewports:

| Control | Where it is today |
|---|---|
| Agent-traffic indicator (the pill with the count) | `AgentTrafficIndicator`, rendered in the header row |
| Project terminal button | `GlobalHeader.tsx:1268` |
| Remote-access QR button | `GlobalHeader.tsx:1303` |

Nothing else in the header is hideable in this round. The rule behind it: **one restore point per container, and if the container already has one, reuse it.**

### 2.5 What must not be in the hide-list

Anything with no visible anchor to right-click and no row to restore. Concretely, **hint navigation** (`F` / `⌘G`): the current branch gates it at `src/mainview/App.tsx:1445`, but the ⌘/ overlay and `docs/keyboard-shortcuts.md` keep advertising it, so the key just silently stops working. Either it isn't hideable at all, or it gets an explicit row in Settings → Keyboard. And if any shortcut is ever gated by a flag, it must go through `FLAGGED_SHORTCUTS` in `src/mainview/keymap.ts` — §5.10 as written in this very PR already requires that.

### 2.6 Docs that move with the code

- `docs/ux/PRODUCT_UX_BIBLE.md` §5.10 rewritten for the preset model, and marked `Observed` rather than `Proposed` once it ships — every implemented section in that file is `Observed`, and the budget-test comments in `src/bun/__tests__/ux-docs-budget.test.ts` currently say "not yet implemented" about code that is right here in the diff.
- `docs/ux/ux-architecture.yaml` → the `simplify_mode` entry, same treatment.
- `docs/ux/UX_DECISIONS.md` → one entry recording the pivot: per-control hiding with Simplify View as a preset, and that a settings screen full of checkboxes was considered and dropped in favour of right-click + overflow.
- `change-logs/2026/09/14/feature-simplify-mode-toggle.md` describes the old model and needs rewriting (and its date should match the day it actually merges).

---

## 3. Findings in the current branch that survive the redesign

These are real today and, except where noted, still real after the rework — the preset still has to hide the non-toolbar surfaces (stats nav row, automations tab, settings categories), and that's where they live.

**Worth fixing**

1. **`src/mainview/App.tsx:1445`** — hint navigation is gated but `FLAGGED_SHORTCUTS` was not extended, so ⌘/ and the docs advertise a dead key. See §2.5 above.
2. **`src/mainview/components/GlobalSettings.tsx:1019`** — settings **search** still returns Keyboard-category entries while the category is hidden. Selecting one sets `activeCategory = "keyboard"`, the new effect bounces it back to Appearance, and `pendingAnchor` points at an anchor that never renders: the user lands somewhere else with no explanation. Search is an entry point like the palette and the menu, so it needs the same filter.
3. **`src/mainview/components/GlobalSettings.tsx:158`** — the reset effect falls back to `SETTINGS_CATEGORIES[0]` instead of the visible list's first entry. Correct today only because `appearance` happens never to be hidden. Related: the heading reads `selectedCategory` while the body renders `renderCategoryPage(activeCategory)`, so a deep link to `?section=keyboard` paints the shortcut editor under an "Appearance" heading for one frame.
4. **Tests.** Thirteen components gained a gate and only the toggle switch itself is covered (`AdvancedExperienceSection.test.tsx`). Nothing would catch a regression that un-hides a surface. After the rework the natural unit is the registry: hidden-set round-trip, "absent id is visible", preset application, and "Show all" clears it — plus one test that a hidden control really leaves the palette and the native menu.
5. **`src/mainview/App.tsx:1790`** — the agent-traffic screen redirects to the dashboard when its flag goes false, but there's no equivalent for `screen === "stats"`, which §5.10 requires ("parked on a hidden destination ⇒ redirect to dashboard"). Reachable with a second window or a remote session parked on Stats.

**Smaller things**

6. `settings.simplifyModeDesc` in all three locales lists the hidden surfaces but omits hint navigation and the diff include-tests toggle, both of which it does hide. Worth re-writing wholesale for the new model anyway.
7. `src/mainview/components/TaskInfoPanel.tsx:1705-1709` — indentation is broken on the second `TaskScripts` guard.
8. `src/mainview/agent-traffic-flag.ts:55` — the simplify listener re-dispatches the agent-traffic change event unconditionally, even when the effective value didn't change. This one dissolves with the single registry.
9. `src/mainview/simplify-mode-flag.ts:33` — `setSimplifyModeEnabledForTests` has no reset path and now silently changes `getAgentTrafficEnabled()` for any later test in the same process. Also dissolves.

---

## 4. What stays exactly as you built it

Off by default. Hiding never deletes anything — existing labels, columns, variants, automations keep working underneath. No new wrapper components, no data-model changes, hiding at the existing conditional seams. All of that was the right call and none of it changes.

One practical note: this is now a decent chunk of work. One PR or a couple of stacked ones is entirely your call — nobody here is counting. And if anything in the spec above reads ambiguous, say so on this thread; an ambiguous spec is our bug, not yours.

Thanks again for pushing on this. The app genuinely needs it.

h0x91b (approved):
Claude here again (the AI assistant working with Arseny). I re-reviewed the rework and it does everything the last review asked for — one registry instead of a boolean, the preset writing into it, right-click Hide, one restore control in the panel chrome grouped by bar, the header folding into its existing kebab, and hint navigation deliberately left alone with the reason written down. The five earlier findings are all closed too. Nice work.

I drove it in a real browser rather than only reading it, and found one thing that had to be fixed before merge, so **Arseny asked me to fix it here directly rather than send you another round.** I pushed two things to your branch:

**1. The blocker.** Turn Simplify View on, then flip any other toggle on the same Settings screen — every hidden id was wiped off disk and the toggle switched itself back off. Reproduced twice on a throwaway board, reading `settings.json` after each click.

The cause is not really yours: every save from `GlobalSettings.tsx` sends the *whole* settings object, built from a snapshot taken when the screen mounted, and the host replaces the stored file with it. Anything written from outside that screen since — `hiddenControls`, from a right-click Hide or from your preset — was silently reverted by the next unrelated toggle. Your feature is simply the first one to write settings from outside that screen, so it is the first to notice.

Fix: `GlobalSettings.tsx` now follows the `globalSettingsUpdated` push and refreshes its snapshot, exactly as `KanbanBoard.tsx` already does. Plus a regression test in `GlobalSettings.test.tsx` that fails without it (I checked that it fails, not just that it passes).

**2. A rebase onto current `main`,** which had moved on. One real conflict, in `TaskInfoPanel.tsx`: `main` had extracted the send-later and hibernate SVGs into `SendLaterIcon` / `HibernateIcon` while your commit wrapped those same buttons in `HideableControl`. Resolved by keeping both — `main`'s icon components inside your wrapper. I also moved the changelog entry to today's date, since that is the rule for the day a change actually ships.

Your commit keeps your authorship; the fix is a separate commit on top.

Verified after the rebase, in the browser: hide → the control disappears and the id lands on disk; the restore dropdown groups by bar and "Show all" brings everything back; Simplify View on now survives an unrelated toggle on the same screen. `bun run lint` clean, and the full local suite green (5838 renderer, 8220 backend, 1132 CLI, zero failures).

Three smaller things I deliberately left alone rather than widening this PR — worth a follow-up whenever you feel like it:

- The per-row restore arrow in the panel dropdown is hover-only (`opacity-0 group-hover:opacity-100`), with no touch branch, although the §5.10 text added in this PR promises it is always visible on touch. On a phone or a remote session you can hide a control but can only get it back via "Show all".
- `settings.simplifyModeDesc` / `…Caveat` are still the v1 copy in all three locales: they claim header readouts get hidden (they no longer do) and that the preset "overrides the beta toggles below" (it no longer does), and they never mention right-click hiding, which is the feature's main gesture.
- `HideControlMenu` positions itself at the raw cursor point with no viewport clamp, while `PanelRestoreControl` right next to it uses `useViewportClamp` — near a screen edge the Hide menu can open partly off-screen.

Thanks for taking the redesign on and for the clean seams — the rework was genuinely pleasant to read. Merging once CI is green.

Issue #1544, opened 2026-08-26
Add support for pi and omp


vit-pavlenko:
I would like to use the following two popular harnesses with dev3:
- https://github.com/can1357/oh-my-pi
- https://github.com/earendil-works/pi

At the moment these do not seem to be supported natively by dev3.

Thanks.

h0x91b:
> *This was generated by AI during triage.*

## Triage Notes

**What we've established so far:**

- dev3 does not currently provide native launch support for the two linked harnesses: oh-my-pi and pi.
- The request is an enhancement, but it does not define the compatibility contract for either harness.
- No prior rejection record exists in the repository knowledge base.

**What we still need from you (@vit-pavlenko):**

- Should support cover launch only, or also hooks, session resume, model selection, auth/account handling, and task prompt delivery?
- Which operating systems and install commands should the presets support?
- Should these be first-class built-in presets, or is a configurable custom-command preset sufficient?
- Are there specific command-line flags, environment variables, or expected default models that must be preserved?

Once those boundaries are clear, the request can move from clarification to an implementation-ready brief.

vit-pavlenko:
Please see my answers in line.
 
- Should support cover launch only, or also hooks, session resume, model selection, auth/account handling, and task prompt delivery? **Yes, ideally it should. I'd like full feature parity with existing dev3 features for Claude Code.**
- Which operating systems and install commands should the presets support? **macOS, Linux (e.g. Ubuntu) for WSL.**
- Should these be first-class built-in presets, or is a configurable custom-command preset sufficient? **First class, ideally.**
- Are there specific command-line flags, environment variables, or expected default models that must be preserved? **No specific flags, environment variables, or default models need to be forced.**

h0x91b:
Hi @vit-pavlenko, Claude here (the AI assistant helping with dev3 triage). Thanks for the detailed answers, and for the omp PRs themselves!

Where things stand on `main`:
- **Oh My Pi (`omp`)**: first-class since #1776 (v1.54.1), with automatic status through the generated `--hook` extension since #1777 (v1.54.2). Launch, protocol delivery, skills, targeted resume, approval modes, `--thinking` and `--model` are covered. Not yet at Claude parity: model-catalog roles, account switching / sign-in detection, and rate-limit tracking. macOS was validated. We haven't seen a Linux/WSL run reported yet, so if you've tried one, that would be great to hear.
[...]

PR #1776, opened 2026-09-14
Add Oh My Pi (omp) as a first-class coding agent

vit-pavlenko:
## Summary

Adds [Oh My Pi](https://github.com/can1357/oh-my-pi) (`omp`) as a built-in coding agent. Part of #1544 — Pi follows in a separate PR.

Until now omp could only run as a custom agent, where the generic adapter delivers no dev3 protocol, cannot resume, and emits `--permission-mode`, `--effort` and `--max-budget-usd`, none of which omp defines.

## What changes

- **Launch flags** — `--model`, `--thinking` for effort, a positional prompt, resume via `-c` / `--resume <id>`.
- **Approval modes** — `acceptEdits` → `--approval-mode write`, `bypassPermissions` / `dontAsk` → `yolo`. "Ask first" passes `always-ask` explicitly, because omp's own default is `yolo`. No Plan preset: omp enters plan mode from a setting, not a launch flag.
- **dev3 protocol as a file** — `--append-system-prompt <file>`, which omp reads when the value names one. Keeps the ~27 KB body out of argv (#1734); `systemPromptFileFor` now covers omp.
- **Presets** follow Claude's shape: Bypass at every thinking level (`off` → `max`), Ask first and Accept Edits once each. No preset pins a model — omp reaches 60+ providers.
- **Skills** install into `~/.omp/agent/skills` and are invoked as `/skill:<name>`.
- **No sign-in probe** — omp keeps credentials in SQLite (`~/.omp/agent/agent.db`), not a JSON store, so readiness reports `unknown` rather than guessing a path and blocking the sandbox.
- **Settings command preview** shows omp's real flags. It previously showed the generic `--permission-mode` / `--effort`; the approval map now lives in `omp-flags.ts`, shared by the launcher and the preview.
- **Install hint** is the upstream curl installer rather than the Homebrew tap, whose read-only binary conflicts with `omp update`.

Rationale: `decisions/2026/09/12/omp-first-class-agent.md`.

## Validation

- `bun run lint` and `bun run test` after rebasing on `main`. One `TaskDiffViewer` case failed under full parallel load; it passes in isolation both on this branch and on `main` without this change.
- Against omp 18.1.18 directly: every emitted flag checked against `--help`; the file-borne system prompt confirmed by putting an instruction in a file and watching the model obey it; an invalid `--approval-mode` confirmed to fall back silently to `yolo`, which is why a test pins the exact mapping.
- By hand in a browser, with the branch running headless on a scratch QA board: preset list and order, command previews, launching omp tasks, approval-mode behaviour, protocol delivery, resume.

## Not covered

- **Automatic status** — cards move only if the agent follows the protocol. Hooks via omp's `--hook` extensions are a separate change.
- **Skills and bug hunters were not exercised.** Headless mode (`dev3 remote`) never runs the skill installer, so on a headless-only machine no agent gets dev3 skills. That predates this change and affects every agent; it is not addressed here.
- **Recovery** resumes omp's most recent session (`-c`); omp has no flag to pre-assign a session id.
- The preset editor still shows fields omp ignores (Max Budget, Plan / Auto modes) — the same generic form every non-Claude agent gets.


h0x91b (changes requested):
Hey — Claude here, the AI assistant doing a review pass on this branch.

First off: thank you for this one. The decision record is unusually good — you verified the behaviour against a real omp build instead of trusting its `--help`, and you wrote down the awkward parts (credentials in SQLite, an invalid `--approval-mode` being swallowed) rather than glossing over them. The skill dirs, the three locales, the command-line budget test and the support matrix were all updated together, so nothing is left half-wired. That made this easy to read.

I'm requesting changes for one thing, plus a handful of smaller ones. Nitpicks are left out on purpose.

---

## The blocker: for omp, "no flag" means auto-approve

`src/shared/agent-adapters/omp-flags.ts:9` maps `acceptEdits`, `bypassPermissions` and `dontAsk`. `plan` and `auto` fall through, so `omp.ts:29-32` emits no `--approval-mode` at all — and by your own investigation, no flag means omp runs at its configured default, which upstream ships as `yolo`.

The permission dropdown in Settings (`AgentSettingsSection.tsx:759-766`) offers Plan and Auto for **every** agent, unfiltered. So a user who picks the most restrictive mode gets the least restrictive behaviour: no error, no warning, and the command preview looks correct because it faithfully shows the same absent flag.

Every other adapter can leave a mode unmapped because the CLI then asks. omp does not ask. That inverts the assumption the surrounding code is built on, so the fallback belongs in the adapter rather than in each caller:

- map `plan` and `auto` onto `always-ask`, **or**
- have the adapter emit `always-ask` for any mode it has no mapping for — fail closed rather than open.

Worth noting `copilot.ts`, which landed on `main` as #1779 while this was open, maps all five modes including `plan` and `auto`. Same shape of problem, and it makes a good template.

## Also worth fixing

**1. The shipped default preset is the same trap** — `src/shared/types.ts:1055,1060`. `omp-default` passes no flags and is the `defaultConfigId`, so a fresh install launches omp unattended under a preset labelled "Default". Every other agent's "Default" means "ask me first". Either make `omp-ask` the default, or rename the preset so it is clear it inherits whatever the user's own omp config says.

**2. Type the approval map to `PermissionMode`** — `src/shared/agent-adapters/omp-flags.ts:9` is `Partial<Record<string, string>>`, so a missing or misspelled mode is not a compile error. Given the silent-`yolo` fallback your own Risks section flags, this is the one place where the type system could catch it for free. Tightening it to `Partial<Record<PermissionMode, string>>` also makes the blocker above show up as a type error.

**3. Stale comment right above the code it documents** — `src/bun/agents.ts:520` still reads "Only Claude has such a flag (`--append-system-prompt-file`)", while `SYSTEM_PROMPT_FILE_COMMANDS` on the next lines now holds `claude` and `omp`.

**4. The command preview drops the `--` separator** — `src/mainview/components/global-settings/utils.ts:219` pushes the prompt directly, while the adapter emits `-- '<prompt>'` (`omp.ts:50`). The file's own comment calls the preview a hand-kept mirror of the launcher, and it is already off by one token.

**5. `--thinking` can be sent twice** — `src/shared/agent-adapters/omp.ts:33` pushes `--thinking` from `config.effort`, while the shipped Bypass presets already carry their own `--thinking` in `additionalArgs`. A user who sets Effort on one of those gets the flag twice, with whatever precedence omp happens to have. Skipping `config.effort` when `additionalArgs` already names `--thinking` would settle it.

**6. The protocol still tells omp to type plain slash commands** — `skillInvocationPrefix` gained `/skill:` for omp, but it is consumed in exactly one place (`tmux-pty.ts:3378`, the bug-hunter launcher). The generic protocol body hardcodes `/dev3-share-artifact`, `/dev3-project-config` and `/dev3-tmux` (`src/shared/agent-skill-content.ts:205,229,248`), which omp will not recognise. This is pre-existing for Codex too (it needs `$`), so I would not block on fixing it here — but it is worth a line in the decision record so it does not look solved.

## Could you rebase?

The branch is 5 commits behind `origin/main` and GitHub currently reports the PR as conflicting. Most of that is #1779, which added GitHub Copilot CLI as another first-class harness and touched exactly the same files this PR does — `families.ts`, `types.ts`, the adapter registry, `agent-skills.ts`, the command-line budget test and `agent-support-matrix.md`. A `git fetch origin main && git rebase origin/main` before the next push will make the conflicts explicit while both changes are still fresh, and CI will finally be running against something merge-able.

Two small things the rebase will also surface: the changelog entry is dated `2026/09/14` and the repo wants the expected merge day, and `agent-support-matrix.md` now needs an Oh My Pi **and** a Copilot column.

---

For what it's worth, the parts I could check locally are clean: `bun run lint` passes and all the suites this branch touches are green (373 tests). I could not verify any of the omp runtime claims myself — omp is not installed on the machine I ran this on — so the silent-`yolo` default and the file-valued `--append-system-prompt` are taken from your investigation, not independently confirmed.

Really nice piece of work overall — it is one map entry away from being safe by default.

vit-pavlenko:
Thanks for the careful pass. Everything is addressed on the new head (`b1feab08`), rebased onto `main` on top of the Copilot adapter, and this time the omp claims were verified by launching real tasks through a branch build with omp 18.1.19 installed.

## The blocker — no flag means `yolo`

Fixed at the adapter, the way you suggested, but one step further: the map is now **total**. `OMP_APPROVAL_MODE` is `Record<PermissionMode, "always-ask" | "write" | "yolo">`, so a new mode is a compile error, and the adapter emits `--approval-mode` on **every** launch — `default`, `plan` and `auto` all say `always-ask`. `default` is included deliberately: on every other CLI "no flag" means "ask me", and for omp that assumption was wrong in the one place it matters most. (`plan` has no launch flag on omp — it comes from `plan.defaultOnStartup` — and `auto` has no classifier-gated tier to land on, so both fail closed.)

Verified live: a task launched on the **Default** preset ran `omp --approval-mode always-ask …` and stopped for approval before every single bash call, including the protocol's own `dev3 task move`. A task on **Bypass · Low** ran `--approval-mode yolo --thinking low` and finished unattended.

## The rest

1. **Default preset** — with `default` now mapped, `omp-default` asks first like every other agent's "Default", so the separate `omp-ask` preset was redundant and is gone. Shape is now Default / Accept Edits / Bypass × six thinking levels.
2. **Typed map** — done, see above.
3. **Stale comment** in `agents.ts` — rewritten to name both flags.
4. **Preview `--`** — the preview now emits the separator for omp. (Claude's preview has the same gap and always has; left alone here.)
5. **`--thinking` twice** — the adapter skips `config.effort` when `additionalArgs` already carries `--thinking` (either `--thinking x` or `--thinking=x`), and the same rule keeps a user-supplied `--approval-mode` in `additionalArgs` as the one that wins. The preview mirrors both.
6. **Slash commands in the protocol body** — noted in the decision record as a known gap shared with Codex, so it does not read as solved. `/skill:dev3` completion itself was verified against the installed skill dirs.

Rebased; the matrix has both new columns, the changelog moved to today. `bun run lint` and the local suites (renderer 5805 / backend 8175 / CLI 1132) are green; CI on the new head is running now.

h0x91b (approved):
Claude again — re-reviewed the rebased branch at `b1feab080`.

Everything I raised is addressed, and the blocker is fixed the stronger way: instead of adding two entries to the map, `OMP_APPROVAL_MODE` is now total and typed on `PermissionMode`, so a future mode is a compile error rather than a silent `yolo` session, and the adapter emits the flag unconditionally — resume included. `default`, `plan` and `auto` fail closed to `always-ask`. Dropping the separate "Ask first" preset once "Default" genuinely asks is the right call; it now matches every other agent's shape.

The rest checks out too: `hasOmpFlag` handles both `--x v` and `--x=v` so a preset's own value wins undoubled, the preview mirrors the launcher including the `--` separator, and the stale `agents.ts` comment, the decision record, the matrix and the changelog date all followed. I especially liked that the `/skill:` gap for the generic protocol body got written into the decision record rather than quietly dropped — that is the honest place for it.

Verified locally on the new head: `bun run lint` clean, backend and renderer suites green (402 tests). CI is green on `b1feab080` across all eleven contexts, and the PR is no longer conflicting.

Still not verified by me, same as last time: omp is not installed on the machine I ran this on, so the runtime facts — that an absent `--approval-mode` inherits `yolo`, that `--append-system-prompt` reads a file — remain yours, not independently confirmed.

Thanks for the quick and thorough turnaround. Approving.

PR #1793, opened 2026-09-16
Comment on anything the agent made, from one review the agent can resolve

arditti:
## What

Review comments become one model across dev3's viewers, stored on the task, closable by the agent.

- **Comment on everything the agent produced.** HTML artifacts (click an element), shared images (drag a region), previewed files (select text, or drag a region when the previewed file is an image; Cmd-click a path in the terminal to open one) and the terminal itself (select text, click the `Comment` chip that appears for six seconds) all open the same composer and land in the same review.
- **Comment on HTML artifacts.** The artifact viewer gets a `Comment` toggle in its header. With it on, clicking any element of the report opens the same composer the diff viewer uses, beside the frame. The comment is stored with an `artifact-element` anchor (artifact id, version, CSS path, element text, nearest heading). The frame draws numbered pins itself; after a republish, pins it cannot place are listed as `Not found in this version` rather than dropped.
- **One review model** (`src/shared/review.ts`): a comment is text plus a typed anchor — `diff-line`, `artifact-element`, `image-region` (normalised 0..1), `file-range` (path, line range, excerpt) or `terminal-text` (excerpt). The composer, thread bubble and prompt serializer move out of `TaskDiffViewer.tsx` into `src/mainview/review/`; the diff viewer only projects the task's review onto its widget map. Behaviour on the diff is unchanged except as noted below.
- **The store is `Task.review`** instead of browser localStorage with a 3-day expiry. Same list on the desktop, in a remote-mode browser, and for the agent. One RPC per mutation so a CLI resolve never races a renderer edit. A review a browser still holds in the old key is imported once on the next diff open, then the key is dropped. Capped at 200 comments, resolved ones evicted first.
- **The agent closes the loop:** `dev3 review list [--unresolved] [--json]`, `dev3 review resolve <id> [--reply "…"]`, `dev3 review reply <id> "…"`, `dev3 review reopen <id>`. The prompt sent to the agent now carries each comment's id and a footer naming these commands. Replies render under the comment; a resolved bubble shows `Resolved` and offers `Reopen` in place of `Send`, so the per-comment budget stays at three.
- Comment ids are UUIDs so 8-character CLI prefixes work.

Sends are unchanged: comments are marked sent, never deleted (`never-destroy-a-review-on-send` still holds).

## Why

The diff viewer already had the full commenting loop for code and markdown, but it was private to one 4 600-line component and its store was per-browser and expiring, so artifacts could not be commented on and the agent could never read a comment. Decision record: `decisions/2026/09/16/review-comments-on-the-task-record.md`.

## How to check

1. Open a task with an HTML artifact → Artifacts → the header shows a speech-bubble **Comment** toggle. Click it: the frame gets a crosshair and a **Review** aside opens on the right.
2. Click any element in the report → the composer opens with the element's text as its anchor. Type a comment, **Add comment** → a numbered pin appears on the element and the thread lists in the aside.
3. **Send N to agent** in the aside header → the agent pane receives `<reviews>` with an `<artifact …>` anchor and a footer naming `dev3 review resolve`. The comment greys out as **Sent**.
4. From the task's worktree: `dev3 review list` shows it; `dev3 review resolve <id> --reply "done"` → the pin turns green, the bubble shows **Resolved** and the reply; **Reopen** brings it back.
5. **Images**: open a task's Images → the same speech-bubble toggle → drag a rectangle (or click a spot) on the picture → composer, then a numbered region is drawn on the image. **Files**: Cmd-click a path in the terminal → select text in the preview → floating **Add comment** → the aside opens, commented lines get an accent bar. **Terminal**: select text in the terminal → a **Comment** chip appears near the cursor → click it → composer.
6. Open **Show Diff** → the Review Export card lists every comment after the diff comments; the batch **Send** covers unsent comments of every kind.
7. Diff comments work as before (gutter, markdown selection, Send now, per-comment Send, Copy, Reset). Reload the app: comments persist. Open the same task from a phone in remote mode: same comments.
8. Negative: a task that has finished, or an artifact opened for a task with no project (archived modal), shows no Comment toggle.


shamrai-nikita:
Hey, the Comment button appears every time you select anything in Tmux. 
Is it possible to disable it in settings?
@arditti @h0x91b

h0x91b:
@shamrai-nikita Thanks for flagging this! We've removed the floating Comment button from terminal text selection in #1824, which is already merged. The change will be included in the next stable release, so it will stop appearing when you select text in tmux—no settings change needed. Text selection and copying still work as before.

